PR #3, opened 2016-05-09
Editormodule

### LINKED ISSUE #2: Make editor a Browserify module
The editor control and transactionhandler is not yet Browserify modules.

afogelberg:
This pull request fixes #2 . It also adds a dropdown menu to choose which layer to edit and adds a control to open the editor toolbar. When closing the editor toolbar feature info is reactivated.

Therer is also a contstraint option added for attribute editing. With this option, different options for an attribute are shown depending on a choice on another attribute.

There has also been a problem with Internet Explorer caching wfs requests. This means that edits are not shown in Internet Explorer when page is refreshed. This has been fixed by changing the wfs source loader.


markus659:
Seems to work OK. Two things I would like to change:

* After creating a new feature, I want to have the attribution window to show up right away. I think most often you would at least want to confirm that the default values are correct (Although I don't know if you can set default values as of right now?). Or if you are using the module for say tree inventory, you would probably want to set the species attribute without having to select the attribute button.

* When re-sizing the browser window to emulate a mobile device the toolbar doesn't scale too well (haven't actually tried it on a mobile device), maybe it can be changed so that it's more suitable for a mobile design as well?

These are things that can be changed later, so I don't see a problem merging it right away. Good job with the module, especially being able to change layer and the constraint option!

afogelberg:
Thanks, I will merge. Good points Markus and I agree that it would be nice to have an option for showing the attribute window after insert. Also the editor should be enhanced to be more responsive. I will create new issues for these remarks.

PR #34, opened 2016-06-29
Upgraded to OL 3.16.0

afogelberg:
This PR updates package.json to OL 31.16.0. Only problem I have found are need for polyfills for animation och layerspy if using IE 9. At this point Origo doesn't depend on those functions.


afogelberg:
Fixes #33

PR #78, opened 2016-09-12
Center

### LINKED ISSUE #75: Need a better way of getting center for search feature
Right now the code looks like this:

```javascript
map.getView().setCenter([coord[0], coord[1]]);
```

But this will not work with different types of geometries (only for points). We need a better way of getting the center point of the search feature.

afogelberg:
Fixes #75 . Added a generic zoomToExtent method to fit a geometry correctly. This will we used in the search control. Any input on this @markus659 ?


afogelberg:
Also added a generic getCenter utility method to make sure the overlay is placed properly.

afogelberg:
Can we merge this @markus659 ?

markus659:
Good job working on this issue, the zoom-to function of the search control seems more versatile now. I tried implementing the code in our version of Origo, and one thing I noted was that the "coord" variable does not get the center point of the polygon in this bit: https://github.com/origo-map/origo/blob/766dde600f49dff0020e829b66287693e42f1d23/src/search.js#L128-L134

So basically if you have set a geometryAttribute, it will get the coords from this attribute, and if it is a polygon, the center will not be correct for the overlay. In fact, this also seems to be true for the zoomToExtent function (as it creates point from an array of polygon coordinates).

Or is it just that in order to search polygon features, you need to have it defined and added as a layer, with layerNameAttribute and idAttribute set? We are currently searching a postgres table on the server side containg polygon features, and do not have the layer published in a web map server. Sorry for the long comment hope it makes sense @afogelberg.

afogelberg:
> Or is it just that in order to search polygon features, you need to have it defined and added as a layer, with layerNameAttribute and idAttribute se

That's correct @markus659 . It was made this way because it is asumed that the coordinates will be a point coordinate in order not to send complete geometries in the autocomplete. However, I see no problem adding the same functionality as for selecting with a map service. I will add this before merge.

markus659:
Would the max zoom option be a limiting factor in the view.fit method if the geometry were to require a zoom level that is more zoomed out? https://github.com/origo-map/origo/blob/766dde600f49dff0020e829b66287693e42f1d23/src/maputils.js#L61

afogelberg:
No, it will always try to fit the geometry. The max zoom option will make sure that the map will not be zoomed in too deep, for example if the geometry is a point.

afogelberg:
@markus659 how do you handle the attributes when you search postgres? Do you have layers defined with attributes?

markus659:
If you mean our client side code, then it can be seen here:

https://github.com/Hallstahammarskommun/origo/blob/master/index.json

https://github.com/Hallstahammarskommun/origo/blob/master/src/search.js

The query is sent to Node which handles querying/selecting data from the tables.

afogelberg:
I see, so everything is handled serverside? We could introduce two more options: titleAttribute and contentAttribute. If both are provided then server side is assumed. I guess the contentAttribute should be html formatted in the response?

markus659:
Yes, pretty much. We have different queries running on two different tables. The queries then both return title and content, regardless of the different field in the tables (we have one for address points and one for parcels). So a typical response for us would look like this:

`[{"st_astext":"POINT(134665.818 6611153.137)","title":"Ica","content":"Hammartorget 3, 734 30 Hallstahammar"}]`

Where "Ica" is from the status field, and content is a combination of the obj, typ, and egn fields.

For us, have a titleAttribute and contentAttribute would work great. I'm not sure how others would implement a server side search handling the attributes, but for the Origo client, just handling title and content from the response would seem like a clean and easy way.

Regarding the html formatting, do you mean that the content should be formatted as html straight away when it's returned from node (the content part in my code example)? Would it require any changes in Origo, or would the popup handle it correctly as it is right now?

afogelberg:
> Regarding the html formatting, do you mean that the content should be formatted as html straight away when it's returned from node (the content part in my code example)? Would it require any changes in Origo, or would the popup handle it correctly as it is right now?

The popup would handle it correctly with your example. In more advanced cases you might want to work with for example html lists. In those casess html should be included directly in the response. Then it will work correctly in Origo.

afogelberg:
I made the search handling more flexible and added an option to get complete feature info from search result as you describe. This is option 3 in the selectHandler. Please try if it works in your environment @markus659  . I have no access to such service and can't verify that it works.

This should be a separate pull request but I think I'll keep it together.

markus659:
I will try it and let you know how it works.

markus659:
Everything seems to be OK as far as the attributes goes (title and content is correct). The problem I am getting is when it goes to the identify function in featureinfo.js.  It is called from https://github.com/origo-map/origo/blob/92aad94d95ca3a229ad6824fa8990a3042c9ae25/src/search.js#L184

which points to https://github.com/origo-map/origo/blob/19497e2e4533212ea4a940711dbb4c5cdcb0858a/src/featureinfo.js#L75

The code then goes on to the overlay case, and it tries to create the owl carousel (which I don't know if it should do it if it's only identifying one feature?). This ends up leaving the body of the popup empty. Added a screenshot that shows the html.

owl

afogelberg:
The response needs to be in a html element container. I added a container for this case. Does it work @markus659 wih this change?

markus659:
Yep, that fixed it. Now that part seems to work well (at least with our setup). I think you can go ahead and merge now, unless there's something else that still needs to be fixed?

Issue #83, opened 2016-09-19
New build for embedded maps


afogelberg:
I suggest that Origo offer an alternative build that can be used for embedded maps. This build would be very simple with only basic functionality such as zoom and open map in full extent. The same configuration file could be used as for the ordinary, but ignoring the control options. Jsonp should be used to get the configuration file as suggested in #70 . With a separate build we could also have a significant smaller build size.

afogelberg:
I have been working on this issue and suggest some changes to make it easier to have separate builds. I suggest that the controls are initiated in the main files and not in the viewer. This means that all controls to be included in the build should be listed in the main files. The main file for origo will still be origo.js but for a separate build we need to have a separate one. I have created a branch where I have implemented these ideas: https://github.com/origo-map/origo/compare/origo-embed

These I think are some important changes. Some opinions on this would be appreciated @markus659 @jokd . If you think this is the way to go I will continue the work. So far there is still only one build.

markus659:
Would all configuration still be in the .json file? Looking at your branch, would all the controls still be included in the origo.min.js-file, or would the ones that are not included not be a part of the minified file (or would you need to manually edit the origo.js file if you don't want them included at all)? 

What would be the difference between the different builds, only the origo.js-file and what controls are included in that file?

Maybe only OL-controls should be in the viewer, meaning the home button should also be part of the Origo controls? See issue #76.

jokd:
I definitely see good use for this and just editing origo.js instead of viewer.js for separate builds is something to strive for.


Från: Andreas Fogelberg [mailto:notifications@github.com]
Skickat: den 23 september 2016 16:57
Till: origo-map/origo <origo@noreply.github.com>
Kopia: Jonas Kumlin <jonas.kumlin@karlstad.se>; Mention <mention@noreply.github.com>
Ämne: Re: [origo-map/origo] New build for embedded maps (#83)


I have been working on this issue and suggest some changes to make it easier to have separate builds. I suggest that the controls are initiated in the main files and not in the viewer. This means that all controls to be included in the build should be listed in the main files. The main file for origo will still be origo.js but for a separate build we need to have a separate one. I have created a branch where I have implemented these ideas: https://github.com/origo-map/origo/compare/origo-embed

These I think are some important changes. Some opinions on this would be appreciated @markus659<https://github.com/markus659> @jokd<https://github.com/jokd> . If you think this is the way to go I will continue the work. So far there is still only one build.

afogelberg:
[...]

PR #86, opened 2016-09-27
Moved away control initialising from viewer

afogelberg:
Can we merge this? The build file for embedded maps will be a separate PR. First step to solve #83


afogelberg:
> is that the copyright text at the top says "Mälardalskartan", which should be changed I guess?

Absolutely, Maybe a new issue?

PR #90, opened 2016-09-28
Tool for measuring length and area

### LINKED ISSUE #89: Tool for measuring
We need a tool for measuring length and area in the map.

jokd:
Fixes #89


markus659:
It seems to work well on both desktop and mobile, the tool is very intuitive to use. One thing I would like to change is the decimal mark. Right now it's the decimal point, but since we currently only have a Swedish version available I think it should be changed to the decimal comma instead.

afogelberg:
> I think it should be changed to the decimal comma instead

I would vote to keep the dot. I think it would make things a little bit complicated as we need to make sure that we use the same delimiter for the rest of Origo as well, for example in featureinfo. Both Eniro and Hitta use dot as a delimiter.

One other thing. Could we use the same blue color as for the pin symbol? That way we keep the style more consistant and it will have higher contrast when used with airial images and greyish base maps.

tonnyandersson:
@markus659 Implementing the array functionality in the viewer.js removeOverlays function seems like a good idea, as long as it doesn't remove other overlays by mistake. I can't see why it would, though,

On another note, the removeOverlays in viewer.js needs an overhaul. Right now it doesn't remove overlays at all, as it seems. I will post an issue regarding this.

@afogelberg I will look into the points you mentioned. Regarding the svg files, the icons were picked from the Minicons and Steadysets packs available at Fontastic. They are defined in separate files for both packs respectively.

Regarding the color, I will try it out with the same blue color as the pin symbol, but I think it might be hard to distinguish on blueish water surfaces and such.

tonnyandersson:
@afogelberg New version of measure tool up for review.

afogelberg:
Thanks for working on this @tonnyandersson . I noticed that the measure control disappears when it is deactivated. It seems to be a lacking o-prefix here https://github.com/origo-map/origo/blob/measure-feature/src/measure.js#L60.

Some small remarks:
* There is a conflict with the rotation button. The measure button hides the rotation button on large screens. I'm afraid I have no suggestion for this.
* The method removeOverlays can be replaced by removeOverlays in #93 .
* The css class tooltip should have a o-prefix, for example o-measure-tooltip.
* I think the style in addInteraction should be defined as a default style when the control is initiated. This way it will be easier to use other styles and define it is as an option in the future.

Finally some small remarks with regard to the airbnb style guide (I know we have some work to do here in the rest of Origo):
* [Blocks](https://github.com/airbnb/javascript/tree/es5-deprecated/es5#variables) : https://github.com/origo-map/origo/blob/measure-feature/src/measure.js#L299
* [Variables](https://github.com/airbnb/javascript/tree/es5-deprecated/es5#variables) : https://github.com/origo-map/origo/blob/measure-feature/src/measure.js#L11
* Use [soft indentation](https://github.com/airbnb/javascript/tree/es5-deprecated/es5#whitespace)

tonnyandersson:
@afogelberg Would there be any conflicts or other issues if I simply move the rotation button down a notch? Can't see there would be. Actually, I meant to post an issue regarding this and related problems. Perhaps it would be a good idea to use a "toolbox div" and add the tool buttons to the div in the order they are loaded and let the "toolbox div" expand accordingly. That way one would not have to change the css depending on which tools are present in the application.

removeOverlays in viewer.js removes ALL overlays in the map, which may not be desirable in all situations. For example, there may be issues using the measure tool and the geolocation tool at the same time, or featureinfo and geolocation at the same time. If we are going to use the removeOverlays function in viewer.js for all such operations, I suggest that we go with @markus659 solution where removeOverlays will take a parameter with the overlay(s) to be removed and simply remove all if the parameter is undefined.

afogelberg:
>  Perhaps it would be a good idea to use a "toolbox div" and add the tool buttons to the div in the order they are loaded

I think this is a good idea @tonnyandersson . I've been planning to make a toolbar for this purpose. To make this work we also need to make an Origo control of the ol Rotate control. The same goes for the ol zoom control.

> I suggest that we go with @markus659 solution where removeOverlays will take a parameter with the overlay(s) to be removed

I agree. I will make an update to #93 .

afogelberg:
I updated removeOverlays #93 as suggested. Is this ok as a solution @tonnyandersson @markus659 https://github.com/origo-map/origo/commit/c730978c23a07a856a29579c7749c88a8b86b7f5

markus659:
Added a comment in #93 regarding your update @afogelberg.

tonnyandersson:
Looks alright to me.

afogelberg:
@tonnyandersson #93 is merged.

jokd:
@afogelberg @markus659 Are we cool merging this?

markus659:
Just a couple of thing my editor picked up when I looked through src/measure.js again:

There should be a blank line after blocks, for instance https://github.com/origo-map/origo/blob/20f417623069c0d37340a801522e8010b4af4c4c/src/measure.js#L69. I think there were one or two more places where a blank was missing. Reference: 
https://github.com/airbnb/javascript/tree/es5-deprecated/es5#whitespace

I don't have a reference for this, but there are some trailing white space here:
https://github.com/origo-map/origo/blob/20f417623069c0d37340a801522e8010b4af4c4c/src/measure.js#L85

I think a new line would be enough?


Good job on getting the changes done so quickly this far.

tonnyandersson:
Fixed the blank lines and white spaces @markus659 mentioned. Please review and comment or merge.

markus659:
I have no more remarks about the code. I will approve this PR. Once @afogelberg approves it, we can go ahead and merge it.

afogelberg:
LGTM. Just one small remark, we should be using soft tabs according to the Airbnb style guide.

And one final question: should we make the line measure tool automatically activated when the measure tool is activated? This way one more click is avoided in many cases.

tonnyandersson:
@afogelberg 2 or 4 spaces? Your first comment on this PR says 4, the Airbnb style guide says 2.

afogelberg:
Sorry about that  @tonnyandersson . It used to be 4 spaces in Origo but since we decided to use Airbnb it's 2 spaces.

afogelberg:
Looks good now. Should we make the line measure tool activated by default @tonnyandersson @markus659 @jokd ?

tonnyandersson:
Works fine either way for me.

jokd:
Sure!

markus659:
Yes, that could work. I'm not sure what the user would expect when pressing the button, but I think length measuring is more common in web maps, I know some don't include area measuring at all. And one click less is always more appealing, and even if you would want to measure an area it wouldn't require any extra clicks if the line measure tool was default. 

So I think we can go with the line measure tool as default.

tonnyandersson:
Line tool is now activated from start by default.

PR #93, opened 2016-09-30
Remove overlays

### LINKED ISSUE #91: removeOverlays does not remove overlays
The removeOverlays function in viewer.js does not actually remove overlays. The overlay elements remain in the document.

Suggest changing the way the overlays variable is set from

```javascript
var overlays = map.getOverlays();
```

to

```javascript
var overlays = map.getOverlays().getArray();
```

afogelberg:
This fixes #91 and makes Popup work better with overlays.

### DIFF
--- a/src/featureinfo.js
+++ b/src/featureinfo.js
@@ -78,7 +78,7 @@ function identify(items, target, coordinate) {
     content = '<div id="o-identify"><div id="o-identify-carousel" class="owl-carousel owl-theme">' + content + '</div></div>';
     switch (target) {
         case 'overlay':
-            var popup = Popup('#map');
+            var popup = Popup('#o-map');
             var overlay = new ol.Overlay({
               element: popup.getEl()
             });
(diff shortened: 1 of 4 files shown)


afogelberg:
I think this should be merged before the measure control #90  as this method should be used to remove overlays.

markus659:
Could you not exclude the first else from this part https://github.com/origo-map/origo/pull/93/commits/c730978c23a07a856a29579c7749c88a8b86b7f5? So that it would look like this instead:

```javascript
    function removeOverlays(overlays) {
      if (overlays) {
        if (overlays.constructor === Array || overlays instanceof ol.Collection) {
          overlays.forEach(function(overlay) {
            map.removeOverlay(overlay);
          })
        }
      } else {
        map.getOverlays().clear();
      }
    }
```

afogelberg:
> Could you not exclude the first else from this part

This is just in case it's not  a collection or array. We get one method for all cases instead of calling map.removeOverlay(overlay) if it's just one overlay.

markus659:
Alright, that makes sense. I figured you would just use Viewer.removeOverlays(); but then again, depending on how you use it, you might not always know if it's just one or an array of overlays, so it would make sense to have it written like it is now so that you can always call Viewer.removeOverlays(overlays) if you're not sure.

afogelberg:
I will merge this then when you make a review acceptance on this @markus659

PR #94, opened 2016-09-30
Added isActive check in transactionhandler

### LINKED ISSUE #92: Triggering enableInteraction event results in error
Triggering enableInteraction event results in an error if an edit session has not been previously initiated. The event triggers the onEnableInteraction function in editortoolbar.js, which in turn calls transactionhandler.removeInteractions() if the interaction is not named 'editor'. This function will fail if the variable settings.map has not been set.

Quick fix: wrap the contents of transactionhandler.removeInteractions() in an if- statement, e.g.:

```javascript
removeInteractions: function() {
    if(settings.map) {
        settings.map.removeInteraction(settings.modify);
        settings.map.removeInteraction(settings.select);		
    }
}
```
Perhaps utilizing flags would be a more flexible solution.

afogelberg:
This fixes #92

### DIFF
--- a/src/transactionhandler.js
+++ b/src/transactionhandler.js
@@ -212,9 +212,10 @@ module.exports = function(){
       });
     },
     removeInteractions: function() {
-        // console.log(settings.select);
-        settings.map.removeInteraction(settings.modify);
-        settings.map.removeInteraction(settings.select);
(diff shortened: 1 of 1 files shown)


tonnyandersson:
Looks good!

PR #95, opened 2016-10-03
Ol builds

afogelberg:
This PR adds a script and build configurations for OL. Using the standard configuration the build size for origo.min.js will be reduced with about 300 kb. This is one further step to solve #83 .

There will be three standard build configurations for OL: one for standard Origo, one for embedded maps of Origo and finally one full build of OL.


afogelberg:
I have also added a simple build file to be used for small embedded maps.

afogelberg:
I noticed jQuery was not minified as much as it should be. By adding the mangle flag to uglify the build size will be reduced significantly. With this flag set the standard build of Origo will now be 525 kb and about 350 kb for the minimal embed build of ol.

markus659:
Excellent! That is a major decrease to the overall build size.

PR #101, opened 2016-10-07
Added home control

### LINKED ISSUE #76: Home button should not be created if no value is set for "home"
If no value for "home" is set in the configuration file, the button should not be created. Unless we want it to be a requirement to set a value for "home"?

afogelberg:
Fixes #76 . A new control has been added for home. If extent is not provided as an option the extent will be calculated from the extent when the map is loaded. Is this ok @markus659 as a solution for #76?


markus659:
Yes, this looks good to me.

Issue #81, opened 2016-09-13
Remove touchend


afogelberg:
Touchend to achieve fast click should be removed. See #80 .

afogelberg:
Fixed with #102

Issue #80, opened 2016-09-13
Scrollning mapmenu with touch


Lund-gren:
When scrolling mapmenu the layer/group you start scrolling on turns on/expands when the touch end. 
Poteniell fix for this: https://github.com/Hallstahammarskommun/origo/commit/bf8e425ac0d0cde083381f224dc15d213ff7d92a

afogelberg:
Thanks for the report and the suggestion for a fix. But I think we are now ready to remove the touchend event that are used to achieve fast click: https://developers.google.com/web/updates/2013/12/300ms-tap-delay-gone-away
It should also work on iOS from 9.3. Removing the touchend should solve the scroll problem.

afogelberg:
Fixed with #102

PR #102, opened 2016-10-10
Removed touchend

afogelberg:
This PR solves #80 and #81 . Can we merge this?

### DIFF
--- a/src/editortoolbar.js
+++ b/src/editortoolbar.js
@@ -128,17 +128,17 @@ module.exports = function(){
     },
     bindUIActions: function() {
       var self = this;
-      $('#o-editor-draw').on('touchend click', function(e) {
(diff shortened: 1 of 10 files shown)


markus659:
There is one touchend remaining here: https://github.com/origo-map/origo/blob/remove-touchend/src/transactionhandler.js#L155

afogelberg:
Thanks for the review @markus659 . I removed the remaing touchend.

markus659:
Seems to work well otherwise (tested on tablet and phone). I will approve the changes.

afogelberg:
@markus659 Can you verify that it also solves #80 ?

markus659:
Yes, it does solve the problem described in #80.

PR #114, opened 2016-10-21
Repaired logo in print-control

### LINKED ISSUE #113: Wrong path to css and logo in printfile
https://github.com/origo-map/origo/blob/master/src/print.js#L48-L49 and https://github.com/origo-map/origo/blob/master/src/print.js#L53 these files have been moved or merged

afogelberg:
This PR fixes #113 and makes the control compliant to our guide lines. Does it solve the regression @Lund-gren ?


afogelberg:
The PR also makes it possible to add attribution text as an option.

Lund-gren:
Yes seems like it works now, my only concern is the the Logo "Fyra mälarstäder" which I think sholud be changed.

src/print.js
```diff
@@ -5,102 +5,100 @@
...
+      }, 10);
+    }, 200);
+  }
 }
+
 function createImage() {
-    var canvas = $('canvas');
-    var image = new Image();
-    // image.crossOrigin = 'Anonymous';
+  var canvas = $('canvas');
```
markus659: jQuery variables should be prefixed with a $.

src/print.js
```diff
@@ -5,102 +5,100 @@
...
-          else if (sourceWidth >= settings.printWith) {
-            printCanvas[0].height = settings.printWith;
-          }
+  try {
+    var imageUrl = canvas.get(0).toDataURL("image/png");
+  } catch (e) {
+    console.log(e);
+  } finally {
+    // printCanvas = copy of original map canvas
+    var printCanvas = $('#o-print');
```
markus659: jQuery variables should be prefixed with a $.

src/print.js
```diff
@@ -5,102 +5,100 @@
...
+    image.onload = function() {
+      var ctxCanvas = printCanvas[0].getContext('2d');
+      var sourceWidth = canvas[0].width; //width of map canvas
```
markus659: comments on newline

afogelberg:
Thanks for the review @markus659 . Made corrections according your observations.

markus659:
Great. The printCanvas variable should also be prefixed with $.

afogelberg:
You're right. Fixed

markus659:
Another thing I noticed just now in the css-file, the o-map-canvas class has its declaration commented, could it be removed or is it kept for some reason? I know there are some commented rows in the regular style.css as well, so maybe the css-files needs some overhauling (which will be needed anyway when we switch over to Sass), and it might not be necessary to include css fixes it in this PR.

afogelberg:
I removed the unused css. I think we can leave the fixes in style for now and fix it when we make the switch to Sass as you say.

Issue #118, opened 2016-10-24
Make map.init more flexible with cors support


afogelberg:
map.init needs to be more flexible and have the possibility to add app configuration that is not part of map options. For example there is a need to add cors support to make a cross origon request of the map options json-file.

This is related to issue #83 .

afogelberg:
Fixed with ad87641

PR #119, opened 2016-10-24
Made map.init more flexible with cors support

afogelberg:
This PR solves #118 and is one more step to solve #83 , The PR suggests that Origo will be initiated this way:

``` javascript
origo.map.init('my map.json', {
  target: 'my map div',
  crossDomain: false //defaults to true
});
```

The second argument is optional and gives the choice to configure app specific setting such as cors support.

All app specfic configs are placed in the directory called conf. This way it will be easier to apply different  configurations of Origo.


afogelberg:
This PR also solves #121

afogelberg:
This PR is an important step to solve #83 but also solves #118 and #121 . All map resources can now be loaded in an embedded map as long as cors is supported. This means that it's not necessary to use an iframe to embed.

A map can for example be initialized like this:
```javascript
<script type="text/javascript">
  origo.map.init('http://yourdomain#layers=topowebbkartan_nedtonad/v/1/s/0,vasteras/v/1/s/0&center=151890,6611154&zoom=6&pin=152302,6610491&map=vas');
</script>
```

The changes summarized:
* Svg sprites are loaded with ajax requests which means svg sprites can be loaded cross origin.
* Relative paths in styles and images are converted to absolute paths with the introduction of baseUrl which is optionally set as a config or set when the map is loaded.
* Svg sprite xlink:href is now written without a path. All sprite paths has been corrected. 
* It is now possible to add an authorization url in the config.

Sorry to have several changes in one PR but the changes are closely related. The branch name is also a bit misleading. 

It would be great if you could have a look at it @markus659 and @jokd

jokd:
Will do when I'm done VABing.

jokd:
Well done @afogelberg! 
However I tried the authorizationUrl config option but instead of prompting for credentials I get a 401. I just had a quick test and perhaps it is solved by using another URL?

afogelberg:
@jokd are you running the test cross origin? I know the authorization trigger will be blocked by chrome and ff if it's cross origin. You should be able to test it in IE with cross origin. The url is a request to a secured resource, for example:
```
    origo.map.init('index.json', {
		authorizationUrl: 'http://localhost:9090/geoserver/test/wms?SERVICE=WMS&VERSION=1.1.1&REQUEST=GetMap&FORMAT=image%2Fpng&TRANSPARENT=true&STYLES&LAYERS=test%3Atest_punkt&SRS=EPSG%3A3010&WIDTH=330&HEIGHT=768&BBOX=153210.26189480544%2C6610525.868478608%2C153210.4541248935%2C6610526.315850449'
	});
```

jokd:
@afogelberg I have tested in Chrome as well as IE with the same result. Request is sent to the Geoserver who returns a 401 (Full authentication is required to access this resource) and then loads the map. 
I would like to be prompted for user/password just like when you turn on a secured layer in the map or load a map with secured content ivsible.

afogelberg:
@jokd is the prompt triggered if you paste the geoserver url directly in the browser?

jokd:
@afogelberg Yes it is

afogelberg:
@jokd is the geoserver request cross origin?

jokd:
I still can´t get it to work. But it might be my setup, seems to be cors and ajax related. I have been able to get around it by appending an iframe to the page when responding 401 but that is not a very pretty solution.
In IE neither way works.

afogelberg:
@jokd how does your setup look like? If you put origo on the same server and port as geoserver, do you still get the same problem?

jokd:
Now I have been able to test it a little more. When running origo on the same server and port as Geoserver it **does** work so there are some cors-issues that at least we will have to deal with before using it in production.
Nevertheless, seems to work nice when setup correctly.

afogelberg:
Thanks for testing @jokd . Is it ok to merge then? I can add some information in the documentation about the cors constraint.

jokd:
It's ok for me so go ahead.

PR #123, opened 2016-10-25
Replaced for loop with forEach

### LINKED ISSUE #122: Replace for loop in legend.js with forEach
I think we can replace the for loop located here: https://github.com/origo-map/origo/blob/master/src/legend.js#L190-L266 with a forEach instead. 

Like it's being used here for instance: https://github.com/origo-map/origo/blob/master/src/getfeatureinfo.js#L46-L57

markus659:
This will fix #122.

### DIFF
--- a/src/legend.js
+++ b/src/legend.js
@@ -187,39 +187,39 @@ function addLegend(groups) {
 
 
       //Add layers to legend
-      for (var i=layers.length-1; i>=0; i--) {
+      layers.forEach(function(layer) {
 
-        var name = (layers[i].get('name'));
-        var title = '<div class="o-legend-item-title">' + layers[i].get('title') + '</div></div></li>';
+        var name = (layer.get('name'));
+        var title = '<div class="o-legend-item-title">' + layer.get('title') + '</div></div></li>';
 
         //Append layer to group in legend. Add to default group if not defined.
-        if(layers[i].get('group') == 'background') {
+        if(layer.get('group') == 'background') {
           //Append background layers to menu
           item = '<li class="o-legend ' + name + '" id="' + name + '"><div class ="o-legend-item"><div class="o-checkbox"><svg class="o-icon-fa-check"><use xlink:href="css/svg/fa-icons.svg#fa-check"></use></svg></div>';
           item += title;
-          $('#o-group-' + layers[i].get('group')).append(item);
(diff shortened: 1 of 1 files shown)


tonnyandersson:
You need to reverse the layers array first. Otherwise the layers will be added in reversed order within a group.

Replace https://github.com/origo-map/origo/blob/master/src/legend.js#L134 with

```javascript
var layers = viewer.getLayers().reverse();
```

markus659:
Thanks for reviewing the changes. The items in the returned array seems to be in reverse order from the way they are listed in the json files. I guess that is why the for loop is going in reverse order as well. I think this has to do with the draw order of the layers, maybe @afogelberg  can shed some light on this. Not that it matters too much  though. 

You could also reverse it later in the legend file, changing this https://github.com/origo-map/origo/blob/65aa20c6269176970f5751d0b390357586f9efac/src/legend.js#L190 line to 

```javascript
layers.reverse().forEach(function(layer) {
...
```

But that is not very readable. I think your code suggestion might be the best one @tonnyandersson, or is your second option of changing it in the viewer a better one..?

tonnyandersson:
@markus659 

I think my first suggestion works better performance wise. You don't always need the array to be in the same order as the json files and the reversal procedure does require some effort. Not a lot, though, but reversing the array only when needed will probably save some resources, especially if the map contains hundreds of layers.

afogelberg:
We can avoid reversing the array by using prepend instead of append. I think we should avoid reversing getLayers in viewer as you say @tonnyandersson . Actually I think the getLayers method should be deprecated as we could use getLayers from the map directly. But that will require some changes since a collection will be returned and not an array.

tonnyandersson:
Good point(s), @afogelberg.

markus659:
I did try to use prepend, but that caused some layers to end up in the wrong group (particularly the background layers) in my limited testing.

afogelberg:
Your're right @markus659 . The problem is that the layer is prepended before the header. We can use prepend for the map legend but for the menu legend we have to use after the header:
```
$('#o-group-' + layer.get('group') + ' .o-legend-header').after(item);
```

markus659:
Added a new commit, would be great if someone could take a look and maybe try the code to verify that it works.

afogelberg:
Looks good. Ready to merge

PR #143, opened 2016-11-02
Added scale-control

### LINKED ISSUE #85: New control show scale in text
I plan to start developing a new control in Origo that will show the map scale in text. The plan is to show scale text after zoom level has changed and the text will be shown in lower left corner in the footer. This space could work as a small console window for other controls as well.

afogelberg:
This PR fixes #85 and adds a scale control. The left part of the footer is used as console to display the scale text when zoom level is changed. The console is added as part of the viewer template and could be used by other controls as well, for example to display map center coordinates. For this purpose a method to get the the console id is added to the Viewer.


jokd:
Works well (when altering the duplicated home control in index.json to scale).
Simple and usable control which can be merged imo.

afogelberg:
Thanks for the review @jokd I will go ahead and merge. I have not added the scale-control as a default control

Issue #69, opened 2016-09-02
Enable snapping


afogelberg:
Enable snapping should be an option for the editor control.

afogelberg:
Fixed with d56f54ea929f17dc927bcdb50180d11d8f75ada4

Issue #148, opened 2016-11-07
enableInteraction is triggered several times in Editor control


afogelberg:

afogelberg:
Fixed with d56f54ea929f17dc927bcdb50180d11d8f75ada4

Issue #145, opened 2016-11-03
Cancel current edit


afogelberg:
It should be possible to cancel ongoing edit of geometry or attribute. This could be done by adding an state to these tools in the toolbar. To cancel active tool the user could deactivate it by pressing it. The state could be displayed by shifting color on the icon.

afogelberg:
Fixed with d56f54ea929f17dc927bcdb50180d11d8f75ada4

Issue #144, opened 2016-11-03
Change transactionhandler error strings to something more generic


markus659:
Currently the error strings here:
https://github.com/origo-map/origo/blob/master/src/transactionhandler.js#L147

and here:
https://github.com/origo-map/origo/blob/master/src/transactionhandler.js#L208

says 
```javascript
alert('Error saving this feature to GeoServer.<br><br>'
            + errorMsg);
```

Maybe we could change them to something more generic, e.g. "Error saving this feature to the server...".

afogelberg:
Fixed with d56f54ea929f17dc927bcdb50180d11d8f75ada4

PR #149, opened 2016-11-07
Added state to edit draw and other fixes

afogelberg:
This PR fixes several problems with Editor #148 #145 #144 . It also cleans up the control and makes it compatible to Airbnb style guide. Sorry to inlcude many fixes in one commit but otherwise there would be many PR:s.


afogelberg:
This PR now also fixes #69 and adds snap support. By default snap is enabled for the chosen edit layer. By adding editLayers snappable layers can be set.

afogelberg:
Also made the selection layer in featureinfo snapable. This way it is possible to snap to selected feature from search result.

markus659:
I tried to get the snap working but failed. First I tried without specifying snap or snapLayers. The code then stops at https://github.com/origo-map/origo/blob/cd79a087d1b45228c915e101982e35f211f578e3/src/transactionhandler.js#L462
because 'layers' is undefined. 

The function is called from https://github.com/origo-map/origo/blob/cd79a087d1b45228c915e101982e35f211f578e3/src/transactionhandler.js#L79, I don't know how the syntax works, does it set the snapSources variable if the getSnapSources function returns a value, else to editSource? Either way, the getSnapSources method at line 462 is called, which then fails (error message says "Cannot read property 'map' of undefined"), since no snapLayers have been defined.

If I change the code to 


```javascript
var snapSources = [editSource];
```` 
so that it sets the snap to the active edit layer, it works OK.


If I debug the code and look at the code where it sets the snap and snapLayers options in editortoolbar, it doesn't seem quite right, I'll include a screenshot as it's easier to understand than me explaining it in text. 

options

So bascially the options/configurations that are set in editor.json are not available in "options".

I downloaded the branch for this PR, and added my own layer in editor.json so that I could try editing.

@afogelberg maybe you could give a hint, am I missing something?

**Comment by afogelberg** [comment]
Thanks for the report @markus659 . The default state is that snap is true and that snap layers are read from editableLayers if snapLayers is not definied. But there was a small bug in https://github.com/origo-map/origo/blob/editor-fix/src/transactionhandler.js#L78 which caused problems if snapLayers is not defined . This is fixed now.

**Comment by markus659** [comment]
If I leave everything default the snap works now. However, I'm still having problems with snapLayers. I think it's the problem I tried to describe with the screenshot above (these lines: https://github.com/origo-map/origo/blob/8c1790b02bdba557a109a866eb6be229b2af69bc/src/editortoolbar.js#L179-L180). It tries to read the options/properties from "options", but it does not contain any proprieties, only the names of the editable layers. This is a problem for the snap property as well, as it always will be true.

**Comment by afogelberg** [comment]
@markus659 I can't repeat the problem. I can set snapLayers and snap. Are you setting the editor options like this:
```json
    {
      "name": "editor",
      "options": {
        "editableLayers": ["test_ytor", "test_punkt"],
        "snapLayers": ["test_ytor"],
        "snap": true,
        "isActive": true
      }
```

I did found a problem when editor is not active from start and snap is set to true but without snapLayers. In this case snap was set to false. This is fixed with the latest commit.

markus659:
Then the fault was on my side, I configured it on a per layer basis. I thought snapLayers meant which layers the active editLayer should snap against, and not which layers had snap enabled, sorry about the confusion.  It does work fine when you configure it correctly. The state functionality seems to be working as well. 

I checked the style for editortoolbar, transactionhandler and editor. A few notes:

There's an extra white space at the beginning of the row here: https://github.com/origo-map/origo/blob/9a976d7995021b5cb08ef823530073e1bb408184/src/transactionhandler.js#L5 

I don't know if it necessarily breaks the guide lines rules, but just to be consistent with the other files we might want to remove it. 

A new line is expected before one line comments: https://github.com/origo-map/origo/blob/9a976d7995021b5cb08ef823530073e1bb408184/src/editortoolbar.js#L163

=== should be used instead of ==, I don't know if this breaks the code so maybe it's not possible to change:
https://github.com/origo-map/origo/blob/9a976d7995021b5cb08ef823530073e1bb408184/src/transactionhandler.js#L119

https://github.com/origo-map/origo/blob/9a976d7995021b5cb08ef823530073e1bb408184/src/transactionhandler.js#L291

https://github.com/origo-map/origo/blob/9a976d7995021b5cb08ef823530073e1bb408184/src/transactionhandler.js#L334

https://github.com/origo-map/origo/blob/9a976d7995021b5cb08ef823530073e1bb408184/src/transactionhandler.js#L428

afogelberg:
I made the changes you suggested @markus659 . As you say it's good to be consistent. It's better to use === than == and if == breaks anything I would say it's a bug. I know we have some other == left in Origo that should be corrected as well.

markus659:
Alright, then I believe you can merge this.

Issue #129, opened 2016-10-26
Add subgroups to groups in menu legend


afogelberg:
Add subgroups to groups in menu legend with the option to tick and untick all layers in a subgroup.

afogelberg:
Fixed by #150

PR #150, opened 2016-11-09
Legend subgroups

tonnyandersson:
Add functionality for subgroups in legend. Solves #129.


jokd:
Great work @tonnyandersson. Just lacking the documentation.
Mergeable to me, any comments @afogelberg @markus659

afogelberg:
Thanks for this nice contribution @tonnyandersson . I was thinking if we really need to introduce the new concept of subgroups. I would suggest that we only use groups to nest groups, for example like this:
```json
"groups" : [
  {
    "name": "group_1",
    "title": "Grupp 1",
    "groups": [
      {
         "name": "subgroup_1",
         "title": "Undergrupp 1"
      }
    ]
  }
]
```

This way it will be easier to define the group a layer belongs to as it wouldn't be necessary to tell which parent a subgroup should be in.

tonnyandersson:
I'm all for it, @afogelberg. But, without the concept of subgroups, there is no way of telling whether a layer is part of a top level group or nested group. Since nested groups have some added functionality compared to top level groups, certain functions depend on this information. We just need to find another way to implement this added functionality.

afogelberg:
@tonnyandersson how about checking if a group has a parent? If it hasn't it's top level.

tonnyandersson:
Yes, that could be a way to solve some of the issues. There are issues regarding styling as well, but I'm working on it as we speak.

tonnyandersson:
The concept of subgroups is now removed. Subgroups are now defined as groups within groups, as suggested by @afogelberg above. I have removed related dependencies and redundant methods, but there is always room for further improvement.

markus659:
Thanks for the PR. A question about the showInfoText function, couldn't it be improved so that it doesn't have to run the first if condition (which makes it look for an layer with the name of the incoming argument,  )? Maybe some extra parameter could be added so that the function knows right away if it is a layer or a sub-group info text button that has been clicked? The css-classes differ from sub-groups and layers for instance. 

I'm not sure if this is something that is easily implemented or not, just something that struck me when looking through the code.

tonnyandersson:
@markus659 I'll look into it.

afogelberg:
Nice work @tonnyandersson on this. Adding subgroups works really great.

We need to add a couple of missing ol dependencies to the ol build files:
```
    "ol.Map#getLayers",
    "ol.Collection#getArray"
```
A small detail. The name of infotext, could we rename it to abstract as in GeoServer?

tonnyandersson:
Made changes according to comments. Had quite a few conflicts to resolve. I think it should be alright now.

Need some input on one thing, though. Any suggestions for the best way to get *all* groups, sub and top, at once? I can think of a few ways, all with pros and cons.

1. Add getAllGroups() to viewer.js - simply a function returning the concatenated results from getGroups() and getSubgroups(). Easy to implement but feels somewhat redundant with three functions doing almost the same thing.
```javascript
function getAllGroups() {
  return getGroups().concat(getSubgroups());
}
```
2. Keep getGroups() and getSubgroups() and concatenate the returns when needed. Also very easy to implement but perhaps not the most obvious and readable solution.
```javascript
var allGroups = viewer.getGroups().concat(viewer.getSubgroups());
```
3. One getGroups() function returning *all* groups. The aesthetically best looking option since it requires only one function, but also the least flexible. Will require a property to tell top level groups from subgroups (e.g. for legend creation etc).

4. Same as above but with a parameter - *top*, *sub* or *all*. No extra property needed. Basically all three functions from 1. jammed into one. Not pretty.

Thoughts?

afogelberg:
@tonnyandersson I like alternative 3 and 4 most because as you say we will have one public method only. I suggest calling getGroups() will return all groups. Then the method could have an optional filter to return either top parents or childs. The public method could be implemented with private methods.

tonnyandersson:
Something like this? Now, I can't think of a lot of situations that would require getting the subgroups only, so perhaps the *else if (opt == 'sub')* part is a bit redundant. But logically, it makes sense. 

```javascript
function getGroups(opt) {
  if(opt == 'top') {
    return settings.groups;
  } else if (opt == 'sub') {
    return getSubgroups();
  } else {
    return settings.groups.concat(getSubgroups());
  }
}

function getSubgroups() {
  [...]
  return subgroups;
}
```

afogelberg:
Looks good @tonnyandersson

afogelberg:
Thank you for the continuous work on this @tonnyandersson . I think this is ready to squash and merge @jokd @markus659 .

markus659:
I will give it an extra look later today, and let you know if I find anything. I agree with @afogelberg, great work with this PR @tonnyandersson.

markus659:
I also think this is ready now, @jokd if you don't have any further comments, can you go ahead and do a squash and merge?

jokd:
I'm sure it's great. Won't be at work until next week though so go ahead and merge it and enjoy the power of subgroups ;)

Issue #67, opened 2016-08-29
Introduce Sass for modular CSS


afogelberg:
I believe we should introduce Sass in Origo for modular CSS. That way we could have separate builds of CSS for different use cases of Origo. For example I would like to make light builds of JS and CSS for simple embedded maps. What do you think @markus659  and @jokd ?

markus659:
Yes, modularity is often a good thing. Using Sass would also help with making the css code more compact/fewer lines, right? As of right now, the css isn't too optimized and using something like Sass would help with that. I'm not sure how much work it would involve though, but I'm guessing it's worth it.

afogelberg:
Fixed with #156

PR #156, opened 2016-12-05
Introduced SASS

afogelberg:
This PR introduces SASS solving #67 . These are the changes:
* Two dev modules are added, npm-sass and npm-run-all. The latter is used to watch changes both in js and scss. This means that by running ```npm start``` both will be watched. Changes in scss-files will be watched directly without reload.
* The old style.css is split into scss-files. External libraries are imported as separate scss-files. The origo.scss-file is the main file.

The main purpose of this PR is to make the css more modularized and make it easier to maintain. It will now be really easy to import external libraries such as bootstrap.

There are still much that can be improved by for example variables and mixins. But I think we should handle those changes in separate PR:s.

A final note. Windows users might have a dependency to Visual studio to compile scss. I havn't made any tests on a computer without it.


afogelberg:
One more thing I should mention. The reset is switched to normalize.css which is more elegant than meyers.

markus659:
Correct me if I'm wrong, but right now you would have to do:
npm run build-sass
npm run build-ol
npm run build

to get all the build files? Could we instead change it to something like:

````
    "build": "npm run build-sass && npm run build-ol && npm run build-origo",
    "build-origo": "browserify origo.js --s origo | uglifyjs --compress --mangle --output dist/origo.min.js",
    "build-sass": "node-sass --include-path scss scss/origo.scss dist/css/style.css",
    "build-ol": "node node_modules/openlayers/tasks/build.js tasks/build-ol-standard.json node_modules/openlayers/dist/ol.js"
````

afogelberg:
@markus659 yes three builds are required for a complete build. I agree it would be helpful to have a build script like you suggest, at least for sass and origo. I'm not sure it's desirable to also include the ol build. If ol is not already built during development then you might get missing ol dependencies during the build. Before ol is built for the first time you get access to the whole ol library which can be misleading.

markus659:
Yes, that is true. I guess you also shouldn't have to run the build commands too often, but it would be nice to streamline it some more (less commands to remember).

I noticed the footer divs have been reworked a bit, for the better. But wouldn't this break the scale control (unless you specify the consoleId?), right now it defaults to 'o-console' in the viewer if it is not set.

jokd:
I have tested a little and I think that great functionality have been added. in this PR. Dont know much about sass but I cant see why we shouldnt merge this unless @markus659 comment about the footer breaks something.

afogelberg:
Thanks for the feedback @jokd . I will take a look at the scale control and also make it easier to run the build as @markus659 suggested.

afogelberg:
I have fixed the scale and footer problem and will now go ahead and merge.

afogelberg:
I also combined build of sass and origo as @markus659 suggested.

Issue #160, opened 2016-12-09
About control to show map information


afogelberg:
It would be useful to have the possibility to show general information about the map for example contact information. Preferable the information is opened from the map menu.

jokd:
@tonnyandersson has done exactly this, right? ;)

tonnyandersson:
@jokd @afogelberg Pretty much, yes. But it was a quick and dirty deal. It is activated from the map menu and uses the Modal control to present a simple message (title, content). Lots of room for improvement.

afogelberg:
Sounds like a good start @tonnyandersson . Looking forward to see a PR for this.

afogelberg:
Fixed with 6ae3bbd42db198b48c7965062135fb37047441a0. Thank you @tonnyandersson for the contribution

afogelberg:
@tonnyandersson we would also need some documentation for the new control.

tonnyandersson:
@afogelberg Of course!

Issue #125, opened 2016-10-26
Toolbar for controls


afogelberg:
Since the merge of #90 we need a toolbar to add controls in nicer way, As it is now the placement of controls is dependent on the absolute position in style which will be easily broken if controls are removed,

PR #163, opened 2016-12-14
Added about control

tonnyandersson:
Added control to show general information about the map. Solves #160.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1629,6 +1629,11 @@ input.o-search-field::-ms-clear {
 	height: 1em;
 	width: 1em;
 }
+.o-icon-fa-info-circle {
+	padding: .25em;
+	height: 1em;
+	width: 1em;
+}
 #o-editor-toolbar .o-icon-fa-times {
 	height: 1rem;
 	padding: 0;
--- a/index.json
+++ b/index.json
@@ -16,6 +16,14 @@
       "name": "print"
     },
     {
+      "name": "about",
+      "options": {
+        "buttontext": "About",
+        "title": "About this map",
+        "content": "<p>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed et scelerisque lectus. Phasellus ultricies diam in sem aliquam tincidunt. Email: <a href='mailto:lorem@ipsum.com?Subject=About map' target='_top'>lorem@ipsum.com</a></p>"
+      }
+    },
+    {
       "name": "legend"
     },
     {
--- a/index.json
+++ b/index.json
@@ -16,7 +16,12 @@
       "name": "print"
     },
     {
-      "name": "about"
+      "name": "about",
+      "options": {
+        "buttontext": "Om Origo",
+        "title": "Om Origo",
+        "content": "<p>Origo är ett ramverk för webbkartor. Ramverket bygger på JavaScript-biblioteket OpenLayers 3. Du kan använda Origo för att skapa egna webbaserade kartapplikationer.</p><br><p>Projektet drivs och underhålls av ett antal svenska kommuner. Besök gärna <a href='https://github.com/origo-map/origo' target='_blank'>Origo på GitHub</a> för mer information.</p>"
+      }
     },
     {
       "name": "legend"
--- a/index.json
+++ b/index.json
@@ -16,12 +16,7 @@
       "name": "print"
     },
     {
-      "name": "about",
-      "options": {
-        "buttontext": "About",
-        "title": "About this map",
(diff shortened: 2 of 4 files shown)


tonnyandersson:
Very simple implementation of the about control. Lots of room for improvement regarding functionality. The message content is defined in the index.json file. The current setup requires HTML tags in the message text string. I suggest we add some level of automated tagging, especially for email and urls. Any thoughts on this?

markus659:
Good job getting this PR up so soon after the issue was posted, a needed addition to Origo!

Maybe the information could be presented in the map menu instead of in a modal window. Something similar to how the layer groups expand, but maybe with a smoother transition. Also, the text strings should be in swedish I think. And maybe we could set a default text string for buttontext and title in the about.js file, and leave content empty? Meaning there would be no options set in the index.json file (the user would have to set them), so that it would look a bit more tidy and more like the other controls. 

Perhaps the map menu could be reworked so that the print (when it has settings available) and about control  are not using modal windows, but rather are built into the map menu. "Share map" could have its own button, like the measure tool. Take a look at this map for some inspiration: http://mapweb.karlshamn.se/mapguide/fusion/templates/mapguide/gsViewer/index.html?ApplicationDefinition=Library://BUP.ApplicationDefinition&locale=en

However, a possible redesign of the map menu is a different topic and not something that should be changed with this PR.

Some final notes regarding the style guide (about.js):

I'm not sure if we are suppose to use tabs or spaces, but on line 17, 34, 35 and 40-42 it looks like tab indentation, instead of spaces. Also jquery variables should be prefixed with a '$'.

tonnyandersson:
@markus659 I will implement the default texts for title and buttontext.

I'm not sure using the map menu instead of the modal screen for presentation is a good idea. I think it would either push the map layers and groups down or slide over them in a weird way, depending on the implementation. I we're not going modal, I suggest using the sidebar function instead. A redesign of the map menu would put things in a different light, but, as you pointed out, that's a different topic.

On another note, I'm on the fence regarding my own comment about automated HTML tagging. It would make the configuration easier (as long as you think inside the box) but much less customizable.

markus659:
I don't think it would look any weirder than it does today when you expand a layer group (I'm thinking it would be implemented and work in basically the same way as the layer groups do). But the sidebar function  might be a good way to go right now, any thoughts on modal vs. sidebar for this control @jokd @afogelberg?

I haven't used handlebars for any customisation, but from my understanding you explicitly need to write html tags, so I don't think we need to have automated tags for this control.

jokd:
If you go modal you can pop it on load as an introduction to the user. 
Beside that, either way works for me.

afogelberg:
Good work with the about control @tonnyandersson . I will take a closer look later today.

Regarding the mapmenu discussion. I don't think we should rework the mapmenu and try to fit in content of various types, for several reasons:
* we will always have the same width constraint
* we will have less flexibility by introducing the principle that control information and options should be placed in the menu
* reading text from a sidebar is hard
* it will not be a good user experience by introducing some sort of page navigation with different functionalities in the sidebar. Sense of control be lost.

I think it's important for user experience to keep the map menu as static, clean and predictable as possible. Actually, in the long run I'm not convinced that the layer manager should be in the map menu either, at least not in desktop layout.

Navigation and map tools should be easily accessible with buttons and grouped. But functions that you don't need while interacting with the map, such as print and share, should not be highlighted and are well suited to be hidden in the menu. We should continue to work on #125 and also have some space separation between the navigation button group and the map tools.

afogelberg:
I agree with @markus659 that we don't need automated tags. We can't use handlebars for now as it needs to be precompiled which would be problematic in this case. We could add the option to compile handlebars on runtime but that would require that part from the handlebars library. But maybe it's worth the cost in extra file size?

If we have the about control as a default control I think we should add some default option values for title and content. Maybe some info about Origo?

tonnyandersson:
@afogelberg Like, "Om Origo" as button text and title, with a short description of Origo as content? Any suggestions for the content?

afogelberg:
How about our readme-text for Origo:

> Origo is a web mapping framework. It is based on the OpenLayers 3 library. You can use Origo to create your own desktop-like web mapping applications. The project is run and maintained by a number of Swedish municipalities.

tonnyandersson:
In swedish:

"Origo är ett ramverk för webbkartor. Ramverket bygger på JavaScript-biblioteket OpenLayers 3. Du kan använda Origo för att skapa egna webbaserade kartapplikationer.</p><p>Projektet drivs och underhålls av ett antal svenska kommuner. Besök gärna <a href='https://github.com/origo-map/origo' target='_blank'>Origo på GitHub</a> för mer information."

Good to go?

afogelberg:
@tonnyandersson looks good to me

Issue #164, opened 2016-12-15
Calculate geometries in featureinfo


afogelberg:
I have started to work on a feature to enable geometry calculations in featureinfo. The aim is to be able to present different geometry calculations when identifying a feature. This can for example be used to build a route url that can be used in a another application such as hitta.se. But it could also be used to show coordinates in various projections. A routing link could for example be built like this:
```javascript
attributes: [
  {
    "html": "<a href='http://the.route.service?route={{@center(EPSG:4326)}},{{@position(EPSG:4326)}}' target='_blank'>Hitta hit</a>"
  }
]
```

afogelberg:
Fixed with c929f59c34b95a41aab5aaaf995efae6ac54c417

PR #168, opened 2016-12-22
Featureinfogeometry

afogelberg:
This PR solves #164 . It will now be possible to include geometry functions in featureinfo. To begin with we get a function to calculate center coordinates for a feature. This is for example useful to start a routing service with a destination coordinate. For this case the attribute in featureinfo could be defined like this:
```javascript
"attributes": [
  {
    "html": "<a href='https://www.google.se/maps/dir//{{@center(EPSG:4326,reverse)}}' target='_blank'>Hitta hit</a>"
  }
]
```

The new part is `{{@center(EPSG:4326,reverse)}` which is a function that takes two argumets, projection and axis orientation.

After this PR we can easily add new geometry functions such as getAreas,


afogelberg:
Also added an area method for polygons and multipolygons.

afogelberg:
@jokd gave green light on this so I will go ahead and merge

Issue #133, opened 2016-10-26
Link control in menu


afogelberg:
Link to external page that opens from the menu.

markus659:
Fixed with https://github.com/origo-map/origo/commit/6bb7a859c3e301673e726032dd98cfb41e415d7b.

PR #171, opened 2017-01-11
Added control for external link

mulfvik:
Issue [#133](https://github.com/origo-map/origo/issues/133)


afogelberg:
This is a good control to have in Origo. A couple of notes:
* there should be a render function as for example in the [home control](https://github.com/origo-map/origo/blob/master/src/home.js#L25-L33)
* the name of control. Do we need the external part of the name? I would prefer to call it just link.

afogelberg:
Thanks for the contribution @mulfvik . Viewer is not used in link and should be removed from link.js. Otherwise I think it's ready to merge @jokd @markus659 .

jokd:
Looks good and works well beside the already mentioned unnecessary viewer.js-require and that the other buttons (share/print/about) are triggered when clicking either text or symbol while the link is only triggered when clicking the text. No biggie though.

afogelberg:
As you say @jokd there's a difference comparing to the other buttons. I think it would be good to be consistent and having the same behavior for all menu buttons. Would it be ok @mulfvik to change the menu button to be similar to the other menu buttons, for example print?

Strangnaskommun:
I agree. They should behave the same, I had'nt noticed. Will do changes after notes. Having (private ;) issues with git and this branch locally, long story, so I will get to that as soon as I get my head around that. Thanks for the input.

mulfvik:
I agree. They should behave the same, I had'nt noticed. Will do changes after notes. Having (private ;) issues with git and this branch locally, long story, so I will get to that as soon as I get my head around that. Thanks for the input.

mulfvik:
I believe it's good to go..

afogelberg:
@mulfvik changes look good. But I think the link should be opened in a new window. You could use window.open(url) instead of location.href.

afogelberg:
I think it's ok to merge @jokd  @markus659

jokd:
Looks good to me as well.

markus659:
@mulfvik can you resolve the conflict in style.css so that we can go ahead and merge?

PR #174, opened 2017-01-13
Removed wrong this

### LINKED ISSUE #141: Geoposition does not update its position
The Geoposition does not update its position during a session. It remains locked at its initial location. Only by reloading the page and activating Geoposition again one can achieve an updated location. A live, continuous location update does not work. This occurs in devices such as Samsung, Iphone and Windows tablets and their corresponding browsers. The site, http://karta.sigtuna.se, is secured with https.

afogelberg:
This PR fixes #141 . Also made some code improvements according our style guide.

### DIFF
--- a/src/geoposition.js
+++ b/src/geoposition.js
@@ -29,10 +29,6 @@ function init() {
   $geolocateButtonId = $('#o-geolocation-button');
   $geolocateButton = $('#o-geolocation-button button');
 
-
-  var src = baseUrl + 'img/geolocation_marker.png';
-  var markerImg = '<img id="o-geolocation_marker" src="' + src + '"/>';
-  $('#o-map').prepend(markerImg);
   markerEl = $('#o-geolocation_marker').get(0);
   marker = new ol.Overlay({
     positioning: 'center-center',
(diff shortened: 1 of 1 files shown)


markus659:
Unfortunately this didn't fix the problem for me described in the referenced issue. I tried it with firefox on android, and while this fixes the console error, the position is still not updated. I still think should be merged, as this is still an improvement of the code, and maybe it will fix the problem for some of the other users experiencing problems with the position not updating.

I also got an error in the console saying

````
ReferenceError: assignment to undeclared variable heading
````
which is pretty clear, heading is never declared. Some other points about the style guide:

variable declarations should be at the top of the scope,
https://github.com/origo-map/origo/blob/030f22ff63ccf0310bf43d82ef16bb54ab403450/src/geoposition.js#L33-L34

https://github.com/origo-map/origo/blob/030f22ff63ccf0310bf43d82ef16bb54ab403450/src/geoposition.js#L142
 

Single quotes for strings:

https://github.com/origo-map/origo/blob/030f22ff63ccf0310bf43d82ef16bb54ab403450/src/geoposition.js#L58

'view' is never used:

https://github.com/origo-map/origo/blob/030f22ff63ccf0310bf43d82ef16bb54ab403450/src/geoposition.js#L155

=== instead of ==:

https://github.com/origo-map/origo/blob/030f22ff63ccf0310bf43d82ef16bb54ab403450/src/geoposition.js#L156


'getCenterWithHeading' is never used:

https://github.com/origo-map/origo/blob/030f22ff63ccf0310bf43d82ef16bb54ab403450/src/geoposition.js#L174

afogelberg:
Thanks for checking @markus659 . I made the style guide fixes and I will go ahead and merge. I will create an example that can used for further tests.

Issue #177, opened 2017-01-17
Remove layer sources from viewer


afogelberg:
Layer sources should be created in separate source files and not as part of the viewer. This will make viewer more intuitive and more modularized.

afogelberg:
Fixed with #180

PR #180, opened 2017-01-30
Separated layer creation from viewer

afogelberg:
This PR solves #177 and separates layer creation from the viewer. This will make development of layers much easier.


jokd:
I edited some options names that were incorrect.
Other than that it looks good and works well.

afogelberg:
Thanks @jokd . I will go ahead and merge then.

Issue #186, opened 2017-02-03
Change the base map in examples


afogelberg:
We should change the base maps that are used in the examples so we don't use the base maps from Mälardalskartan. In the mean time we have to change the Mälardalskartan url:s to use https.

afogelberg:
Switch to https was made with d048155b39d53277dae34fc279e4c3f15dc0be4b

jokd:
Fixed with #188

Issue #73, opened 2016-09-05
Add osm as source


afogelberg:
Since mapquest support has been deprecated in OpenLayers we need to have some other free map service in Origo.

jokd:
Fixed with #188

PR #188, opened 2017-02-09
Added OSM and extended the mask

jokd:
Created new mask and updated index.json when switching source for the
basemap to OSM.
Solves #73 and #186


afogelberg:
Great and fast work @jokd . Please go ahead and merge. Resolutions are not necessary to include but I think we could keep them anyway. Do you update the documentation @jokd ?

jokd:
Shall I alter editor.json as well? I forgot about that file.

afogelberg:
I think we can keep the editor as it is for now. We can change the editor file when we make other improvements. It would be good to have a working editor example with a public editable service.

PR #190, opened 2017-02-15
Removed bloodhound dependency

### LINKED ISSUE #152: Remove bloodhound
We can remove the dependency to bloodhound in search by replacing the source in typeahead with an ajax request. This will make the code clearer and makes it easier to configure.

markus659:
This will fix #152. 

The PR removes bloodhound as a dependency for the search and replaces it with an ajax request. Changes has also been made so that search.js more closely follows the style guide. Most of the changes for the style guide was in the selectHandler function, but I don't think it should break anything, I would however appreciate an extra look on that function specifically.

### DIFF
--- a/src/search.js
+++ b/src/search.js
@@ -7,58 +7,60 @@
 var ol = require('openlayers');
 var $ = require('jquery');
 var Viewer = require('./viewer');
-var wktToFeature = require('./maputils')['wktToFeature'];
+var wktToFeature = require('./maputils').wktToFeature;
 var Popup = require('./popup');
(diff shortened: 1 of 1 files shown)


afogelberg:
Looks great @jokd . This was exactly what I had in mind, Please Squash and merge.

afogelberg:
Sorry @markus659 . I see that I wrote the wrong name. Great work on this @markus659 !

markus659:
No problem, it's easy to mix up user names. Thank you for the quick review.

Issue #191, opened 2017-02-16
Option to save edit manually


afogelberg:
With a save button in the editor toolbar we can enhance the save options in several ways. We can use it to have an option to either autosave edits or require edits to be be saved manually. Even if edits are autosaved it's good to have it as a fallback if something goes wrong during save. The you have the chance to save the edits again.

afogelberg:
Fixed with #192

PR #192, opened 2017-02-16
Save edits

afogelberg:
This PR solves #191 . In order to solve #191 the logic for saving has been reworked:
* All edits are stored in a temporary store until they are saved to server. This gives the option to save manually or autosave.
* The transaction handler has been separated to only handle server transactions. This will also make it easier to add other edit sources besides wfs.
* The editortoolbar always have the save button. The button is disabled if all edits are saved to server. If edits are not saved the save button is activated. Should the save fail the save button will still be activated.

A new option autoSave has been added to the editor. It can be true or false.


src/editor/editform.js
```diff
@@ -0,0 +1,44 @@
...
+      break;
+    case 'textarea':
+      el = '<div><label>' + label + '</label><br><textarea id="' + id + '"' + maxLength + 'rows="3">' + val + '</textarea></div>';
+      break;
+    case 'checkbox':
+      var checked = val === true ? ' checked' : '';
+      el = '<div class="o-form-checkbox"><label>' + label + '</label><input type="checkbox" id="' + id + '" value="' + val + '"' + checked + '></div>';
+      break;
+    case 'dropdown':
+      var firstOption;
```
markus659: variables should be assign at the top of their scope

src/editor/editform.js
```diff
@@ -0,0 +1,44 @@
...
+      if (val) {
+        firstOption = '<option value="' + val + '" selected>' + val + '</option>';
+      } else {
+        firstOption = '<option value="" selected>Välj</option>';
+      }
+      el = '<div class="' + cls + '"><label>' + label + '</label><br><select id=' + id + '>' + firstOption;
+      for (var i = 0; i < dropdownOptions.length; i++) {
```
markus659: variables should be assign at the top of their scope

src/editor/edithandler.js
```diff
@@ -0,0 +1,437 @@
...
+  } else {
+    return true;
+  }
+}
+
+function onDeleteSelected() {
+  var features = select.getFeatures();
+  if (features.getLength() === 1) {
+    var feature = features.item(0);
+    var r = confirm("Är du säker på att du vill ta bort det här objektet?");
```
markus659: single quotes for strings

src/editor/edithandler.js
```diff
@@ -0,0 +1,437 @@
...
+
+  //Get attributes from selected feature and fill DOM elements with the values
+  var features = select.getFeatures();
+  if (features.getLength() === 1) {
+    dispatcher.emitChangeEdit('attribute', true);
+    var feature = features.item(0);
+    if (attributes.length > 0) {
+
+      //Create an array of defined attributes and corresponding values from selected feature
+      var attributeObjects = attributes.map(function(attributeObject) {
```
markus659: variables should be assign at the top of their scope

src/editor/edithandler.js
```diff
@@ -0,0 +1,437 @@
...
+          if (constraintProps.length === 3) {
+            obj.eventType = constraintProps[0];
+            obj.dependencyVal = feature.get(constraintProps[1]);
+            obj.requiredVal = constraintProps[2];
+            obj.dependencyVal === obj.requiredVal ? obj.isVisible = true : obj.isVisible = false;
+            obj.addListener = addListener();
+            obj.elId = '#input-' + obj.name + '-' + obj.requiredVal;
+            obj.elDependencyId = '#input-' + constraintProps[1];
+          } else {
+            alert('Constraint properties are not written correct, it should written as for example change:attribute:value');
```
markus659: alert message should perhaps be in Swedish?

src/editor/editsstore.js
```diff
@@ -0,0 +1,128 @@
+var $ = require('jquery');
+var ol = require('openlayers');
+var viewer = require('../viewer');
```
markus659: viewer is never used

src/editor/edithandler.js
```diff
@@ -0,0 +1,437 @@
...
+        layerName: currentLayer,
+        action: 'delete'
+      });
+      select.getFeatures().clear();
+      editSource.removeFeature(editSource.getFeatureById(feature.getId()));
+    }
+  }
+}
+
+function onSelectAdd(evt) {
```
markus659: onSelectAdd is never used

src/editor/editsstore.js
```diff
@@ -0,0 +1,128 @@
...
+var dispatcher = require('./editdispatcher');
+
+var editsStore = function featureStore() {
+
+  $(document).on('changeFeature', featureChange);
+
+  var edits = {};
```
markus659: variables should be assign at the top of their scope

src/editor/wfstransaction.js
```diff
@@ -0,0 +1,103 @@
...
+        });
+      }
+    } else {
+      error();
+    }
+  }
+
+  function error(e) {
+    var errorMsg = e ? (e.status + ' ' + e.statusText) : "";
+    alert('Det inträffade ett fel när ändringarna skulle sparas till servern...<br><br>' +
```
markus659: <br><br> does display in full text when you get the error message. I think we should replaced it with newline (\n): 
'Det inträffade ett fel när ändringarna skulle sparas till servern...\n'

markus659:
Just to add some extra thoughts to my review:
 

- I don't remember if this has been discussed previously, but I think we should be able to change the background map of the editor to OSM as well?

- I noticied there is a "disabled" class for the floppy icon, but I'm not sure what the difference is? I didn't notice any difference, which made it difficult to tell if I could/should use the button or not. 

- Maybe some extra confirmation once you saved the changes would be good as well? Like some text in the footer console or something else.

Also just a general question, what is the purpose of the uuid?

Great work with these improvements for the editor. :smile:

afogelberg:
Thanks for the review @markus659 . I fixed the changes according the style guide which was mostly old sins. I also readded the missing disable class for the save state. Now you should be able to see a difference. As soon as there unsaved changes the save button will be activated and vice verse. 

I agree we should change the backgound map but we can handle it in another PR.

I have also considered that there should be some report of the remote save. Can we handle it in another PR? For now we have some info of the state with the save button being disabled or not.

The purpose of the uuid is to have a temporary unique id for unsaved changes. If several new features are created before save we need to keep track of the features. When the features are saved to remote the uuid is replaced by a proper id. As we now have the option to save manually there can be many unsaved inserts.

markus659:
Yes, the background map and confirmation/report can be handled separately. The disabled class worked fine for me, I think it's good to make use of it. 

I don't have any further input so you can merge if you feel it's ready.

Issue #135, opened 2016-10-26
Control to show coordinates


afogelberg:
Proposed solution:
- the coordinates are shown in the footer and are updated while panning
- a marker is shown in the center of the map while panning

afogelberg:
Fixed with #199

PR #199, opened 2017-02-28
Show coordinates

afogelberg:
This PR solves #135


afogelberg:
The control is defined like this:
```
    {
      "name": "centerposition",
      "options": {
        "title": "Pseudo-Mercator",
        "projections": {
          "EPSG:4326": "WGS84"
        }
      }
    }
```
If only the map projection is used to show coordinates it is enough to define title. Several projections can be added as long as they are defined in proj4Defs (except EPSG:3857 and EPSG:4326). Then the projections can be toggled by pressing the button next to the coordinates.

markus659:
The center marker is not displaying for me when panning the map, what could be the cause of this? 

Another solution would perhaps be to use the map pin you get when you click in the map (showing the coordinates of the pin in the console)? This way it would be easier to pin point the feature/place you want to get the coordinates from. Maybe the pin could be used for more than it is currently?

afogelberg:
> The center marker is not displaying for me when panning the map, what could be the cause of this?


I have removed the conflict. Maybe that was the problem?

> Another solution would perhaps be to use the map pin you get when you click in the map
This could be a nice enhancement. However, I think it's still valuable to get the position by panning. When panning you can adjust the position more accurately, for example on a touch screen when you are using the fingers. The center marker also has another purpose, namely to get the center of the map which can be useful when you're creating a rather small embedded map. Maybe a configurable option?

jokd:
It's working well for me. I like the mobile friendly approach. 
But perhaps the user should be able to toggle the function on/off? It might get irritating if you for instance are uninterested of the coordinates and  use a small screen.

afogelberg:
We could make the default state non active with the default button text 'Koordinater'.

jokd:
I think that's a good solution.

afogelberg:
Made improvements and included the mousePosition control that is now default. Center position can be be toggled to and there is also an new feature added, finding coordinates. When center position is active coordinates can be edited and zoomed to when enter is hit.

jokd:
I get an error when using the crosshair and Sweref99 TM-projection. 
And maybe the great zoom-to-coordinates feature should be more obvious and visible on larger touch devices as well?

afogelberg:
@jokd I forgot to add the proj definition for TM in proj4Defs.

Issue #200, opened 2017-02-28
Simplify geoposition


afogelberg:
The geoposition control can be simplified. For now only a portion of the settings are used and the animation is unnecessary. The animation function used in geoposition is furthermore deprecated in OL 4. The animation function also prevents the moveend event to dispatch which is also a problem for the centerposition control.

afogelberg:
Fixed with #201

PR #201, opened 2017-02-28
Simplify geoposition

afogelberg:
Solves #200


jokd:
Works well. 
Perhaps we should animate to the current position. In OL 3.20+ there's a nice new view.animate() method we can use.

afogelberg:
@jokd I agree that animation would be nice. Can we add it in a separate PR as this also require an update of OL?

jokd:
Sure. You can merge the PR.

Issue #196, opened 2017-02-21
Make proj4Defs not mandatory for EPSG 3857 and 4326


jokd:
You should not need to define those projections in the index.json since OpenLayers support them natively.
Documentation needs to be updated when this is solved.

jokd:
Fixed with #232

PR #210, opened 2017-03-15
Added toolbars for control target

### LINKED ISSUE #125: Toolbar for controls

afogelberg:
Fixes #125

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -468,14 +468,13 @@
   .o-map .ol-control:hover {
     background-color: rgba(255, 255, 255, 0.6); }
   .o-map .ol-zoom {
-    top: .5em;
-    left: .5em; }
+    width: 2rem; }
   .o-map .ol-rotate {
-    top: 12.5em;
-    left: .5em;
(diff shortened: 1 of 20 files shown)


jokd:
Nice improvement! Works well for me, except for the scaleline is missing. Or has it been deleted?

afogelberg:
@jokd are you using the default map with osm? The scale-line will not show with osm and how the projection is defined. Can you try a map with another projection? We should be able to solve the scale-line problem with #196 .

jokd:
I'm sure you're right. I tested with the default layers with 3857 projection. Won't be able to test other config until next week so discard my previous comment about the scale line since that's another issue.

afogelberg:
Ok to merge then?

jokd:
Its good to go. Well done!

afogelberg:
Great! Thanks for the review

PR #216, opened 2017-03-20
Replaced media query with el query

### LINKED ISSUE #205: Replace media queries
Origo uses media queries to make a responsive map which works really well when Origo is used as a full page app. However, if Origo is used as a widget for example as an embedded map on a web site then media queries aren't that effective. The problem is that media queries uses the whole viewport of the browser and not the viewport of the widget element. Unfortunately element queries are absent in present CSS although there are interesting discussions on the subject, see for example http://www.jonathantneal.com/blog/thoughts-on-media-queries-for-elements/ .

My proposal is that we remove media queries from Origo. Instead we can use javascript to listen for map resize as we do for attribution in viewer:
```
  //Check size for attribution mode
  $(window).on('resize', checkSize);
```
We can replace media queries for different screen sizes with defined classes for corresponding screen sizes.

afogelberg:
Fixes #205

### DIFF
--- a/conf/origoConfig.js
+++ b/conf/origoConfig.js
@@ -3,6 +3,13 @@ module.exports = {
   target: '#app-wrapper',
   svgSpritePath: 'css/svg/',
   svgSprites: ['fa-icons.svg', 'miscellaneous.svg', 'custom.svg'],
+  breakPoints: {
+    xs: [240, 320],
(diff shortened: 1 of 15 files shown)


jokd:
Great job! Works really well. 
Can't see why we wouldn't merge this.

PR #218, opened 2017-03-23
Edit ags

afogelberg:
Fixes #194

This PR adds ArcGIS rest as an editable source. I guess it might be hard for you @markus659 @jokd to verifiy that it works?

PR #222, opened 2017-03-31
Offline

### LINKED ISSUE #142: Create single svg sprite for non-Font Awesome icons
A single svg sprite could be used for all non-Font Awesome icons. That way, only two svg files need to be maintained. This makes general maintenance easier and facilitates implementation of occasional non-Font Awesome icons.

The non-Font Awesome icons should be named after the library they originate from, in a similar manner as the Font Awesome icons (e.g. "**fa**-home"), and also grouped accordingly.

afogelberg:
This PR fixes #142 and adds offline support to Origo. The localforage library is used to store layers with indexedDB.

Currently wfs, geojson and topojson are supported formats for offline storage. Wfs is supported as an offline editable format. ArcGIS Server rest feature will be added as an offline format as soon as #218 is merged.

The offline layers are managed with 'Nedladdningar' in the menu. A layer is treated as an offline layer if the property offline is added with the value true to the layer definition.


markus659:
I made a few style guide fixes, hope that was OK. Some other things I noted:

* Could we not move the downloader.template.handlebars file to the templates folder?
* There's an extra row break in the handlebars file, can it be removed?
* There's a conditional statement here: https://github.com/origo-map/origo/blob/e7a94d778aee56dd5cc0fbcbf4e212d9012d9a3f/src/offline/offlinestore.js#L93, but currently it doesn't do anything. Is there something missing or can it be removed/rewritten?

Functionally wise it worked great for me,

afogelberg:
> I made a few style guide fixes, hope that was OK

Thanks @markus659 

> Could we not move the downloader.template.handlebars file to the templates folder?

For consistency we should at least follow the same principle for templates. But I would say it's more clear if we keep the template file together with the rest of the control. And maybe we should move the editor template to the editor directory and the featureinfo templates to a directory called featureinfo. What do you think?

> There's an extra row break in the handlebars file, can it be removed?

Yes

>There's a conditional statement here: https://github.com/origo-map/origo/blob/e7a94d778aee56dd5cc0fbcbf4e212d9012d9a3f/src/offline/offlinestore.js#L93, but currently it doesn't do anything. Is there something missing or can it be removed/rewritten?

I had something in mind there. I can't remember exactly what for the moment but I think we should leave it for now.

markus659:
I tend to agree, it would be nice to have all the files in one folder. It would make it easier to navigate the file/folder structure. So the more long-term goal would then be to get completely rid of the templates folder?

afogelberg:
> So the more long-term goal would then be to get completely rid of the templates folder?

I would vote for getting rid of the templates directory.

markus659:
Yep, let's keep the template file where it is and work towards removing the templates folder. I believe this is ready for merging once the conflicts have been fixed.

PR #230, opened 2017-04-11
Set footer content

### LINKED ISSUE #12: Option to set footer content
Origo uses a handlebars template to generate the layout. In order to able to set content for the footer, an option is needed for that. For example to have a logo in the middle of the footer.

markus659:
This fixes #12. It makes it possible to set content for the middle footer. You can set is as a image, a link or plain text (or a combination). The configuration is done in the .json-files. I have not updated the documentation but will do so if this gets accepted.


jokd:
Nice and useful addition. Works well, I think you can merge this.

afogelberg:
Nice addition. I think it would be good if we add a render method to viewer including the setting of the footer. This way the init code will be easier to read.

markus659:
Thanks for the input. I agree, a render method would make the code easier to read. I will make the change.

markus659:
Render method added in the latest commit.

afogelberg:
Looks good @markus659 . If you just fix the conflict I believe it's ready to merge.

PR #232, opened 2017-04-18
Fix to make proj4Defs not mandatory

jokd:
Fix to make proj4Defs not mandatory when using EPSG:3857 or 4326 that's
supported natively in Openlayers.
Solves #196

### DIFF
--- a/index.json
+++ b/index.json
@@ -32,6 +32,9 @@
     },
     {
       "name": "measure"
+    },
+    {
+      "name": "scaleLine"
     }
   ],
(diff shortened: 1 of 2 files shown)


afogelberg:
Great work @jokd . One small remark only to change this for sake of readability:
`proj4.defs(proj) ? units = proj4.defs(proj).units : units = undefined;`
to:
`units = proj4.defs(proj) ? proj4.defs(proj).units : undefined;`

Issue #116, opened 2016-10-21
Upgrade to ol 4.1.0


afogelberg:

afogelberg:
Is dependent on #201

afogelberg:
Fixed with #233

PR #233, opened 2017-04-18
Upgrade to OL 4.1.0

jokd:
Removed deprecated methods. Solves #116. Required for solving other issues.


afogelberg:
Looks good @jokd . I just removed size from view.fit() which changed in 4.0.0.

Issue #158, opened 2016-12-06
Support for setting tolerance


jokd:
We should be able to set tolerance for interacting with features in the map. Points and lines can be hard to interact with on mobile devices.

afogelberg:
I agree @jokd . I believe this will be supported in the next release of ol.

afogelberg:
This feature has now landed in OL 3.20. When setting tolerance do we want the tolerance to apply to all type of geometries including polygons? My experience is that lines are hardest to hit.

jokd:
Nice spotted @afogelberg 
Perhaps the tolerance should be able to set global as well as on layer basis. Polygon layers could be excluded by default but sometimes you might have small polygon features you would like to have a certain amount of tolerance. Lines ans points will for sure benefit from this feature.
There are other nice improvements in 3.20 making it worth upgrading, but the MVT vector tiles went broken for me.

afogelberg:
Then I suggest we start with a featureinfo option to set a general tolerance. We can make further enhancements later if required.

afogelberg:
@jokd I guess we can close this?

jokd:
Yes, fixed with #234

PR #234, opened 2017-04-18
HitTolerance

jokd:
Option to set hitTolerance for features. Requires #116. Solves #158


afogelberg:
Works fine @jokd . Just a really tiny thing, could you fix the indentation for hitTolerance in featureinfo and maybe move up setting the option before adding the listeners?

PR #237, opened 2017-04-24
Added static as modal option

### LINKED ISSUE #221: Don't close modal on click outside
I some cases for example, while editing, the modal window is too easy to close by accident. One way to go is to add an option for this. The other way is to always force the user to use the close button. I would vote for the first alternative.

afogelberg:
Fixes #221


afogelberg:
As this is a really small change I will go ahead and merge it

Issue #198, opened 2017-02-21
Image upload option in editor


tonnyandersson:
An image upload option would be a nice addition to the editor.

This feature could be implemented in a number of ways. To make the feature as back-end independent as possible, I suggest a base64 string upload. This works well with WFS-T and does not require any server side scripts. This method should be compatible with most back-ends since the file is uploaded as a "simple" string.

afogelberg:
@tonnyandersson I'm going to develop a feature for image upload in the near future for Mälardalskartan. Is this something you are planning to do as well? I agree that it should be possible to do with WFS-T but I think it will be a feature in Origo server as well.

tonnyandersson:
@afogelberg I just implemented a simple image upload feature for an in-house project. Tried a php based solution as well, but settled with the more WFS-T friendly base64 string for various reasons. But I agree with you regarding an Origo server feature. Perhaps we should do both, with an option to choose method? For flexibility.

markus659:
Fixed with https://github.com/origo-map/origo/pull/246.

PR #246, opened 2017-05-03
Image upload in editor

tonnyandersson:
Adds simple function to upload images in editor. Images are stored as base64 strings. No server side scripts needed. Solves #198.


tonnyandersson:
Only basic functionality added so far. Missing ability to see (and if needed remove) previously uploaded images. Also missing options to set maxWidth or maxHeight for the stored image. Only possible to set maxSize, which applies to both width and height.

tonnyandersson:
Images selected for upload and previously uploaded images are now visible in the form. A delete button has been added.

The config options for image upload are as follows:

{
  "name": "attribute_name",
  "title": "Upload image: ",
  "type": "image",
  "maxSize": 600
}

markus659:
Thanks for the contribution. A lot of functionality in a small number of lines. I think this is an important component to the editing capabilities.. Some quick comments about the style guide, on the following lines the variable declarations should move further up 
https://github.com/origo-map/origo/blob/5ef8f54c8efa5f8c1f2ea3c34be04f1226a1e75b/src/editor/edithandler.js#L197-L199
https://github.com/origo-map/origo/blob/5ef8f54c8efa5f8c1f2ea3c34be04f1226a1e75b/src/editor/edithandler.js#L290

There should be a newline before return:
https://github.com/origo-map/origo/blob/5ef8f54c8efa5f8c1f2ea3c34be04f1226a1e75b/src/editor/edithandler.js#L306

Using iOS devices the image is not displayed correctly (wrong rotation). It seems to always default to a certain rotation, which only is correct if you took the image holding the device in that specific way. On Android it works fine. Is this something that you can look into @tonnyandersson?

tonnyandersson:
@markus659 Sure, I'll look into it!

tonnyandersson:
@markus659 Fixed code according to style guide and implemented image orientation correction. Haven't tried it on Android, though.

markus659:
The fix seems to work fine. A few variables that can be declared earlier in the getimageorientation file (at lines 9,11, 15, 17 and in the for loop). If you could, try and replace the for-loop, see https://github.com/airbnb/javascript/tree/es5-deprecated#iterators-and-generators. I know there are a number of places still in the source code that uses for-loops, so if you feel like you don't have the time to do it, it's OK.

Once the style guide changes are made and the conflict is resolved, it will be ready for merging :-)

tonnyandersson:
@markus659 Fixed variable declarations. The for-loop, on the other hand, has to stay. There is no array involved there so I can't use forEach() or every() or any such function.

markus659:
Alright, good job with this PR. I will merge it.

Issue #247, opened 2017-05-03
Measure tool not working with EPSG.3857


jokd:
When using the measure tool with the default coordinate system EPSG:3857 the tool shows incorrect values.

afogelberg:
Are you up for a fix @jokd ?

jokd:
I will dig into it.

PR #249, opened 2017-05-04
Changed length and area calculation

jokd:
Solves #247 
Using geodesic measures instead of projected. This will work with all
projections and testing shows approx 0,2% difference compared to
calculations with projected coordinates.


jokd:
@afogelberg Please review my changes.

afogelberg:
@jokd sorry to be picky, but I had in mind a higher order function like forEach or in this case maybe reduce.

jokd:
It gets a bit ugly since you want two values of the array but I replaced the for loop with .reduce.

Issue #126, opened 2016-10-26
Support for vector tiles


jokd:
Add support for the ol.layer.VectorTile layer type for client side rendering of high resolution basemaps.

afogelberg:
Fixed with #252

PR #252, opened 2017-05-05
Support for vector tiles

jokd:
Added support for vector tiles. Also introduced custom styling with
stylefunctions for more advanced styling that's necessary for e.g. a
basemap.
Solves #126


afogelberg:
Nice work @jokd . Looking forward to try it out, hopefully next week.

afogelberg:
Great contribution @jokd . It's really useful for other cases as well to have the option to make custom style functions. But maybe we don't need to add a special property called customStyle? We could resuse the style property but add a function that checks if the style is a custom function or not. To do that we could extend the styles object with the custom styles. Then check if the named style is a function or a an array/object.

If I could add one more thing to the wish list, it would be that the custom style is requested with an url. That way it will be easy to add and make changes without rebuilding Origo.

jokd:
@afogelberg Any idea on how to request the style with an url?

afogelberg:
@jokd how about ajax get js?

jokd:
@afogelberg Works, but how do I get the ol-classes used for styling when not being part of the build?

afogelberg:
@jokd if the fetching of the custom styles is handled in style (where the ol classes are available) together with "ordinary" styles, it should work. But I realize that can require quite a lot of work, for example when initializing the layers with these styles. Another suggestion is that we leave the url part for now and only change the customStyle to use the style property. The url could be a future enhancement if requested.

jokd:
@afogelberg Deal

jokd:
@afogelberg Please review my changes. Sorry about the messy commithistory.
The syntax for styling is:
      "basemap_bright": [
        [
          {
            "customStyle":"basemap_bright",
            "icon": {
              "src": "img/png/orto.png",
                "size": [20,20]
            }
          }
        ]
      ],
That way we can reuse the functions for legendstyling.

afogelberg:
@jokd I like the idea with the new syntax, will take a look at it. Since custom styling now will be a part of the ordinary styling, how about renaming it to just custom instead of customStyle?

jokd:
@afogelberg Makes sense, I have made changes accordingly

afogelberg:
@jokd it works very well. I have one final thought, when defining the custom style functions I think it would be easier if each style function can be defined in a separate file. All the the style functions could then be referenced in stylefunctions.js as an object with named style functions.

jokd:
@afogelberg Sounds nice. Do you want to dig into it?

afogelberg:
@jokd sure

afogelberg:
@jokd I made the changes. If you think it's ok we can go ahead and merge.

jokd:
@afogelberg Perhaps the default style should be usable for styling and not just an example of the syntax. For instance a simple style for point, line and polygon geometries?

afogelberg:
@jokd makes sense, I'll take a look at it.

afogelberg:
@jokd made a simple default style

jokd:
@afogelberg Looks good, I think we can merge this.

Issue #251, opened 2017-05-05
XYZ source needs to be more flexible


jokd:
XYZ services are requested with different syntaxes. We need to have more flexibility when configuring the layers.

afogelberg:
@jokd can we close this?

jokd:
@afogelberg Yes I believe so. Have a nice weekend ;)

afogelberg:
@jokd same to you :)

PR #253, opened 2017-05-05
Extended TMS support

jokd:
Edited the XYZ-source to support servers with different syntaxes for
TMS-services. Now you can specify the layer-part of the TMS request in
the config with the layerURL-parameter. 
Solves #251

### DIFF
--- a/src/layer/xyz.js
+++ b/src/layer/xyz.js
@@ -22,16 +22,22 @@ var xyz = function xyz(layerOptions) {
   sourceOptions.projection = viewer.getProjectionCode() || 'EPSG:3857';
   sourceOptions.tileGrid = viewer.getTileGrid();
 
-  var xyzSource = createSource(sourceOptions);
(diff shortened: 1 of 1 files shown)


afogelberg:
A useful improvment @jokd . One thing I would change though is the addition of the xyzOptions in createSource. I think I would be more clear if we move this part from createSource:
```
  if(xyzOptions.layerURL){
    options.url += xyzOptions.layerURL;
  }
```
And changed it to something like this:
```
  if(xyzOptions.layerURL){
   sourceOptions.url += xyzOptions.layerURL;
  }
```
Then we don't need to add xyzOptions as an extra argument in createSource.

afogelberg:
@jokd can we close #251 ?

jokd:
@afogelberg Actually not since this commits are merged into the vectortile-branch so the issue is not solved until that PR also is merged. I am obviously no Git-ninja. ;)

afogelberg:
Ok, then we have another reason to start dig into the vectortile-branch :)

PR #263, opened 2017-05-19
Editor toolbar

### LINKED ISSUE #7: Editortoolbar responsive
The editortoolbar should have more appealing appearance on small screens.

afogelberg:
This PR fixes #7 . The dropdown layer switcher is redesigned to make it more suitable for a responsive layout, Also in this PR material icons is introduced and removes the font awesome icons from the toolbar, The editor toolbar now looks like this:
image

Still there can be some improvements for smaller layouts.


jokd:
I like the new look of the toolbar. Like you said, when emulating an iPhone 5/6 the icons are overlapping a bit so they should be moved slightly. 
image

Could we make the edit-button expand a menu and there select what geometry type to draw? Perhaps an separate issue?

afogelberg:
>Could we make the edit-button expand a menu and there select what geometry type to draw? Perhaps an separate issue?

PR coming soon see #264

markus659:
Is the save button always visible? If autoSave is set to true, maybe it should not render at all.

afogelberg:
> Is the save button always visible? If autoSave is set to true, maybe it should not render at all.

There is another purpose of showing the button. If autoSave is set to true the status of the button indicates when the saving to server is done. If there is problem while saving, for example problem with the network, you can retry to save without loosing edits.

markus659:
Ah, yes, I think I might have asked the same question earlier at some point. That is good functionally to always have available.

PR #265, opened 2017-06-01
Edit box

afogelberg:
This PR enhances the draw tool for ther editor control with a draw tool menu. It's now possible to add a tool for drawing a box. If the box tool is added to the editor then a menu with drawing tools will be shown when the draw tool is clicked. The draw tools can be added in two ways:
1) as an option to the editor control. Extra tools are added for each geometry type. The standard tools are always included. Example:
```
    {
      "name": "editor",
      "options": {
        "editableLayers": ["test_byggnadsytor", "test_punkt"],
        "isActive": true,
        "autoSave": false,
        "drawTools": {
            "polygon": ["box"]
        }
      }
    },
```
2) as an extra layer option. All the available tools must be defined and overrides drawTools defined according the first option. Example:
```
      "name": "test_byggnadsytor",
      "title": "Byggnadsytor",
      "group": "test",
      "source": "local",
      "style": "mask",
      "type": "WFS",
      "visible": true,
      "drawTools": ["Point", "Line", "Polygon", "box"]
```
This PR depends on #263


jokd:
This is a great addition to the editor. However I cant get it to work as expected. All editableLayers gets the same drawTools as the first layer in the array. 
I also get "Uncaught TypeError: Cannot read property 'setActive' of undefined" when clicking the layerswitcher-button

afogelberg:
@jokd how is your editor control defined?

jokd:
@afogelberg Just like in the example but with a capital P for the geometryType

afogelberg:
Is the layer not visible or empty? If a layer is not visible then a geometryType is not read yet since the geometryType is read from the first feature loaded to the map. Set the geometry type for the layer to solve it, for example "geometryType": "Polygon".

jokd:
I have set the geometryType for each editable layer in the editor.json so that's not the solution.

afogelberg:
@jokd the geometryType check has been broken which is apparent when the layer is not visible from the beginning. But with the offline control that is now merged with master the geometryType is verified in a better way. It should work now with the latest updates in this branch.

afogelberg:
I stumbled upon a problem when using editor and offline together, due to source change in offline. This is fixed with the latest commit in this branch.

afogelberg:
I got ok from @jokd and will go ahead and merge

Issue #258, opened 2017-05-17
WMS getLegendGraphic option


MahnazN:
Make WMS type layer definitions not require style: , if style is supplied for WMS type layers then it defines an alternative style; use getLegendGraphic for the legend.

tonnyandersson:
We have a solution in place in a pre-Origo version of the code. Never used it in production, though, and I'm not sure it can handle more advanced styling in a good way. But it's a start. I can modify it and provide an example, unless anyone else already have a working solution at hand.

tonnyandersson:
@MahnazN @afogelberg PR is up.

Grammostola:
Sweco made changes in the Summer of '17 to mostly legend.js of our fork to make the map load without style: for wms layers, and for "WMS" type layers to be able to be configured so: 
"New properties for WMS-layer:
{bool} useLegendGraphics
{string} legendGraphicsType ("inline", "toggleable")
{object} legendGraphicSettings (http://docs.geoserver.org/latest/en/user/services/wms/get_legend_graphic/index.html)
" where legendGraphicSettings makes a few options possible and "inline" means on same row as layer title and checkbox and "toggleable" a vertical disclosure arrow, suitable for themed legends.  
![bild](https://user-images.githubusercontent.com/5138906/40498126-66caa0ac-5f7e-11e8-91ae-54586f4ebdc8.png)

where "Områdesbestämmelser" is:
                                        "name": "sokvyx_detaljplan_planomryta_20_omradesbest",
                                        "title": "Områdesbestämmelser",
                                        "group": "bygga",
                                        "queryable": true,
                                        "type": "WMS",
                                        "visible": false,
                                        "source": "etuna",
..
                                        "useLegendGraphics": true,
                                        "legendGraphicType": "inline"
                                      },
https://github.com/Eskilstuna-kommun/EK-extern/commit/bbdd450bbfb26e03cf2e15ed3eca6afab0aa3327

In order to accomodate styles having scale variations new getlegendgraphic-calls were later made to be made on each zoom on the basis that they aren't fetching large amounts of data.

WMS layers can still use style: instead of legendGraphics (not for an alternative WMS-sld-style, rather for a traditional Origo style) if one wishes. (Themed and scale dependant and polygon and line styles we tend to use legendGraphics for)

tonnyandersson:
Outdated. Will probably be reworked and posted as new issue.

Grammostola:
[...]

PR #268, opened 2017-06-09
WMS getLegendGraphic option

tonnyandersson:
Rough draft that solves #258.


tonnyandersson:
Contains two options - get full legend as a single image or get a symbol for each rule in the SLD.

The first option requires no other configuration than setting layer style to "getLegendGraphic", like this:
```json
"style": "getLegendGraphic"
```
Option number 2 requires a style defining rules found in the relevant SLD (see example below). The rules are defined by name and title. The name **must** be the same as the name of the rule in the SLD. The title is optional and is only used for labelling in the legend.
```json
"wms_style": [
  [{
    "sldRules": [
      {
        "name": "Rule1",
        "title": "Rule 1 objects"
      },
      {
        "name": "Rule2",
        "title": "Rule 2 objects"
      },
      {
        "name": "Rule3",
        "title": "Rule 3 objects"
      }
    ]
  }]
]
```

The first option - getting full legend - is preferable for raster styles, e.g. ColorMap etc. Main downside is that the legend is styled by Geoserver and does not blend well with the Origo legend. Looks a bit tacky.

The second option only gets the actual symbol for each rule and uses it as symbol in the legend. Looks a lot better than the first option, but relies on defined rules, both in the Origo config and SLD. The configuration could be made easier by retrieving the SLD from Geoserver automagically, and getting all the rules that way. AFAIK, this would require using the Geoserver REST API, which requires authentication (unless authentication has been disabled). Would be a cool solution, though, but perhaps not as generic since it depends on back-end configurations.

Any thoughts?

markus659:
Thanks for contributing @tonnyandersson. I don't think we need to look at anymore advanced way of handling styling from SLD, at least not currently. If the need occurs, we can revisit it at that point.  

One thing I noticed with the code is that when it replace the source url it expects ows in the url. Perhaps it can be rewritten so that it replaces the last four characters with ```/wms?``, or even just the last three? I'm thinking of these:
```javascript
var url = source.url.replace("/ows", "/wms?");

```
 I configured my source as https://karta.hallstahammar.se/geoserver/wms which obviously did not work with the way it is written now.

I don't know if we should wait for the first release before merging this? What do you think  @afogelberg @jokd?

tonnyandersson:
I agree with your suggestions, @markus659. Unfortunately, I have not been able to set up a decent development environment at my new place yet, so it may take a while before I get around to update the code.

On the other hand, Eskilstuna is working on a different solution with external help. Is theirs a better solution or is it too specific for Eskilstuna's needs to fit the Origo core?

afogelberg:
I think this is a little bit tricky. As I understand the ows/wms way of writing url:s is only used by GeoServer. I haven't looked at Eskilstunas solution and eSamverkan is also working on a solution. Anyways, as you say @markus659 , lets at least wait until after the first release.

tonnyandersson:
For the record, it seems eSamverkan is running some sort of hybrid solution. For non-thematic layers, only the symbol is retrieved. For thematic layers, the full legend is retrieved in one image. I will look into the details when I'm up and running developmentally.

PR #277, opened 2017-06-29
Fixed legend subgroup expand bug

### LINKED ISSUE #276: Subgroups incorrectly marked as expanded in legend
Subgroups (including layers with filter styles) are marked as expanded in the legend on load, even if the subgroup is actually closed. This is due to icon parent div class being incorrectly set to _o-icon-expand-true_ by default.

tonnyandersson:
Fixes #276 and some related bugs.


jokd:
Works well when removing line 376-381. Whats the purpose of these lines?

tonnyandersson:
@jokd To toggle the parent group if the layer is in a subgroup?

jokd:
That seems to work anyway.

tonnyandersson:
@jokd Perhaps those lines have become redundant due to other changes over time.

jokd:
@tonnyandersson Now with a fresh OL-build it works well with the expanded-configuration. However instead I have an issue with the layers style in the legend. The symbols are not rendered correct. Can someone check if thats an issue or just me.
image

jokd:
@tonnyandersson I still have issues with expanded groups though they are set to expanded:false, Can you check whether lines 376-381 are necessary or not. I think that they aren't.

tonnyandersson:
I'll check, @jokd!

tonnyandersson:
@jokd Just removing lines 376-381 does not fix the problem.

jokd:
@tonnyandersson OK, It seems to do the trick for me but the issue sort of come and go.

tonnyandersson:
@jokd I think it depends on the configuration. Different combinations of expanded: true/false, visible: true/false, number of nested groups and number of layers in the nested groups gives different results. For me, it only makes things worse if I remove those lines, or at least same same but different errors. This only applies to the master branch, though. Removing those lines in _this_ branch gives fewer errors, but some issues still remain.

markus659:
This seems to work fine for me. Although, like @jokd mentioned, I'm not sure: https://github.com/origo-map/origo/blob/c10e5d3b8916d4301699f767192d293f89a9b2df/src/legend.js#L376-L381 is needed, I could never get the if condition to be true. I couldn't tell any difference with or without those lines.
I did not have any problems with icons in the legend.

What are the issues remaining? If this fixes #276, maybe it can be merged and we should open new issues for any other problems?

jokd:
Then I think we can merge this. I might have had some garbage in my setup that produced the errors.

________________________________
Från: Markus <notifications@github.com>
Skickat: den 31 oktober 2017 10:12:50
Till: origo-map/origo
Kopia: Jonas Kumlin; Mention
Ämne: Re: [origo-map/origo] Fixed legend subgroup expand bug (#277)


This seems to work fine for me. Although, like @jokd<https://github.com/jokd> mentioned, I'm not sure: https://github.com/origo-map/origo/blob/c10e5d3b8916d4301699f767192d293f89a9b2df/src/legend.js#L376-L381 is needed, I could never get the if condition to be true. I couldn't tell any difference with or without those lines.
I did not have any problems with icons in the legend.

What are the issues remaining? If this fixes #276<https://github.com/origo-map/origo/issues/276>, maybe it can be merged and we should open new issues for any other problems?

afogelberg:
I agree, lets merge this

tonnyandersson:
@markus659 The if condition is true if a non-expanded subgroup contains a visible layer on load.

Basically, those lines check if the parent and grandparent groups of a visible layer are expanded or not on load. If not, the group will be expanded. This is to expose visible layers on load. Perhaps this is not a desirable behaviour?

Though, while trying (and failing) to reproduce the errors I mentioned on Oct. 11 (stupidly enough I did not save that specific config), I ran into a different bug. This bug appears when there is a visible layer in a non-expanded "normal" group, while there are no visible layers in any subgroup. The result is that the group will be expanded, but the expand icon remains a right arrow. Probably just a small fix needed.

So, do NOT merge this just yet.

markus659:
Ok, thanks for clarifying. I'm not sure if it's a desirable behaviour. Basically this means that the layer configuration decides whether the group is expanded or not and takes precedence over the "expanded" option of the group.

I think from the administrator (configuring the map) point of view it might be a little bit confusing. But as a user visiting the map, I would probably expect the group to be expanded. But maybe this is up to the administrators to make sure it works as expected? What do you think @afogelberg @jokd? Is this documented by the way?

I don't know if you are able to merge it yourself @tonnyandersson, but when you feel it's ready it can be merged.

tonnyandersson:
@markus659 That's the way it is set up with regular groups - if a layer is visible, the group will expand, regardless of how the group is configured. This is, in part, where the now infamous lines 376-381 comes into play. Without them, the top level group will not expand if there are visible nested layers.

Not sure which way is the more intuitive, though. But keep in mind that this also has an impact on how the legend looks when using the share map function.

tonnyandersson:
Final (?) bug fixed. Reworked the subgroup parent toggling routine to match the default behaviour, i.e. if _any_ layer is visible within a group structure, the root group (and only the root group) will expand, regardless of its configuration.

PR #282, opened 2017-07-04
Image upload bug fix

### LINKED ISSUE #281: Error when removing image
An error occurs when removing an image and saving if the image attribute is not the very first attribute in the attributes array. Bug inherited from old code.

tonnyandersson:
Fixes #281.


afogelberg:
@tonnyandersson I see you removed the line https://github.com/origo-map/origo/pull/282/files#diff-1426e6712f25bdea9f1464d6817cf14cL551 . This line removes the bbox field since GeoServer wfs requests in many cases send a bbox pseudo value together with each feature.

tonnyandersson:
@afogelberg I know, I put the line there in the first place but forgot to remove it before commit. Didn't know if it was a global issue or a local discrepancy since no issue had been reported on the subject. Also, since the issue was not directly related to the image upload function, I had it removed in this bug fix commit.

afogelberg:
@tonnyandersson I see, if I remember correctly we had the bbox unset earlier also, but probably was removed by mistake at some point. I will merge this and add the bbox unset in a separate commit, even though the problem is GeoServer specific. Thanks for the fix.

tonnyandersson:
@afogelberg That explains it, thanks!

PR #285, opened 2017-07-19
Replaced setview and setzoom in geoposition

### LINKED ISSUE #231: Add animation to geoposition
Requires that we update ol, see #116 .

markus659:
This one-liner fixes #231. Not too much to it really, but I would appreciate an extra look at it just in case as I'm not completely sure what the issue is referring to.

### DIFF
--- a/src/geoposition.js
+++ b/src/geoposition.js
@@ -13,14 +13,14 @@ var geolocation = undefined;
 var marker = undefined;
 var markerEl = undefined;
 var baseUrl = undefined;
-var zoom;
+var zoomLevel;
 
 function init(opt_options) {
   var options = opt_options || {};
   var target = options.target || '#o-toolbar-navigation';
   map = viewer.getMap();
   baseUrl = viewer.getBaseUrl();
-  zoom = options.zoom || undefined;
+  zoomLevel = options.zoomLevel || undefined;
 
   render(target);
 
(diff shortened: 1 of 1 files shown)


afogelberg:
Thanks for working on this @markus659 . Initially I thought the animation shouldn't be chained, but after some testning I think it's more clear to chain it like you have done.

One remark only, I think it would be good to change how the zoom level is set, since we don't know the resolutions in different implementations of Origo. Maybe we can calculate the zoom level from the max resolution, for example max resolution - 3? And just to be safe, also have the zoom level as an option.

markus659:
Thanks for the feedback @afogelberg. I wasn't too sure about the chaining either, I would prefer if the animations started simultaneously (not sure if this is possible or how you would go about implementing it). I think the chaining works fine though. I agree with your other comments, I made changes accordingly in the latest commit. Now it is possible to set zoomLevel as an option for the geoposition control, like so:

```json
    {
      "name": "geoposition",
      "options": {
        "zoomLevel": 15
      }
    }
```
This will result in the animation zooming to the specified level,  if the zoom level is not specified it will be calculated from the resolutions.

afogelberg:
If you don't want chaining you can do it like this:
```
 map.getView().animate({
   center: position,
   resolution: (viewer.getResolutions().length - 3)}
});
```

markus659:
That was easy enough :-).  I prefer it without the chaining, but either works for me. Which one should we go with?

afogelberg:
> I prefer it without the chaining

Agreed :)

PR #293, opened 2017-08-28
Gulp build

### LINKED ISSUE #256: Add dist directory to gitignore and create build directory

afogelberg:
This PR fixes #256 and depends on #289  . A gulp script is added to make a complete "production ready" build of Origo including alla dependencies and examples. This will also be useful when a new release is to be made.

To make a build, simply run the command `gulp` which will run the default task in the gulpfile. Several other tasks can be run as needed, for example `gulp pack` will copy the current built files to the build folder.

The dist folder is only used to create js and css builds of Origo and will not be tracked by git. This will decrease problems with pr conflicts.


markus659:
Good job adding gulp, seems to be a straightforward way of getting a release ready. So to create a release build, you would just need to run ````gulp````? Would running ````npm run build```` and then ````gulp pack```` be the equivalent of running ````gulp````, or when would use ````gulp pack```` over just ````gulp````?

afogelberg:
>  So to create a release build, you would just need to run gulp?

Yes
> Would running npm run build and then gulp pack be the equivalent of running gulp, or when would you use gulp pack over just gulp?

Running just `gulp` will run `gulp pack` and the gulp tasks to make the builds. If you run `gulp pack`, files will only be copied not built. The equivalent to `npm run build` would be running `gulp origo-js` and then gulp `origo-css`.

jokd:
I tried it briefly and it worked great. I think we can merge this.

markus659:
OK, thanks for clarifying @afogelberg. I agree, this can be merged.

PR #296, opened 2017-08-29
Splash

### LINKED ISSUE #295: Splash control
I have made a splash control for a project and I think it could be useful to have it available in Origo. The splash control will show a modal when the map is loaded. Content can be included either as a content option in the control or as an url to a html file.

afogelberg:
Fixes #295


jokd:
Thanks for this great addition. Works well and easily configured!

PR #302, opened 2017-09-07
Adjusted modal z-index

### LINKED ISSUE #297: The about control can be opened on top of it self
If the about control is already opened and it is opened again, it will opened again and be placed on top of the previous.

markus659:
This is sort of related to #297. This PR raises the z-index of the modal screen (so that it is above the map menu, legend and control buttons.) The modal window itself will be on top of the modal screen.

### DIFF
--- a/scss/_modal.scss
+++ b/scss/_modal.scss
@@ -9,7 +9,7 @@
   position: absolute;
   top: 0;
   width: 100%;
-  z-index: 90;
+  z-index: 105;
 }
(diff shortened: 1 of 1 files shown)


afogelberg:
Looks good @markus659 , I think you can merge this. Regarding #297 I actually had another problem in mind, that the modal window can be opened again and again, making the background darker and darker. But we can solve this in a separate PR.

PR #309, opened 2017-09-22
Made wfs loadingstrategy configurable

### LINKED ISSUE #173: Changing loading strategy for WFS?
Today we use loadingstrategy tilegrid for loading WFS features. This results in a lot of requests being sent to the server, many of which return no features (depending on the layer). I don't know if there's any reasoning behind using tilegrid, but could we not use "all" instead? This would greatly decrease the number  of request that are sent, as just one is sent.

https://openlayers.org/en/latest/apidoc/ol.loadingstrategy.html

jokd:
Fixes #173 The loadingstrategy for WFS layers are configurable to load either the whole bbox or tiled. However, I excluded the all-option since it didn't work very well. The strategy can be set either on layer or source.

### DIFF
--- a/src/layer/wfs.js
+++ b/src/layer/wfs.js
@@ -41,7 +41,7 @@ var wfs = function wfs(layerOptions) {
     var queryFilter;
 
     //If cql filter then bbox must be used in the filter.
-    if(options.strategy == 'all'){
+    if(options.strategy === 'all'){
       queryFilter = options.filter ? '&CQL_FILTER=' + options.filter : '';
     }
     else{
@@ -59,7 +59,7 @@ var wfs = function wfs(layerOptions) {
           'version=1.1.0&request=GetFeature&typeName=' + options.featureType +
           '&outputFormat=application/json' +
           '&srsname=' + options.projectionCode;
-        url += options.strategy == 'all' ? queryFilter : queryFilter + extent.join(',') + ',' + bboxProjectionCode;
+        url += options.strategy === 'all' ? queryFilter : queryFilter + extent.join(',') + ',' + bboxProjectionCode;
         $.ajax({
             url: url,
             cache: false
--- a/src/layer/wfs.js
+++ b/src/layer/wfs.js
@@ -20,13 +20,16 @@ var wfs = function wfs(layerOptions) {
 
   sourceOptions.strategy = layerOptions.strategy ? layerOptions.strategy : sourceOptions.strategy;
   switch (sourceOptions.strategy) {
+    case 'all':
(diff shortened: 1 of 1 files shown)


afogelberg:
Nice to get this issue fixed @jokd . I suggest we make the present strategy default, as the change otherwise could be a surprise for our users, Excluding 'all' is fine for me.

markus659:
Thanks for fixing this. What was it that didn't work with the 'all' option? We're currently using it in our map and haven't run into any issues as far as I can tell (granted we have fairly small datasets for the layers loading as WFS). 

Couldn't an if statement be used instead of a switch since it's only two options currently?

jokd:
@afogelberg I vote for bbox as the default since you get loads of requests otherwise. I dont think it will break anything if we change the default as we discussed in the issuethread.
@markus659 The switch was because of the all option which I later on removed. But please try the all-strategy with your data and see if it works. all is the default behaviour for OL so leaving out the strategy for wfs sources should be the same. But even that makes OL crash for me.

markus659:
I will try it with this branch sometime early next week and let you know how it works.

afogelberg:
As agreed earlier, lets go then with bbox as default. I will give the 'all' option a try too.

markus659:
The problem seems to be with the bbox parameter of the url. We don't want/need an url when using the 'all' strategy. I just tried changing the url from
```js       
var url = serverUrl +
          '?service=WFS&' +
          'version=1.1.0&request=GetFeature&typeName=' + options.featureType +
          '&outputFormat=application/json' +
          '&srsname=' + options.projectionCode +
          queryFilter + extent.join(',') + ',' + bboxProjectionCode;
```

to 

```js    
var url = serverUrl +
          '?service=WFS&' +
          'version=1.1.0&request=GetFeature&typeName=' + options.featureType +
          '&outputFormat=application/json' +
          '&srsname=' + options.projectionCode
```
then it worked setting the strategy to 'all'. Could you build the url depending on the value of sourceOptions strategy @jokd? I suppose you would have to consider the queryFilter as well?

jokd:
Thanks @markus659 I have now added all as strategy

markus659:
Excellent! Seems to work as intended. Maybe just change these two lines so that they use '===' instead of '==', in order to follow the guidelines?

https://github.com/origo-map/origo/blob/91ef1ee9a47e684ae7dc8303f09dfddea08f2f4a/src/layer/wfs.js#L44 

https://github.com/origo-map/origo/blob/91ef1ee9a47e684ae7dc8303f09dfddea08f2f4a/src/layer/wfs.js#L62

jokd:
@markus659 I have now added the ='s

PR #310, opened 2017-09-22
Made the measure tool configurable

### LINKED ISSUE #139: Option if both length and area should be active in measure control
I think there should be an option to choose which type of measurements are avalible. For example if you only want to have length measurement, like http://hitta.se/kartan or https://kartor.sollentuna.se/ . It seems to be more common to only have length than have both, therefore I think the default should be only length.

jokd:
Fixes #139 The measure tool is configurable. Default is both tools available and the line tool being the default. 
Configured this way:

    {
      "name": "measure",
      "options": {
        "area":false,
        "default":"line"
      }
    },

### DIFF
--- a/src/measure.js
+++ b/src/measure.js
@@ -8,6 +8,7 @@ var style = require('./style')();
 var styleTypes = require('./style/styletypes');
 
 var map;
+var options;
 var activeButton;
 var defaultButton;
 var measure;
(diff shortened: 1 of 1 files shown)


afogelberg:
How about change the option "area" to measureTools, the default being:
```
{
  "name": "measure",
  "options": {
    "measureTools": ["length", "area"],
    "default": "length"
  }
}
```
This way we can easily add other measure tools, for example measure height.

jokd:
Ah, thinking outside the box. Sounds like a better way to go, I’ll change.

afogelberg:
lgtm

PR #312, opened 2017-09-22
Made clustering more configurable

### LINKED ISSUE #65: Disable cluster when zoomed in
When using clustered layers clustering is always turned on. Clustering should be turned off when zoomed in and when features are identifiable. It is now possible to achieve this since a method to set cluster distance has been added to OL and will be available in the next OL release.

jokd:
Fixes #65 Clusterdistance and maxzoom for clustering can be set on either layer or sourcelevel. Configured like this:
      "layerType":"cluster",
      "clusterDistance":30,
      "clusterMaxZoom":5,
The default is 60 and the highest zoomlevel.
The PR also brings the possibility to disable rotation of the map with the "enableRotation":false setting in index.json

### DIFF
--- a/index.json
+++ b/index.json
@@ -101,6 +101,10 @@
     0.5972
   ],
   "source": {
+    "geoserver": {
+			"url": "http://gi.karlstad.se:8080/geoserver/ows",
+			"version": "1.1.0"
+	},
     "local": {
       "url": "http://localhost/geoserver/wms"
     }
@@ -118,6 +122,28 @@
     }
   ],
   "layers": [
+    {
+			"name": "webbkartan:vy_webbkarta_mf_buf_enskforskola",
+			"title": "forskola",
+			"source": "geoserver",
+			"group": "test",
+			"style": "cirkel",
+			"type": "WFS",
+			"visible": true,
+			"featuretype": 5,
+      "filter": "namn = 'Vitlöken'",
+			"attributes": [{
+				"name": "namn",
+				"title": "Skola: ",
+				"url": "lank"
+			}, {
+				"name": "typ",
+				"title": "Typ: "
+			}, {
+				"name": "adress",
+				"title": "Adress: "
+			}]
+		}, 
     {
       "name": "origo-cities",
       "title": "Origokommuner",
--- a/index.json
+++ b/index.json
@@ -101,10 +101,6 @@
     0.5972
   ],
   "source": {
-    "geoserver": {
-			"url": "http://gi.karlstad.se:8080/geoserver/ows",
-			"version": "1.1.0"
-	},
     "local": {
       "url": "http://localhost/geoserver/wms"
     }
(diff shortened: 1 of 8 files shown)


afogelberg:
Thanks for fixing this @jokd . The adding of a listener for resolution change looks good, it was something like this I had in mind too. I would like to do one change only. I don't think we need to add an extra source option when creating the vector sources. How about adding a layer property called cluster that is used to set cluster settings only? These settings will only be read if the layer type is cluster. It think this will be a little bit cleaner.

The rotation setting looks good and will be nice to have.

jokd:
@afogelberg Then we would have to do that for every layer type. Do you mean somthing like:

  wfsOptions.cluster = {
    clusterDistance : wfsOptions.clusterDistance || sourceOptions.clusterDistance || undefined,
    clusterMaxZoom : wfsOptions.clusterMaxZoom || sourceOptions.clusterMaxZoom || undefined
  };

afogelberg:
@jokd No, I mean that the layer is configured like this for example:
```
      "name": "forskoleverksamhet",
      "id": "0",
      "source": "ags_vasteras",
      "style": "forskola",
      "clusterStyle": "clusterForskola",
      "cluster": {
          clusterDistance : 40,
          clusterMaxZoom : 2
      },
      "title": "Förskola",
      "group": "ags_extern",
      "type": "AGS_FEATURE",
      "layerType": "cluster",
      "filter": "CATEGORY='Förskola' OR CATEGORY='Öppen förskola'",
      "visible": false,
      "queryable": true
```
And that the cluster settings are read in vector.js only, since the cluster settings are already included the layer settings.

jokd:
@afogelberg I would like to have the option to set it on sourcelevel as well.

afogelberg:
> I would like to have the option to set it on sourcelevel as well.

@jokd I agree. Maybe on the application level as well, as for example clusterOptions (as we have for featureinfoOptions).

jokd:
@afogelberg Please review the changes

afogelberg:
@jokd can't we skip these lines, for wfs, GeoJson etc, since we do the cluster layer type check in vector.js?
```
if(layerOptions.layerType=='cluster'){
   topojsonOptions.cluster = {
   clusterDistance : topojsonOptions.clusterDistance || sourceOptions.clusterDistance || undefined,
   clusterMaxZoom : topojsonOptions.clusterMaxZoom || sourceOptions.clusterMaxZoom || undefined
 };
```

jokd:
@afogelberg Not unless we add the sourceOptions when creating the layers.

afogelberg:
@jokd but we get all the layer options in vector.js...

jokd:
@afogelberg We can skip the if expression if that's what you meant, but we must set .cluster afaik

afogelberg:
@jokd both the layer options for cluster and the source options are available in vector.js, no need to set them for each source type. Just get the source options with viewer.getMapSource()[options].cluster

jokd:
@afogelberg You were totally right. I rewrote the whole shebang and now its configured like this:
"clusterOptions": {
          clusterDistance : 40,
          clusterMaxZoom : 2
      }
regardless if its a layer, source or the map.

afogelberg:
lgtm

Issue #313, opened 2017-09-27
Pinch zoom problem in iOS 11


afogelberg:
We have a really urgent problem with the latest upgrade of iOS to 11. The are big problems with pinch zoom, rotate and panning. There is a fix in the latest ol release that solves the problem, which we need to upgrade to.

afogelberg:
I can’t verify that the problem appears on iPhone. So far I have only seen the problem on iPad.

markus659:
I have not noticed on either the iPad or the iPhone. But it does indeed seem to be an issue with at least certain models.

afogelberg:
At least on iPad pro and iPad mini the problem has been verified.

markus659:
I tried it on an iPad Air 2 which seems to work.

afogelberg:
Fixed with #316

PR #316, opened 2017-09-28
Upgraded to ol 4.3.4

afogelberg:
This PR is an ol upgrade tol 4.3.4 and solves #313 . There is a problem with custom builds in this version of ol which is solved and will be included in the next release of ol. Meanwhile see https://github.com/openlayers/openlayers/pull/7229 how to fix.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 dist
 css/style.css
 node_modules
+npm-debug-log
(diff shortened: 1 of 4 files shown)


markus659:
I've also noticed the custom build problem, but since a fix exist I think we can go ahead and merge this. I don't think the custom build problem will affect any of the Origo users.

PR #320, opened 2017-10-03
wms format and tilegrid options

### LINKED ISSUE #319: WMS format options
We need to be able to set the format for WMS-requests.

jokd:
Fixes #319 
This PR gives the possibility to set the format on either layer or source level for WMS requests. It also gives more options when creating a tilegrid.

### DIFF
--- a/src/layer/wms.js
+++ b/src/layer/wms.js
@@ -12,7 +12,8 @@ var wms = function wms(layerOptions) {
   };
   var sourceDefault = {
     version: '1.1.1',
-    gutter: 0
+    gutter: 0,
+    format: 'image/png'
   };
   var wmsOptions = $.extend(wmsDefault, layerOptions);
   wmsOptions.name.split(':').pop();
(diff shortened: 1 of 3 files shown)


afogelberg:
@jokd maputils.tileGrid only have two arguments and not four as in this:
`maputils.tileGrid(settings.projectionExtent, settings.resolutions, settings.tileSize, settings.bottomLeft)`
https://github.com/origo-map/origo/pull/320/files#diff-113ff46cf15064abc1c127f2a79f9c14R68
If we change to four arguments, maybe we should use an options object instead?

I was thinking about this line:
```settings.bottomLeft = mapOptions.tileGrid.bottomLeft;```
https://github.com/origo-map/origo/pull/320/files#diff-113ff46cf15064abc1c127f2a79f9c14R57


This means that `settings.bottomLeft` will be true if bottomLeft is defined with false .

jokd:
@afogelberg Please review my edits.

afogelberg:
lgtm

PR #322, opened 2017-10-11
Simplified measurefunction

### LINKED ISSUE #307: ol.Sphere functions for spherical measures
When moving to OL 4.3+, we should rewrite the measure tool now when there's new measure methods in the ol library; ol.Sphere.getArea() and ol.Sphere.getLength().

jokd:
Fixes #307 Simplified the measure-control by switching to the new ol.Sphere.getArea / ol.Sphere.getLength methods.

### DIFF
--- a/src/measure.js
+++ b/src/measure.js
@@ -272,10 +272,9 @@ function createMeasureTooltip() {
 };
 
 function formatLength(line) {
-  var length, output;
-  var coordinates = line.getCoordinates();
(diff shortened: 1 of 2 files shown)


afogelberg:
@jokd the change looks good. I see you added a format function. We have a utility function for the purpose of formatting, is it something that can be reused in this case? https://github.com/origo-map/origo/blob/master/src/utils/numberformatter.js

afogelberg:
lgtm

Issue #328, opened 2017-10-30
Group layer in map makes cluster fail


afogelberg:
If a group layer is present in a map, the set cluster function will fail.

afogelberg:
Fixed with #329

PR #329, opened 2017-10-30
Moved cluster distance handler

afogelberg:
Solves #328 . In this PR the cluster distance handler is moved to the creation of the cluster layer. This way we avoid potential problems as in #328 .


jokd:
Works well, go ahead and merge

PR #331, opened 2017-10-31
Added cluster support for geojson and topojson

### LINKED ISSUE #330: Add clusteroptions to other vector sources
We should be able to cluster geojson and topojson layers as well.

jokd:
Fixes #330


afogelberg:
lgtm

PR #335, opened 2017-11-01
 custom tileGrid for vectortiles

### LINKED ISSUE #334: Custom tileGrid for vector tiles
Vector tiles should be able to have a custom tileGrid set on source level. This makes it easier to combine vector tiles with e.g wms-services.

jokd:
Fixes #334
Makes it possible to have a different tileGrid for vector tiles sources
which is often necessary when combining with wms. Also changes the
alignTopLeft option to alignBottomLeft to match the default Openlayers
behaviour.


jokd:
Messed up

PR #339, opened 2017-11-03
Added datetype to editform

### LINKED ISSUE #338: Add date-field to editform
We should be able to have the type date and in supported browsers (at least Chrome, Edge Firefox v57) show the native datepicker.

jokd:
Fixes #338

### DIFF
--- a/src/editor/editform.js
+++ b/src/editor/editform.js
@@ -45,6 +45,9 @@ module.exports = function createForm(obj) {
       el += '<input id="o-delete-image-button" class="' + imageClass + '" type="button" value="Ta bort bild">';
       el += '</div>';
       break;
+    case 'date':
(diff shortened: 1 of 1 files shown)


markus659:
Seems to work as expected, a nice addition to Origo. I'm not sure we need the maxLength attribute though? Or what is it purpose in this case?

afogelberg:
@jokd works great in Firefox and Chrome. How about adding todays date as default value? That could help IE users a bit too,

markus659:
Perhaps today's date should auto populate the date field, without any added configuration? Right now you need to specify defaultDate in the config (if I'm doing it right).

jokd:
Now the default behaviour is todays date/time. I you want the field to be empty, set the defaultDate to false.

markus659:
Looks good to me now. I had not noticed the color option previously. Maybe add some information about the color option being added as well when squashing and merging?

PR #343, opened 2017-11-08
Function to zoom to feature in permalink

### LINKED ISSUE #342: Zoom to feature in permalink
Being able to zoom to a certain feature in permalink would be nice.

jokd:
Fixes #342


afogelberg:
@jokd it seems to work fine with wfs layers. How about other types? Couldn't make it work for geojson or ags feature layers. A small thing, how about renaming feature to selection in the permalink?

jokd:
@afogelberg That would need some changes. I have relied on the geoserver "layer.id" syntax to identify the features. The geojson id is straight off (if present) with no layer name. Could this perhaps be a later enhancement? 
The 'selection' is already in use in permalink, though it seems inactivated.

markus659:
Updated to ES6 and added support for ags and geojson. Sidebar should also work now, couldn't get it to work previously.

PR #345, opened 2017-11-08
Added searchable layers

### LINKED ISSUE #344: Searchable layers
It would be nice to be able to set layers as searchable, either when visible or always, and pass this to the serverside search service.

jokd:
Fixes #344 
Makes it possible to set searchable (false(default)/true(search if visible)/always) for layers in the map. Adds &l=layer1,layer2,layer3 to the search service url


afogelberg:
Instead of `settings.layers` I think it would be better to use `map.getLayers()`. This way it gets more dynamic for example if layers should be added to the map.

jokd:
@afogelberg Agreed, and perhaps that should be done for the getQueryableLayers as well.

jokd:
Maybe the l-parameter in the querystring should only be added if search is configured to do so. That way the getSearchableLayers wont be called unless you want it to.

afogelberg:
@jokd makes sense.

jokd:
I have now edited to work with awesomplete. It would be nice to merge this.

afogelberg:
Havn't tried but it looks good. How about making the 'always' setting a general setting for the search control and not a per layer setting? And making searchable just true or false,

jokd:
Would that simplify it? Always are searched for even when the layer is not visible, eg addresses. True is searchable when visible and false is the default.

afogelberg:
Maybe both. I think at least it would be good to be able to change the default behaviour.

jokd:
@afogelberg Now you can set the default behaviour for the control

afogelberg:
This looks good @jokd . One final thought, would it be ok to rename markSearchableLayers to something like includeSearchableLayers or useSearchableLayers?

jokd:
@afogelberg Changed to includeSearchableLayers

PR #349, opened 2017-11-09
Eslint

### LINKED ISSUE #340: Replace jshint with eslint
I suggest we remove jshint and replace it with eslint. There is alson an airbnb extension we can use.

afogelberg:
Fixes #340 . With this PR it will be much easier to follow the airbnb guide lines. A lot of errors will be visible from start in old code, but I suggest we fix that along the way and not everything in one commit. Eslint will not be active at this moment in the build process, as we will get too many errors. We can turn it on in the future when we are ready.


markus659:
Using a shareable config and an airbnb specific config for the linter will certainly make it easier to follow the guidelines. Do we need the react-plugin?

afogelberg:
It seems like a mandatory dependency, I got errors without ut. But it would be nice if we could get rid of it though,

markus659:
Yes, it does indeed seem to be a mandatory dependency. Running: 
```
npm info "eslint-config-airbnb@latest" peerDependencies
```
lists 

```
{ eslint: '^4.9.0',
  'eslint-plugin-import': '^2.7.0',
  'eslint-plugin-jsx-a11y': '^6.0.2',
  'eslint-plugin-react': '^7.4.0' }
```
Could you run init with eslint, if so, were there any prompts about react? Otherwise we might just have to leave it be for now.

markus659:
Is this ready for merging? Seems like we will have to live with the react-plugin, for now at least.

afogelberg:
This is ready for merge, only need to fix the conflict

PR #352, opened 2017-11-16
Added drag-and-drop control

### LINKED ISSUE #351: Drag-and-drop control
OpenLayers has an interaction for drag-and-drop support. I think we should add this as a control to Origo. It makes it super easy to view and identify features by simply dragging the file(s) to the map. If, for instance, you want to view a file it might be easier to view it in an Origo-map than to have to open some GIS desktop software and manually load the background maps, layers etc. that you already have configured in your web map application. 

I've added a gif animation below to demonstrate how it works. Is this something that we want to add support for?

dragandrdrop

markus659:
This fixes #351. It adds support for drag-and-drop. A new group is created at the top of the legend when a file is added. It supports adding multiple files at once. 

The control is configured like so:

```javascript
    {
      "name": "draganddrop",
      "options": {
        "groupTitle": "Mina egna lager"
      }
    }
```


tonnyandersson:
@markus659 I suggest adding support for TopoJSON as well, since the format is included in the standard OL build. Full support, including IGC, would be even more awesome, but perhaps not widely used. GPX, GeoJSON, and KML works like a charm, though.

Also, wouldn't it be nice if you could delete added layers?

markus659:
Thanks for the feedback @tonnyandersson. I added some more format support (I have not actually tried any files though, so if someone has any IGC or Topojson files laying around, feel free to confirm that they work). 

I'm not sure what the use case would be for deleting layers? I'd think it be sufficient with the current map menu capability to change the visibility. Would you care to elaborate?

tonnyandersson:
@markus659 IGC and Topojson works perfectly fine now.

Regarding the ability to delete layers, I just think it's a logical addition - if you can add layers, you should be able to remove them. Use cases? Not a whole lot, I'll admit. But let's say you add a bunch of layers, cluttering up the legend. You want to keep working with some of them, the rest is just excess baggage. Instead of reloading the map, losing all added layers and other settings, you can just remove the unwanted layers. So, not a big deal, just a nice feature.

markus659:
Yeah, I think that would be a case where you would want to be able to remove some layers. If you add a multiple layers, and you are not quite sure what they contain or which one you would like to keep. I added a button next to the layers which can be used to remove layers. This would also work with layers configured in the .json-file (with the exception of background maps), by adding the following configuration:

```json
removable: true
```

jokd:
Removing layers seems to only remove them from the legend, they are still visible in the map.

markus659:
You are right, added commit that removes the layer from the map.

jokd:
@markus659 Works well now

afogelberg:
Seems like something that was corrected in #311 PR was overwritten in this PR. The subgroups are not grouped on the same levels any more, when using nested sub groups.

markus659:
@afogelberg Care to explain a bit more? Not sure what you mean by that they are not grouped on the same levels?

PR #353, opened 2017-11-17
Added option to set mapmenu to open on load

### LINKED ISSUE #346: Option to set map menu visibility on load
A useful addition would be a config option to set if the map menu should be open or closed on map load.

I suggest this should only apply to large screen devices, e.g. desktop, larger tablets etc. For mobile devices with smaller screens, the map menu should be closed on load by default. The line between "larger" and "smaller" screens could be drawn at a screen width of 768px. This means the map menu could be set to "open" on load on an Ipad, but not on an Iphone 6/7/8, not even in landscape mode.

tonnyandersson:
Fixes #346.

Option added to mapmenu control:
```
{
  "name": "mapmenu",
  "options": {
    "isActive": false
  }
}
```
Setting the ```isActive``` option to false is redundant and only for show - the variable is set to false by default.

### DIFF
--- a/index.json
+++ b/index.json
@@ -4,7 +4,10 @@
       "name": "home"
     },
     {
-      "name": "mapmenu"
+      "name": "mapmenu",
+      "options": {
+        "isActive": false
+      }
     },
     {
       "name": "sharemap"
--- a/src/mapmenu.js
+++ b/src/mapmenu.js
@@ -12,9 +12,15 @@ var options;
 var isActive;
 
 function init(opt_options) {
+    var breakPointSize;
+    var breakPoint;
+
     options = opt_options || {};
     isActive = options.isActive || false;
+    breakPointSize = options.breakPointSize || 'l';
+    breakPoint  = viewer.getBreakPoints(breakPointSize);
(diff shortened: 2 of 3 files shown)


afogelberg:
@tonnyandersson a couple of thoughts:
* It would be better to check the innerWidth of .o-map in case it's an embedded map.
* How about getting the size from the breakpoint definition of l in origoConfig.js? That way the break point settings will be consistent. 768px could maybe be a fallback value.

tonnyandersson:
@afogelberg Good points! The window.innerWidth will work fine for maps embedded in iframes, but not if the map is just put in a smaller wrapper div. Will fix.

tonnyandersson:
Added breakpoint option based on definition in origoConfig.js. The option ```breakPointSize``` can be set to any of the defined breakpoints. If ```isActive``` is set to true, the menu will open on load if map size is larger than the specified breakpoint. The ```breakPointSize``` defaults to ```l``` if not set.

Opinions?

afogelberg:
Works fine @tonnyandersson .

PR #356, opened 2017-11-17
Added pageSettings

### LINKED ISSUE #354: Optional map grid
I think it should be configurable whether you want the map grid visible or not. If for instance you select to not have a background map visible, you might not want the grid to be visible either.
Also if you have a background map with a transparent background, you might not want the grid enabled.

### LINKED ISSUE #355: Add a pageSettings option
I think we should intruduce a new configuration option called pageSettings. This would deal with settings
in the viewer.handlebars file. Currently we have a footer option, this could be a part of pageSettings.
This would also make the map grid configurable and therefore be a fix for #354.

markus659:
This would fix #354. It also fixes #355.

### DIFF
--- a/index.json
+++ b/index.json
@@ -50,11 +50,16 @@
       "name": "scaleLine"
     }
   ],
-  "footer": [{
+  "pageSettings": {
+  "footer": {
     "img": "img/png/logo.png",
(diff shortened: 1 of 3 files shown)


afogelberg:
lgtm

PR #357, opened 2017-11-21
Tooltip for background maps in legend

### LINKED ISSUE #301: Tooltip text for map legend
A tooltip text is desirable for background maps in the map legend.

markus659:
This fixes #301.


markus659:
Thanks for the review @jokd. Come to think of it, it would probably work just setting the title attribute when creating the legend item, instead of adding an mouse over event. I think I will try to rework it.

jokd:
I guess you are right^^

asemoller:
Nice work, just what I hoped for @markus659

Issue #350, opened 2017-11-15
Search control doesn't show all hits


jokd:
Sometimes the search result misses hits that are returned in the .json-response. One workaround that works in some cases is to set the searchlimit to 6 but it is a bug nevertheless.

afogelberg:
Fixed with #358

Issue #318, opened 2017-10-02
Linear delimiter in the search result list


asemoller:
When using search from multiple tables, we would like to have a linear delimiter between different sources/tables in the search result list. Something like  [Hallstakartan](https://karta.hallstahammar.se) and [Karlstadskartan](https://karta.karlstad.se).

afogelberg:
I'm working on improving the search control, including #161 with a separate result list. Would it be sufficient to sort the result by source and enhance the search result with an icon for each cateogory/layer? This is very common, see for example hitta, Google maps and Apple maps. If the grouping is till wanted, maybe the result list in #161 can be used in combination?

jokd:
I’d say it’s sufficient. At least for now.

afogelberg:
Fixed with #358

PR #358, opened 2017-11-27
Awesomplete

### LINKED ISSUE #317: Search is unavailable on Andriod
The search field is visible but unavaible for data input. It auto closes when the field is selected for input. This occurs on Samsung using both Samsungs internet and chrome.

afogelberg:
This PR fixes #317 #318 #350 . With this PR typeahead is replaced with the much lighter awesomplete, also meaning that we will get rid of the dependency to jQuery for the autocomplete library.

With this PR also a division is introduced for the search list. The limit will portion an equal number of hits for each searchable layer with hits. 
awesomplete


jokd:
I have tested and have a couple of issues. 
First when the response contains an é (could be other special characters as well) the result list gets buggy and presents all results in one group. 
I have also noted that when continuing to type the search frase, instead of emptying the list when the server sends an empty response the list is intact. Is this the expected behaviour?
The same happens when I type a search frase, empty the search field and then type something that doesnt result in any hits. The result list will then present the results from before I emptied it.

afogelberg:
> I have also noted that when continuing to type the search frase, instead of emptying the list when the server sends an empty response the list is intact. Is this the expected behaviour?

This is expected behaviour. Would you prefer to not keep previous result as a cache?

afogelberg:
> First when the response contains an é (could be other special characters as well) the result list gets buggy and presents all results in one group.

I don't get this problems with special characters. But maybe we have different encodings in the server response?

jokd:
> This is expected behaviour. Would you prefer to not keep previous result as a cache?

Im not sure. It wasnt like that before but it might be a nice feature. 
But when clearing and searching again I think the cache should be wiped away.

jokd:
>I don't get this problems with special characters. But maybe we have different encodings in the server response?
UTF-8 ?

jokd:
I found the bug. It seems to have little to do with charsets or special characters. 
In the `groupToList`-function in search.js theres a .splice-method inside the forEach that can cause results not being displayed. One solution is to instead use a for-loop and back the index when splicing.

`    for (var index = 0; index < types.length; index++) {
      var type=types[index];
      if (nr < limit) {
        var item = group[type][turn];
        if (type in selection === false) {
          selection[type] = [];
        }
        selection[type][turn] = item;
        if (!group[type][turn + 1]) {
          types.splice(types.indexOf(type), 1);
          index--;
        }
      }
      nr++;
    };`

jokd:
@afogelberg Now it works as expected. Well done!

afogelberg:
Ok to merge?

jokd:
Can't see why we shouldn't =)

PR #367, opened 2017-12-04
Fixed full screen problems

### LINKED ISSUE #366: Fullscreen problems
There are a couple of problems with the fullscreen control.
* The full screen button is not visible for small embedded maps, which can be really annoying.
* When an url is used to init the map, the map config will not be read properly if the map is not index.html and when going to full screen mode for an embedded map.
* Not directly related to the fullscreen control, but the map setting in the hash is not read for a permalink depending on have the map is initialized.

afogelberg:
Fixes #366 . It would be great if we can merge this before the first release, since this is a quite important thing to function properly,

### DIFF
--- a/scss/_viewer.scss
+++ b/scss/_viewer.scss
@@ -121,10 +121,8 @@
   }
 }
 &[class*=#{$media-l-m-s-xs}] {
-  .o-tools-left {
-    display: none;
-  }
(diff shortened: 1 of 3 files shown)


markus659:
Looks fine to me, you can go ahead and merge if you like.

Issue #379, opened 2017-12-11
Rotate keeps rotating


mulfvik:
If rotate control is not called the rotation should not be.

PR #380, opened 2017-12-11
Rotation disable

mulfvik:
Proposal for disabling rotation, issue #379

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -90,12 +90,12 @@ function init(el, mapOptions) {
   // Disable rotation if rotation control is not called
   var getRotate;
   for (var i = 0; i < mapOptions.controls.length; ++i) {
-    if (mapOptions.controls[i].name == 'rotate') {
+    if (mapOptions.controls[i].name === 'rotate') {
       getRotate = i;
       break;
     }
   }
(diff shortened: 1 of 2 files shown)


tonnyandersson:
@mulfvik There's an easier way to disable rotation: https://github.com/origo-map/api-documentation/blob/master/content/settings.md#enablerotation

I get what you are after, though. But I think removing a control from origoConfig.js is only supposed to remove the physical control - i.e. buttons and stuff - not removing the interaction connected to the control. Take a look at the zoom control, for instance. This makes it possible to get a cleaner UI while maintaining funtionality.

If we want to be able to remove the functionality together with the physical control, I think this should be done in a more generic way, applicable to all relevant controls and interactions. And perhaps it could be done outside of viewer.js.

But that's just my opinion.

mulfvik:
@tonnyandersson Ah, did'nt get that setting. Time to reread the docs. Thanks Tonny :) Maybe zoom and attribution controls are the exeptions in this case. This should be handled in another issue i guess. Closing this one.

afogelberg:
Also note that you don't have to use the default controls. To remove the default controls you can do this:
```javascript
var map = origo.map.init('test.json', {
    defaultControls: []
});
```

mulfvik:
@afogelberg Its so obvious when you show it :) Cheers Anders

PR #383, opened 2017-12-15
Option to open modal in full screen as iframe

### LINKED ISSUE #381: Full screen modal with iframe
Thoughts on this?

jokd:
Fixes #381 
With this PR it is possible to config an url to be opened inside a modal window instead of a new window. Adding "target" = "iframe" will add a class to the url which will open the link as an iframe inside modal window. 
It also fixes some styleguide issues.

### DIFF
--- a/scss/_modal.scss
+++ b/scss/_modal.scss
@@ -9,12 +9,12 @@
   position: absolute;
   top: 0;
   width: 100%;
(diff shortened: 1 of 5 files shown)


jokd:
Closing this since its rewritten with #480

PR #386, opened 2018-01-15
Upgraded to node-sass 4.7.2

### LINKED ISSUE #385: Upgrade node-sass to 4.7.2
With current version of node-sass that we use in Origo there are problems in the build process when using node 8 or later. We should upgrade to 4.7.2 to avoid the problems.

afogelberg:
Fixes #385

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ Just follow these three steps and you will be up and running in no time.
 The minimum requirements are:
 
   * [Git](https://git-scm.com/)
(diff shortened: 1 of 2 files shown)


jokd:
@afogelberg I think we shall switch to Node 8 and change in the documentation. Its nice to keep up with new versions.

markus659:
Agreed.

afogelberg:
Great, I will go ahead and merge then.

Issue #419, opened 2018-03-15
Add singletile/image as renderMode for WMS


afogelberg:
In some situations it is necessary to use singletile (imageWMS in OpenLayers) to render a WMS service.  Especially it improves the placement of labels.

afogelberg:
Fixed with #421

PR #421, opened 2018-03-15
Added image as rendermode for wms

afogelberg:
Solves #419 by adding renderMode to the WMS layer type. The default renderMode is tile.


tonnyandersson:
Seems to work just fine.

Issue #425, opened 2018-04-13
Search suggestion bug


markus659:
Currently the search shows suggestions even if there are no matches. For instance, if you search for "drott" you get suggestions like "drottningatan", which is expected. However, if you write "drott1", you would still get suggestions from the previous response, so "drottninggatan" would still show up.

This becomes a problem when trying to actually zoom to the feature, I believe this is because the latest response from the server is empty (no hits for "drott1"). You will get an error in the developer console. It's possible to reproduce on any maps using the new awesomplete search, like http://kartor.vasteras.se/sam/ or https://karta.enkoping.se.

AgatonSax:
I vote for the solution @markus659 proposed. If the user inputs a search string with no matches, the expected result is a list of search hits with no matches. I fail to see the logic in showing a list of search hits from a previous search, as discussed in #426.

Also when first doing a search with matching results, removing all of the search query with backspace (for instance) and then typing something with no matches you still get the previous list of search hits.
![image](https://user-images.githubusercontent.com/5526928/67944821-b71d3680-fbdd-11e9-87ba-b94356f73c2b.png)

Also the UI visual state after backspace-clearing the query and pressing the X-button to clear search, are (misleadingly) the same. Pressing X-button triggers clear search db, while removing query with backspace keeps the list of search hits.

Grammostola:
Markus' suggestion (in the closed PR) appears to do the right thing; "Drott" gives hits, "DrottTheFlyingDutch,.." does not, erasing the input field and typing "TheFlyingDutch.." does not show any hits. 

I.e, in src/controls/search.js:initAutocomplete()
`
    awesomplete = new Awesomplete('.o-search-field', {
      minChars: minLength,
      autoFirst: false,
      sort: false,
      maxItems: limit,
      item: renderList,
      filter(suggestion, input) {
        if (suggestion.value.toLowerCase().includes(input.toLowerCase())) {
          return suggestion.value;
        }
      }
    });`

PR #426, opened 2018-04-13
Added condition to suggestion filter

markus659:
Possible solution for #425 . It works by simply not showing any suggestions if the input does not match any part of the suggestions.

### DIFF
--- a/src/search.js
+++ b/src/search.js
@@ -139,7 +139,9 @@ function initAutocomplete() {
     maxItems: limit,
     item: renderList,
     filter: function(suggestion, input) {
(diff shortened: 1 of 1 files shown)


markus659:
I don't remember that discussion (might have missed it). If that is indeed the desired behaviour, then this PR won't fix it, but the issue should still be opened I think. I'm not sure if it's what the user would expect though, looking at how other maps handle search.

markus659:
I will close this PR, but leave the issue open. It's up for grabs if someone has the urge to fix it.

PR #428, opened 2018-04-16
Updates options.default

### LINKED ISSUE #427: Refactor options.default in Measure.js on init
Make options.default line in measure.js in the init function more readable.

daRasmussen:
fixes #427


afogelberg:
I would suggest we define a variable for the default tool, something like this:
```javascript
var defaultTool = 'default' in options ? options.default : 'length';
```

afogelberg:
See #458

PR #440, opened 2018-04-20
Fixed wfs encoding problem

### LINKED ISSUE #439: We found that the Origo map does not work in Internet Explorer when you filter from an attribute containing åäö.

afogelberg:
Fixes #439

### DIFF
--- a/src/layer/wfs.js
+++ b/src/layer/wfs.js
@@ -60,6 +60,7 @@ var wfs = function wfs(layerOptions) {
           '&outputFormat=application/json' +
           '&srsname=' + options.projectionCode;
         url += options.strategy === 'all' ? queryFilter : queryFilter + extent.join(',') + ',' + bboxProjectionCode;
+        url = encodeURI(url);
         $.ajax({
             url: url,
             cache: false


Abelon38:
Hi Andrea!

I hope you are doing well! just installed the new version of Origo and I am getting the Encoding problem. See the picture here:
image

Is that possible to get som advice from you?
Best regards!
Abel

jokd:
How about changing the encoding for the index.html?

tonnyandersson:
@jokd, @Abelon38 

First of all, the problem described above is actually not related to the issue that was fixed by this PR. This problem seems to be a somewhat different story.

It seems that changing charset in index.html won't do the trick here. I did some diggin' and came up with the following:

Origo is using `fetch` to get WFS data. `fetch` interprets data as UTF-8 encoded by default. In this case, however, the WFS data is encoded as ISO-8859-1 at the source. This means that `fetch` is interpreting ISO-8859-1 data as UTF-8 encoded, hence the funny looking characters.

**For the time being, I suggest making sure that the WFS data is encoded as UTF-8 at the source.**

Abelon38:
Hi guys !

Thanks för answering! I did tried with changing on index.html to charset= ISO-8859-1 but that didn't work.
I got it working after changed on Geoserver the Global configuration to UTF-8. I guess this issue should be fixed on oncoming vesions?

I hope you have a great day!
Br/ Abel

PR #451, opened 2018-05-28
Adds error handling

### LINKED ISSUE #450: Invalid WFS request
If a WFS request is invalid it would be nice to see why.

daRasmussen:
Fixes #450


jokd:
This needs to be updated to ES6. What about using the console rather than alert when request fails?

daRasmussen:
@jokd sounds good!

daRasmussen:
This PR is outdated and has been updated in #600

PR #461, opened 2018-07-10
Adds progressbar

### LINKED ISSUE #402: Add Progressbar
I would like to add a progress bar to Origo. 
For each layer the progress bar displays the loading time. 
This feature would prompt the user for patience when loading time is an issue.

daRasmussen:
Fixes #402 works with origo@v1.0.0


afogelberg:
See #462

PR #475, opened 2018-07-18
css fix to click through container

### LINKED ISSUE #374: Issue when interacting with map under container
We should be able to interact with the map even when there is an transparent container above it as for left and bottom toolbar. Eg measuring and editing close to the buttons.

jokd:
Fixes #374


jokd:
Fixed in Origo 2.0

PR #476, opened 2018-07-18
Limit map legend height

### LINKED ISSUE #336: Autocollapse left legend
The left legend can expand quickly when activating a group layer. This way the collapsebutton can be hidden behind the left toolbar. 
One solution is to autohide it when it reaches the toolbar or make it scrollable.
Another way is to make the whole upper row clickable but we still need to set a max height for the legend.

jokd:
Fixes #336


jokd:
Map legend is deprecated in 2.0

PR #477, opened 2018-07-19
Function to pan when selected feature is behind sidebar

### LINKED ISSUE #453: Pan map when using sidebar
It would be nice if the map panned when the selection gets hidden when using the sidebar.
gif

jokd:
Fixes #453 
Introduces sidebarPan as option for featureinfoOptions. Defaults to false:
"overlay": false,
"sidebarPan": true


jokd:
Closing this since its written for Origo v1

PR #479, opened 2018-07-19
Show attribute as title

### LINKED ISSUE #449: Show attribute as title
It would be nice to be able to configure a layer to show an attribute in the titlebar of a popup/sidebar instead og the layername.

jokd:
Fixes #449
Adds a titleAttribute option for layers to show a attribute value instead of the layer name as the title in the popup/sidebar.

### DIFF
--- a/src/featureinfo.js
+++ b/src/featureinfo.js
@@ -9,6 +9,7 @@ import featurelayer from './featurelayer';
 import Style from './style';
 import StyleTypes from './style/styletypes';
 import getFeatureInfo from './getfeatureinfo';
+import replacer from '../src/utils/replacer';
 
 const style = Style();
 const styleTypes = StyleTypes();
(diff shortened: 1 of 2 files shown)


afogelberg:
Good idea @markus659 . I was thinking the same.

jokd:
Of course, will fix that!

jokd:
...though you typically dont wanna have a dynamic title for the layer as suggested, right? In the mapmenu for example. @markus659 @afogelberg

markus659:
Yeah, you're right, that wouldn't work. Maybe declare as such when you want to use an attribute as title:

```json

        "title": [{
          "layer": "Origokommuner",
          "featureinfo": "{{name}}" 
        }]
```

If not, you do it the way it works currently:

```json
       "title": "Origokommuner",
```
Not sure if it's better than the original suggestion.

tonnyandersson:
@jokd @markus659, perhaps a little bit of both? Go with the original suggestion but change the property name from _titleAttribute_ to _featureinfoTitle_ or _popupTitle_ or something, just to avoid ambiguity.

jokd:
I vote for just renaming the attribute to featureinfoTitle (since it also applies on the sidebar). Fix is pushed. 
Also use replacer utility and fall back on layer title / name.

jokd:
Can we merge this? Would be a nice addition for a project of mine.

PR #480, opened 2018-07-19
Option to open modal in full screen as iframe (ES6)

### LINKED ISSUE #381: Full screen modal with iframe

jokd:
Fixes #381
With this PR it is possible to config an url to be opened inside a modal window instead of a new window. Adding "target" = "iframe" will add a class to the url which will open the link as an iframe inside modal window.
Rewrote earlier PR according to es6

### DIFF
--- a/scss/_modal.scss
+++ b/scss/_modal.scss
@@ -9,7 +9,7 @@
   position: absolute;
   top: 0;
   width: 100%;
-  z-index: 105;
+  z-index: 1015;
 }
 .o-modal {
   background-color: white;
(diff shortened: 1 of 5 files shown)


tonnyandersson:
Unable to scroll the page inside the iframe in IOS 11 Safari.

jokd:
@tonnyandersson Fixed it, though it scrolls the whole modal div and not just the container

jokd:
Closing this since it needs to be rewrtten for Origo 2

Issue #393, opened 2018-01-18
Scale picker


tonnyandersson:
Dropdown scale picker control that lets the user set map scale from a list of given scales.

daRasmussen:
Scale picker is depended of scale line and should not be enabled without a proper reference. The scale picker feature should be implemented in the print controller. If Origo provides scale as a controller users might measure lengths and sizes directly on the screen. But that is not accurate since DPI determine the base unit, hence a scale picker is directly depended of a scale line. And Origo is a webbmap used on the webb, user that perfer analog measure tools should not use Origo. They should use printed maps constructed in CAD, AGIS or QGIS...

PR #481, opened 2018-08-03
Scale picker issue #393

osram:
First pull request from Oemap. I hope I've followed the process correctly.
Tell me if something is off.


tonnyandersson:
Seems to work as intended.

One issue, though: when removing the control from the config, the controls in the left toolbox stays in place, leaving an aesthetically displeasing empty space above.

image

I suggest a more dynamic positioning of the various toolboxes, e.g. when no tool is present in the top toolbox it should be collapsed/hidden, allowing the left toolbox to go all the way to the top. This is perhaps a more general issue rather than related to the scale picker issue specifically, and should probably be treated accordingly. Posting for awareness, mainly.

osram:
I guess it would be fairly easy to make it collapse if empty. You want me to do it?

tonnyandersson:
Sure, if you have the time and if you have a somewhat general solution in mind, something that isn't tied to the scale picker specifically.

tonnyandersson:
Replaced and fixed with #725.

Issue #394, opened 2018-01-18
Extended legend


tonnyandersson:
Makes is possible to show a more detailed legend for thematic WMS layers. If applicable, a complete legend is shown docked to the map legend box.

Developed for the Open eMap version of Origo.

PR #482, opened 2018-08-08
Extended legend issue #394

osram:
This is activated by adding "legendIsMultiRow": true to a icon style.
The src parameter point to the image that will be used as a multirow
legend. This can be either a static image or a getLegendGraphics url to
a WMS server

Example:

"styles":{
  "fornminnen": [[{
      "icon": {
        "src": "https://extmap.sundsvall.se/teckenforklaring/Riksantikvarieambetet/fornlamningar_publik.png"
      },
      "legendIsMultiRow": true
  }]]
}


tonnyandersson:
@osram I suggest that the extended legend tab should be hidden when no such layer is visible, Gives a cleaner look and would probably be less confusing as well.

tonnyandersson:
This PR is obsolete. Replaced by #613.

PR #486, opened 2018-08-10
Fix for popup bug

### LINKED ISSUE #485: Popup doesn't wrap lines correctly
I suppose this has to do with upgrading of the Owl Carousel
image

jokd:
Fixes #485
Added fixed width div

### DIFF
--- a/scss/_featureinfo.scss
+++ b/scss/_featureinfo.scss
@@ -1,3 +1,4 @@
 .o-identify-content {
   width: 240px;
+  word-wrap: break-word;
 }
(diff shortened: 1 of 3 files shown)


tonnyandersson:
Good work! Although, it does not break long strings without spaces or hyphens, e.g. long unhyphenated URLs.

I suggest adding _**word-wrap: break-word;**_ to the _**o-identify-content**_ container.

Besides that, it is good to go IMO.

jokd:
@tonnyandersson Made changes as suggested

PR #488, opened 2018-08-13
Fixes #487

### LINKED ISSUE #487: Expanded layer groups bug
When a layer in a subgroup is set to "visible": true, the expanded option for the groups results in the opposite way intended.
Test config:

>   "groups": [
    {
      "name": "test",
      "title": "Test",
      "expanded": true,
      "groups": [
      {
			"name": "sub",
      "expanded": false,
			"title": "Subgroup"
			}
		]
    },
    {
      "name": "background",
      "title": "Bakgrundskartor",
      "expanded": true
    }
  ],
  "layers": [
    {
      "name": "origo-cities",
      "title": "Origokommuner",
      "group": "sub",
      "style": "origo-logo",
      "type": "GEOJSON",
      "attributes": [
          {
            "name": "name"
          }
      ],
      "visible": true
    }

jokd:
Fix to prevent groups and subgroups getting toggled when layer is set to visible:true

### DIFF
--- a/src/controls/legend.js
+++ b/src/controls/legend.js
@@ -445,11 +445,6 @@ function addCheckbox(layer, name, inSubgroup) {
   } else if (layer.getVisible() === true) {
     $(`.${name} .o-checkbox`).addClass('o-checkbox-true');
     if (inSubgroup) {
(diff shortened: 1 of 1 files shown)


jokd:
Fixed in Origo 2.0

PR #498, opened 2018-09-05
OL 5.3.0

### LINKED ISSUE #474: Upgrade to OL5
I have started on a branch and made the necessary changes to get Origo running with OL5:
https://github.com/origo-map/origo/tree/OL5

jokd:
Fixes #474


markus659:
Have you tried if 
```js
import Map from 'ol/Map';
```
would work instead of

```js
import Map from 'ol/map';
```
in the viewer? That's one example OL give in their upgrade notes to 5.x https://github.com/openlayers/openlayers/wiki/OpenLayers-v5.x

It'd be nice if we could be consistent with how we write the imports.

jokd:
@markus659 You are correct, that is the way we should do it. There are lots of changes that needs to be made, I have just fixed the warnings. However, as long as we are consistent we won't get any warnings even if we use the wrong casing.  So there's obviously some work to be done.

jokd:
Now I think we are good to go. Haven't tested in an AGS-environment though. Please have a look @markus659 @tonnyandersson  @afogelberg

tonnyandersson:
Seems to work as expected in AGS-environment.

jokd:
OK to merge this?

markus659:
Maybe we should remove the caret from the openlayers line in the package file, so that it uses a fixed version?

jokd:
@markus659 Sure, can't see why we shouldn't

daRasmussen:
Very Nice!

PR #541, opened 2019-02-13
fixes #538

### LINKED ISSUE #538: Permalinks, not working properly after 2.0 release
This PR should resolve permalinks working for all types.

daRasmussen:


markus659:
Seems to work as expected. Just a few formatting matters I would like to change. The import statements for the component should perhaps be on one line, so instead of:

```javascript
import {
  Component
} from './ui';
```
it could be changed to:

```javascript
import { Component } from '../ui';
```
Airbnb seems to recommend multiline imports when importing more than one export. ESlint complains about indentation in some places, don't know if you are getting the same errors? 

@jokd could you also have a second look at this fix?

jokd:
The fix works well. I also get linter errors in Atom, are you using ESLint provider @daRasmussen

PR #543, opened 2019-02-14
Updated webpack-dev-server version

### LINKED ISSUE #542: Webpack-dev-server should be updated
It should be updated to a newer version.

markus659:
Fixes #542.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "webpack": "^4.27.1",
     "webpack-bundle-analyzer": "^3.0.3",
     "webpack-cli": "^3.1.2",
-    "webpack-dev-server": "3.1.8",
+    "webpack-dev-server": "3.1.14",
     "webpack-merge": "^4.1.5"
   }
 }
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "webpack": "^4.27.1",
     "webpack-bundle-analyzer": "^3.0.3",
     "webpack-cli": "^3.1.2",
-    "webpack-dev-server": "3.1.14",
+    "webpack-dev-server": "3.2.1",
     "webpack-merge": "^4.1.5"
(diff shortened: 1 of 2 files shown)


tonnyandersson:
Tested and build went through just fine. But I get _ReferenceError: Origo is not defined_ when trying to load the build. Works fine when changing script source to origo.min.js, though,

jokd:
I think that’s the expected behaviour. It’s a bug, but an old one.

tonnyandersson:
@jokd Yeah, I know about that bug. Was hoping this PR would deal with that somehow.

tasks/webpack.dev.js
```diff
@@ -3,7 +3,6 @@ const common = require('./webpack.common.js');
 
 module.exports = merge(common, {
   output: {
-    path: `${__dirname}/../js`,
```
daRasmussen: Removing path did not work me first I had to change path to: 
```
path: `${__dirname}`,
```
Then webpack-dev-server ran just fine. Even if I removed path altogether.

package.json
```diff
@@ -54,7 +54,7 @@
     "webpack": "^4.27.1",
     "webpack-bundle-analyzer": "^3.0.3",
     "webpack-cli": "^3.1.2",
-    "webpack-dev-server": "3.1.8",
+    "webpack-dev-server": "3.1.14",
```
daRasmussen: npm audit suggests: 
```
webpack-dev-server@3.2.1
"webpack-dev-server": "^3.2.1",
```

jokd:
I think we can step up to 3.2.1 as well. Tested and works well (both with and without path)
I had to manually install acorn though, not sure why

markus659:
Does it look OK now @daRasmussen @jokd?

PR #545, opened 2019-02-20
Splash fix

### LINKED ISSUE #544: Splash uses unnecessary render dispatch
Since splash is basically just an modal window which is created on load, the render dispatch seems unnecessary. When 
````js 
this.dispatch('render');
````
is run in splash, it will make onRender run again for the modal control, causing the click listener to be added twice. This causes an error when trying to click outside the modal window to close it, as it tries to remove something that was already removed.

### PR DESCRIPTION
**Pull request by markus659**
Fixes #544.

### DIFF
--- a/src/controls/splash.js
+++ b/src/controls/splash.js
@@ -5,7 +5,6 @@ const Splash = function Splash(options = {}) {
   const defaultContent = '<p></p>';
   const cls = 'o-splash';
   let viewer;
-  let modal;
 
   let {
     title,
@@ -17,34 +16,32 @@ const Splash = function Splash(options = {}) {
     url
   } = options;
 
+  const createModal = function createModal(modalContent) {
+    return Modal({
+      title,
+      content: modalContent,
+      cls,
+      target
+    });
+  };
+
   return Component({
     name: 'splash',
     onAdd(evt) {
       viewer = evt.target;
       target = viewer.getId();
       if (!title) title = defaultTitle;
+      if (!content) content = defaultContent;
 
       if (url) {
         const fullUrl = viewer.getBaseUrl() + url;
         const req = new Request(`${fullUrl}`);
         fetch(req).then(response => response.text().then((text) => {
-          content = text;
-          this.render();
+          this.addComponent(createModal(text));
         }));
       } else {
-        if (!content) content = defaultContent;
-        this.render();
+        this.addComponent(createModal(content));
       }
-    },
-    render() {
-      modal = Modal({
-        title,
-        content,
-        cls,
-        target
-      });
-      this.addComponent(modal);
-      this.dispatch('render');
     }
   });
 };
--- a/src/controls/splash.js
+++ b/src/controls/splash.js
@@ -2,7 +2,7 @@ import { Component, Modal } from '../ui';
 
 const Splash = function Splash(options = {}) {
   const defaultTitle = 'Om kartan';
-  const defaultContent = '<p></p>';
+  const defaultContent = '';
   const cls = 'o-splash';
   let viewer;
 

### CODE REVIEW

File: src/controls/splash.js [hunk 1]
```diff
@@ -17,34 +16,32 @@ const Splash = function Splash(options = {}) {
     url
   } = options;
 
+  const createModal = function createModal(modalContent) {
```
daRasmussen: suggestion to change param name to content
```
const createModal = function createModal(content) {
````

File: src/controls/splash.js [hunk 2]
```diff
@@ -17,34 +16,32 @@ const Splash = function Splash(options = {}) {
...
+    return Modal({
+      title,
+      content: modalContent,
```
daRasmussen: if param changes: 
````
title,
content,
cls,
````

File: src/controls/splash.js [hunk 3]
```diff
@@ -17,34 +16,32 @@ const Splash = function Splash(options = {}) {
...
+      target
+    });
+  };
+
   return Component({
     name: 'splash',
     onAdd(evt) {
       viewer = evt.target;
       target = viewer.getId();
       if (!title) title = defaultTitle;
```
daRasmussen: defaultTitle not set as defaultContent

src/controls/splash.js
```diff
@@ -5,7 +5,6 @@ const Splash = function Splash(options = {}) {
   const defaultContent = '<p></p>';
```
daRasmussen: There is actually no need to add:
``` 
<p></p> 
```
as default. I would suggest an empty string.

src/controls/splash.js
```diff
@@ -17,34 +16,32 @@ const Splash = function Splash(options = {}) {
...
+      if (!content) content = defaultContent;
 
       if (url) {
         const fullUrl = viewer.getBaseUrl() + url;
         const req = new Request(`${fullUrl}`);
         fetch(req).then(response => response.text().then((text) => {
-          content = text;
-          this.render();
+          this.addComponent(createModal(text));
         }));
```
daRasmussen: Error handling would be nice. Something like:
```
.catch(e => console.warn(e)); 
```

src/controls/splash.js
markus659: The linter does not approve of this as content is already declared.

src/controls/splash.js
markus659: Not sure what you mean by this, care to clarify or give an example?

src/controls/splash.js
markus659: Sounds good.

src/controls/splash.js
markus659: The linter does not like the use of console (https://eslint.org/docs/rules/no-console). I know it is in use in a few other places, but do we want to add more? Any thought on this @jokd @tonnyandersson?

daRasmussen: @markus659 yeah markus you are right. Maybe we should implement a error message box as a default control component and replace all console log but still have errors displayed?

markus659: Yes, I know we talked about this previously. Origo should have some kind of error handling, errors should be displayed but maybe not by using console statements/alerts etc.

src/controls/splash.js
daRasmussen: Sorry, for some reason I missed the defaultTitle declaration.

src/controls/splash.js
tonnyandersson: Personally, I don't mind using console messages but perhaps we should avoid those for the sake of consistency. An error handling and displaying component sounds like a good idea, as long as it is non-intrusive, i.e. the error messages should not be displayed in a modal. I'd like to see a more subtle solution.

daRasmussen: I agree @tonnyandersson maybe the error component should be an icon with push message capabilities when a severe error occurs, not warning (which still could be stored but not pushed notified). The user gets notified and could click on that icon to get more info, in a menu. Any thoughts?

tonnyandersson: Something along those lines, yeah.

jokd: I am looking forward to your error component @daRasmussen, Origo would benefit from it.
Meanwhile maybe we can merge this and create a new issue for the error component?

daRasmussen: @jokd sure

PR #548, opened 2019-02-22
Updates getattributes.js

### LINKED ISSUE #547: Value 0 is interpreted as undefined
If attribute value is null or a number (not string) it does not show in the popup which can be misleading.

daRasmussen:
fixes #547

### DIFF
--- a/src/getattributes.js
+++ b/src/getattributes.js
@@ -29,7 +29,7 @@ export default function (feature, layer, map) {
         title = '';
         val = '';
         if (attribute.name) {
-          if (feature.get(attribute.name)) {
+          if (feature.get(attribute.name) !== undefined) {
             val = feature.get(attribute.name);
             if (attribute.title) {
               title = `<b>${attribute.title}</b>`;
--- a/src/getattributes.js
+++ b/src/getattributes.js
@@ -29,7 +29,8 @@ export default function (feature, layer, map) {
         title = '';
         val = '';
         if (attribute.name) {
-          if (feature.get(attribute.name) !== undefined) {
+          const featureValue = feature.get(attribute.name) === 0 ? feature.get(attribute.name).toString() : feature.get(attribute.name);
+          if (featureValue) {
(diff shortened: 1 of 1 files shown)


tonnyandersson:
Not sure if this is the preferred behaviour regarding null values. Personally, I prefer a cleaner popup, without a bunch of null value attributes listed. But I understand it can be seen as a bit misleading depending on type of application and intended use.

0 values, on the other hand, is a different story. That should definitely be regarded as a bug.

daRasmussen:
@tonnyandersson I agree that Null values are undesired in the popup. Which would mean that strings like "Null", "undefined", "NaN", "Infinity", "[]" and "{}" would also be undesired in the popup as preferred behaviour?

tonnyandersson:
Well, I guess "Null", "undefined", "[]" and "{}" can be regarded as different representations of "empty" and should therefore, in my opinion, not be shown in the popup. "NaN" and "Infinity", on the other hand, indicates potential errors in the data. Not sure if those should be treated the same way.

daRasmussen:
@tonnyandersson Maybe we could solve these issues with a attribute mode option like:

- verbose, shows all 
- errors, shows only errors in data
- default, removes errors and empty (is default)

Any thoughts @markus659 @jokd

jokd:
@daRasmussen Go ahead with the attribute mode option as suggested.

jokd:
Can we solve this value 0 bug now and look into the attribute mode option thing later if there is much work involved?

tonnyandersson:
@daRasmussen I agree with @jokd regarding his last comment. The 0 value bug is a quite serious bug that needs to be fixed immediately.

jokd:
If this PR solves the "0 value bug" I think we should merge it and perhaps later pick up the issue regarding the other possible values.

tonnyandersson:
Yep, this PR solves the "0 value bug". TMNT. Merge it.

jokd:
FTW

PR #553, opened 2019-03-01
Update editform.js

### LINKED ISSUE #552: isVisible in editform for all types
It would be nice if editform would respect the isVisible configuration setting for all types.

daRasmussen:
Fixes #552

### DIFF
--- a/src/controls/editor/editform.js
+++ b/src/controls/editor/editform.js
@@ -15,14 +15,14 @@ const createForm = function createForm(obj) {
   let firstOption;
   switch (type) {
     case 'text':
-      el = `<div><label>${label}</label><br><input type="text" id="${id}" value="${val}" ${maxLength}></div>`;
+      el = `<div class="${cls}"><label>${label}</label><br><input type="text" id="${id}" value="${val}" ${maxLength}></div>`;
       break;
     case 'textarea':
-      el = `<div><label>${label}</label><br><textarea id="${id}" ${maxLength} rows="3">${val}</textarea></div>`;
+      el = `<div class="${cls}"><label>${label}</label><br><textarea id="${id}" ${maxLength} rows="3">${val}</textarea></div>`;
       break;
     case 'checkbox':
       checked = val ? ' checked' : '';
-      el = `<div class="o-form-checkbox"><label>${label}</label><input type="checkbox" id="${id}" value="${val}"${checked}></div>`;
+      el = `<div class="o-form-checkbox ${cls}"><label>${label}</label><input type="checkbox" id="${id}" value="${val}"${checked}></div>`;
       break;
     case 'dropdown':
       if (val) {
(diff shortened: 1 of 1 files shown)


jokd:
This PR makes it possible to set the class for an attribute to "cls":"o-hidden " and hide it. Is that what you meant? Or how is isVisible to be used? If that is the case then it should add a whitespace between the cls and the concatenated id.
When digging into the code I realized the editor control is poorly documented, eg the constraint property is not documented at all.

daRasmussen:
If isVisible is set to false in config settings the attribute will be hidden yes. If cls param is set to for example my-class the attribute div should honor that class definition in editform. 

I am not sure what you mean with add a whitespace, maybe you could leave a review? 

On a more general note. Maybe the isVisible param should be removed from editform? Setting cls to o-hidden in config settings would suffice. I really don't see the use for setting value to empty if attribute is set to not visible. Is it not better to just remove that attribute from the list of attribute in the config setting?

jokd:
"isVisible":false has no effect when I test. 
"cls": "o-hidden " works and hides the attribute but I have to add a whitespace after o-hidden otherwise its concatenated with the ID in the function. 
Having hidden attributes is useful if you don't want the user to edit the value or if you want to set it programmatically/to a default value. However I'm not sure if the latter is possible?

daRasmussen:
@jokd thanks for input I can fix this but I have one more question/suggestion regarding “isVisible” option parameter. The question is related to your previous comment about attributes and default values. 

https://github.com/origo-map/origo/blob/f7922c05febdb6cd807a6de0027ae03d5680a260/src/controls/editor/editform.js#L7

I think isVisible should not impact the value of the attribute since this behavior is redundant. That behavior on line 7 makes it possible to remove the value from attribute in editform if isVisible is set to false, undefined, null or 0. 

This enables user to not prioritize attribute settings. Instead the user should remove attributes from settings if they want to make editform attribute not Visible and not useable. 

If we remove isVisible to not impact values in editform:

```
const val = obj.val;
```

Then we could remove these lines: 

https://github.com/origo-map/origo/blob/4191638e4234e10f44c2fee95b49b10336aec61e/src/controls/editor/edithandler.js#L466
https://github.com/origo-map/origo/blob/4191638e4234e10f44c2fee95b49b10336aec61e/src/controls/editor/edithandler.js#L470

Line 470 in edithandler is the reason why isVisible param does not take effect from the config.
If it's all good I can update this PR, or does @markus659 @tonnyandersson have any thoughts?

tonnyandersson:
@daRasmussen I agree that the visibility of an attribute should not have an impact on its value, that's just illogical. That being said... isVisible is strongly connected to the concept of constraints, i.e. attribute values depending on other attributes' values. That is something you'd have to consider before making any changes.

Non-visible attributes can be really useful, though, for example if you want to set default values or set values programmatically, like @jokd mentioned. Not an option in Origo today (AFAIK), but making attributes "invisible" by removing them from the config effectively kills any such endeavours.

jokd:
Wouldn't introducing a `type: 'hidden'` with defaultValue option solves this issue, instead of working with css classes. Or as a compliment.

daRasmussen:
@jokd I guess that is a valid option and I support it.

tonnyandersson:
@jokd But what if you, for example, want to set a timestamp automatically when adding features? Then you would probably want a hidden date field, defaulting to the current date/time.

jokd:
@tonnyandersson Wouldn't a regular input type="hidden" with a default value of 'now' solve this?

tonnyandersson:
@jokd Possibly. I guess we would have to define a few default value types for that to work, though, like "now"/"today" and such.

jokd:
Solved with #696

PR #571, opened 2019-03-07
Add removable option for layers

### LINKED ISSUE #569: Drag-and-drop missing remove functionality
Prior to the 2.0 release it was possible to remove layers that had been added using drag-and-drop. This is no longer possible.

### LINKED ISSUE #570: Add ability to remove layers
It would be useful to have the option to set layers to removable, this would for instance be useful for drag-and-drop (it would fix #569), and in other cases where layers are added on-the-fly, or even if they are added using the config.

markus659:
This fixes #569 and also fixes #570.

### DIFF
--- a/src/controls/legend/overlay.js
+++ b/src/controls/legend/overlay.js
@@ -47,8 +47,6 @@ const OverlayLayer = function OverlayLayer(options) {
     return !visible;
   };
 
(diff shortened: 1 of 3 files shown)


jokd:
Nicely done, it even works in IE11. You can merge this.

PR #575, opened 2019-03-11
New mask, point layer and style

### LINKED ISSUE #551: Origo mask and point layer should include the new members of the project

jokd:
Fixes #551
Made the icon a bit smaller as well as added the new members of Origo map


tonnyandersson:
Looks good.

PR #578, opened 2019-03-12
Fixes #566

### LINKED ISSUE #566: Empty sharemap link on search result.
It would be nice if sharemap link would still be useable on search result.

daRasmussen:


jokd:
Are you supposed to get the popup visible or just zoomed to the right extent?
In case we want the former src/featureinfo.js needs to be rewritten since the functions adds 'name' to the ID. So if your search service returns a valid ID, it gets corrupt since selection.type will be undefined coming from search results in a featurelayer. I rewrote parts of getSelection function in featureinfo.js and published it as sharemap-fix branch as an alternative to this PR. I think GEOJSON and AGS_FEATURE are the types that needs to have name added to the id.

daRasmussen:
@jokd it works, you can merge.

PR #580, opened 2019-03-13
Alternative to #578

### LINKED ISSUE #566: Empty sharemap link on search result.

jokd:
Fixes #566
Share map with search results from visible layer


daRasmussen:
@jokd this works you can merge.

PR #585, opened 2019-03-18
Fixed build script

### LINKED ISSUE #584: Unable to deploy ./build without modification due to invalid js/origo.js
When deploying the ./build folder on a web server it wont work since the js/origo.js is invalid. The minified js/origo.min.js will work though if you manually make the switch.

jokd:
Fixes #584


tonnyandersson:
Works fine! But is it really necessary to include the build/js/origo.js file in this PR? Not saying that it shouldn't be included, but historically, we've only updated build files before making new releases. Sort of.

jokd:
Well true, but this way you can dl master and it'll actually work out of the box.

jokd:
But maybe we should make this fix 2.0.2 unless we rerelease 2.0.1

tonnyandersson:
New features has been added since 2.0.1, so I guess it would be more correct to make a 2.1.0 release. But that feels like too big a leap.

jokd:
Removed origo.js and merged

PR #501, opened 2018-09-07
Legend circle

jokd:
Fixes #68 
Scales down circle styles to fit the legend
Fixes #166 
Styles using the 'default'-stylefunction where no style is set or not defined.

PR #586, opened 2019-03-18
Falls back on default style instead of throwing error

### LINKED ISSUE #166: Provide default style
If style is not provided for a layer a default style should be used

jokd:
Fixes #166 and replaces #501

### DIFF
--- a/src/style.js
+++ b/src/style.js
@@ -208,8 +208,9 @@ function createStyle({
   const styleSettings = viewer.getStyle(styleName);
   const baseUrl = viewer.getBaseUrl();
 
-  if (Object.keys(styleSettings).length === 0 || !styleSettings) {
-    throw new Error(`Style ${styleName} is not defined`);
+  if (!styleSettings || Object.keys(styleSettings).length === 0) {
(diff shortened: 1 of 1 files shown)


tonnyandersson:
Works well, mostly, but is seems the default point style is not being respected. According to the default style function, Point and MultiPoint layers should be rendered with yellow fill and stroke but are rendered in black.

jokd:
Yeah, the function shouldnt create the style before the fill is set. That is perhaps another issue.

tonnyandersson:
Yeah, perhaps you're right. Another thing, though. The legend is not picking up on the default styles. I'm guessing that's because the legend gets the layer style definitions from the config file. Different (and slightly bigger) issue, but something that needs to be addressed for this to work as intented.

Other than _that_, this PR seems fine.

jokd:
Agreed thats quite a job. A simple solution meanwhile could be to show a generic icon when using a stylefunction. Can we merge this anyway and form an issue for the stuff thats left?

tonnyandersson:
I'm cool with that.

PR #593, opened 2019-04-01
Fix for cluster and share map issue

### LINKED ISSUE #590: Cluster and share map issue
When layer is using the cluster option, shared link ignores the zoom to and popup magic

jokd:
Fixes #590

### DIFF
--- a/src/featureinfo.js
+++ b/src/featureinfo.js
@@ -58,6 +58,7 @@ const Featureinfo = function Featureinfo(options = {}) {
     if (currentItem !== null) {
       const clone = items[currentItem].feature.clone();
       clone.setId(items[currentItem].feature.getId());
+      clone.layerName = items[currentItem].name;
       selectionLayer.clearAndAdd(
         clone,
         selectionStyles[items[currentItem].feature.getGeometry().getType()]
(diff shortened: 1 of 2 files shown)


tonnyandersson:
I was about to test this but I'm not able to reproduce the original error. Tried the [2.0.1 release test map](https://www.malardalskartan.se/origo/v2.0.1/test.html), using the clustered AGS Feature Service layer, and the share map function seems to work just fine. Am I missing something, @jokd?

jokd:
@tonnyandersson For AGS feature layers you will get a link but it's lacking layername so Origo won't know in which layer the feature is. And for WFS layers, you won't get a link at all.
So it is sort of two issues that is solved with this PR.

tonnyandersson:
@jokd Still not able to reproduce the bug. AGS_FEATURE, WFS, GEOJSON - they all seem to work perfectly fine when sharing, even when clustered. The only problem I can see is with WMS/AGS_TILE and TOPOJSON, but those are different issues.

That said, I can see that this PR does have some merit since it handles layername issues for TOPOJSON layers and also unmanaged layers.

jokd:
@tonnyandersson Strange, are you testing against the test server or your own? You must have minimum two features in the popup (carousel that is) to trigger the bug.

jokd:
Is this OK to merge?

PR #594, opened 2019-04-01
Fixes #589

### LINKED ISSUE #589: When background layers are missing style, legend is not rendered

jokd:


tonnyandersson:
Not a fan of hardcoded resource paths but I don't have any better ideas, so... LGTM.

PR #600, opened 2019-04-12
Adds Error handling for WFS

### LINKED ISSUE #450: Invalid WFS request

daRasmussen:
Fixes #450


daRasmussen:
[.json config](https://www.malardalskartan.se/origo/v2.0.1/test.json)

The following cases have been tested. 

case 0  :  *no features are returned*
Add filter `"filter": "ETABL_OMR == 'Gävle'", ` WFS layer type with name etableringsomraden.

case 1 :  *TypeName is missing*
Change name to `"name": "etableringsomradenQ",`

case 2: *invalid source*
change `mdk` source to `"url": "https://www.malardalskartan.se/",`

jokd:
Case 0 where no features are returned isnt necessarily an error; this happen as soon as you pan the map and no features are in the bbox (if you use bbox as loadingStrategy which I think is the default).

daRasmussen:
@jokd agree that the bbox as loadingStrategy is a nice feature, this PR would enhance this feature for layers that toggle visibility. If case 0 is an error, warning or info is semantics that could be reworked later when the error handler is inplace, alerts will disappear eventually this is just a first step towards more error handling in origo.

jokd:
I think we need to refurbish this before merge. As now you just have to pan to an area where theres no features to trigger the alert.
gif

daRasmussen:
@jokd Ofc, I can look into that. My original thought was to ascertain this response for the "logger" or "error handler" for future use in for instance verbose mode.

jokd:
@daRasmussen Works better now. In production maybe I don't want to pipe out case 0 to the console, will setting log level in #662 to fail solve this? That should perhaps be the default.

daRasmussen:
@jokd yes working on that, so that case 0 only displays in verbose mode. Default settings would be to only show errors not warnings.

daRasmussen:
@tonnyandersson @jokd @johnnyblasta

jokd:
@daRasmussen Looks good, however maybe we shouldnt merge this before the prompt control is in place?

daRasmussen:
@jokd sound good

jokd:
Had a look at this and while it is a nice feature it needs some further development to be useful, as well as conflict resolving with latest Origo release. I suggest we close this and reopen it in case someone wants to put some efforts into it.

PR #601, opened 2019-04-12
updates agsTransaction

### LINKED ISSUE #455: agstransaction url
The url in agstransaction could be fixed.

daRasmussen:
Fixes #455

### DIFF
--- a/src/controls/editor/agstransaction.js
+++ b/src/controls/editor/agstransaction.js
@@ -122,7 +122,7 @@ function agsTransaction(transObj, layerName) {
   types.forEach((type) => {
     if (transObj[type]) {
       const u = source.url.slice(-1) === '/' ? source.url : `${source.url}/`;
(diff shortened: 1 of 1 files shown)


daRasmussen:
This is basically what this PR does. 

agsTranscationTEST

daRasmussen:
Yepp my point is...

tonnyandersson:
Your point is...?

daRasmussen:
@tonnyandersson could you give an example? The slice function is not applied to **ID** its applied to **URL**.

tonnyandersson:
Yes, that's exactly what I mean. Id's are not always strings, and the integer datatype does not have the slice function.

Btw, do we have a policy regarding template literals vs toString()? I'm okay with both, just asking for future reference.

tonnyandersson:
LGTM

daRasmussen:
I would say that it depends on the context and that it is better to be consistent and avoid mixing and in general literals delivers better performance than functions.

PR #603, opened 2019-04-16
Adds Progressbar

### LINKED ISSUE #402: Add Progressbar

daRasmussen:
Adds progressbar

### DIFF
--- a/conf/origoControls.js
+++ b/conf/origoControls.js
@@ -12,6 +12,7 @@ import measure from '../src/controls/measure';
 import offline from '../src/controls/offline';
 import position from '../src/controls/position';
 import print from '../src/controls/print';
+import progressbar from '../src/controls/progressbar';
 import rotate from '../src/controls/rotate';
 import scale from '../src/controls/scale';
 import scaleline from '../src/controls/scaleline';
@@ -35,6 +36,7 @@ const controls = {
   offline,
   position,
   print,
+  progressbar,
   rotate,
   scale,
   scaleline,
--- a/scss/_progressbar.scss
+++ b/scss/_progressbar.scss
@@ -3,7 +3,7 @@
     bottom: 0;
     left: 0;
     height: 0.2em;
-    background: rgba(124, 186, 248, 1);
+    background: $blue;
     width: 0;
     transition: width 1ms;
 }
--- a/scss/_progressbar.scss
+++ b/scss/_progressbar.scss
@@ -1,0 +1,9 @@
+.o-progress {
+    position: absolute;
+    bottom: 0;
+    left: 0;
+    height: 0.4em;
+    background: hsla(0, 100%, 19%, 0.8);
+    width: 0;
+    transition: width 1ms;
+}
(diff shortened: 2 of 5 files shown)


daRasmussen:
[config file](https://www.malardalskartan.se/origo/v2.0.1/test.json
)

Add to Controls. 
` { "name": "progressbar" }`

tonnyandersson:
Works fine! But why not use the standard blue color, same as active controls?

Also, I suggest using hex or RGB(A) color codes instead of HSL for consistency.

daRasmussen:
@tonnyandersson [thats why](https://contrast-ratio.com/)

tonnyandersson:
Personally, I think this...

image

...stands out more than this...

image


When using an orthophoto as background, a brown/burgundy progressbar will be even less visible.

daRasmussen:
@tonnyandersson good suggestions maybe the color could be an option? If so in a separate ISSUE/PR. 

Personally I think the color should be subtle and as much in the background since the progress bar is not the center of the attention in the application. The Map is where primary focus should be. The only time a user should look at the progress bar is if the content is stuck in loading etc.

tonnyandersson:
@daRasmussen I agree that the progress bar should not be in the center of attention, but I still think it should stand out enough for the user to instantly recognize it as a progress bar. Also, I think we should keep the list of colors used in Origo as short as possible, in order to maintain a clean UI.

So, I'll stick with my suggestion to go with the standard blue color. If you think that's too obtrusive, you could make the progress bar thinner. I'd be okay with that.

@jokd @markus659, opinions?

daRasmussen:
@tonnyandersson great suggestion! I am ok with that, but I still think an Issue/PR would be fine for options like position, color and shape. I will fix this and get back to you!

markus659:
Nice work. I agree with @tonnyandersson  about keeping the numbers of colours used to a minimum in order to keep a clean and coherent UI. Like @daRasmussen mentioned though, the progress bar should not be the main focus, as I think we all can agree. 

Would it be possible to change how soon the bar appears? I felt like it showed up a bit too much when trying it. Maybe making it thinner as suggested would be a thing worth trying as well.

daRasmussen:
@markus659 @tonnyandersson update complete, progressbar is now thinner and I changed the color. 

@markus659 @tonnyandersson @jokd Regarding the color, then the use off the global sass variables should be requirement for a PR to accepted? That would help to keep the amount of colors to a minimum? And for consistency we should update ui/_variables.scss? AND only have one _variables.scss not multiple...

image

I think its fine to mix hsl and rga or rgba, but if majority of you want to change to a specific function we should do so and maybe state somewhere that it is an requirement for future PR's. 

@markus659 Regarding bar apperans, I think this could be resolved in a future PR as an enhancement.

markus659:
Yes, the guidelines are lacking information about the UI. And there might be a need for a more subtle colour, the current ones aren't set in stone. What colours should be used, when should they be used etc. is documentation that is needed in order to maintain a consistent UI, I think. 

Yep, when the bars appears can be solved in a future PR if there is enough support for it.

tonnyandersson:
LGTM. Clearly visible but doesn't steal the show.

PR #610, opened 2019-05-01
Click interaction

### LINKED ISSUE #572: Only featureinfo responds to 'toggleinteraction' event
If a 'toggleinteraction' event is triggered all other interactions are supposed to get deactivated. And this was how it worked in the previous version, but now only featureinfo listens and reacts to this event. This means that if other modules or controls use interactions the other interactions remains active which is a strange behavior.
Measure control, Draw control, and Multiselection control are those who use ol interactions as far as I am concerned.
How do we solve this?

jokd:
Fixes #572 
Moved interaction handling to viewer.js. This requires the controls that uses click interactions to have disableInteraction and enableInteraction-functions; see the edits for measure.js for this and how to dispatch the toggleInteraction-event. Todo is the editor control unless its reformed as a plugin.


daRasmussen:
It's a solution but I think @afogelberg would disagree

asemoller:
Are any of you other core devs up for a review of this pr? It would be nice to have this issue resolved :)

daRasmussen:
@asemoller we did this week so it will be resolved soon.

asemoller:
Sounds great, looking forward to it!

tonnyandersson:
Looks good! My only question is whether the interaction handling should be kept in a separate file or not, for easier maintenance.

jokd:
@tonnyandersson Please review the edits

tonnyandersson:
LGTM.

afogelberg:
Sorry coming late to the party :). But I think you could make this easier without extra interaction handler. What you actually seem to do is to add an already existing functionality. The way I would have done it, is to use the components already built in event handling. For example you could 
 do viewer.dispatch(som event). Then viewer can listen to that with viewer.on(some event). This means that measure and other controls can dispatch the same type of event to viewer which viewer then can act on.

What would also be a good idea in the long run is to introduce a common store to keep shared state for Origo. The you would dispatch changes to the common store instead of individual components.

jokd:
@afogelberg I have rewritten the function, is it like you suggested?

afogelberg:
👏 looks good @jokd . A future enhancement, as I mentioned above, would be to have a global store for shared state. It would be perfect for a case like this when several components need access to and share state. That way it would also reduce the complexity with dependency to a parents state, in this case viewer.

PR #611, opened 2019-05-02
Alternative WMS style

imantahriri:
configuration in index.json be like:
`{
                    "icon": {
                        "src": "http://www...."
                    }
                },
                {
                    "sldStyle": "sld_style_name_in_geoserver"
                }`


daRasmussen:
denied! 
no test and geoserver sld... OGC standard or build of geoserver if so, what version? <3

tonnyandersson:
ArcGIS Server throws a ServiceException error when testing this. It seems ArcGIS Server WMS requires a STYLES parameter, even if it is empty. Now, when no sldStyle is provided, _sourceOptions.style_ will be _undefined_, which means there will be no STYLES parameter in the request url.

If an sldStyle is not provided, _sourceOptions.style_ should return an empty string, rather than _undefined_.

tonnyandersson:
@imantahriri Also, please remove the changes in .gitignore. Those changes are not related to this issue.

tonnyandersson:
Just one more thing, @imantahriri.

Perhaps you could rename all mentions of "sld" to "wms" to make it more... generic. "sld" feels a bit to Geoserver specific. Example:

sldStyle -> wmsStyle
sldStyleObject -> wmsStyleObject

Oh, also please remove the changes to .gitignore. When you removed the plugins folder you also removed the newline at the end of the file. All files must end with a newline, so please put it back.

Other than that, I think we're good to go.

tonnyandersson:
Renamed this PR to make the title more accurate in relation to what this PR actually does.

PR #613, opened 2019-05-05
Extended legend

imantahriri:
if there is an icon in the style config of a layer that is flagged as extendedLegend like:
```
{
  "icon": {
    "src": "url request to GetLegendGraphic or any other statis asset"
   },
  "extendedLegend": true
}
```
then it will be embedded in the legend instead of the usual legend.

Related issue: #394

### DIFF
--- a/src/utils/legendmaker.js
+++ b/src/utils/legendmaker.js
@@ -121,7 +121,6 @@ export const renderLegendItem = function renderLegendItem(svgIcon, label = '') {
 };
 
 export const renderExtendedLegendItem = function renderExtendedLegendItem(extendedLegendItem) {
-  const style = `style="width: ${size}px; height: ${size}px;"`;
   return `<li class="flex row align-center padding-y-smallest">
             <img src=${extendedLegendItem.icon.src} />
           </li>`;
--- a/src/utils/legendmaker.js
+++ b/src/utils/legendmaker.js
@@ -44,12 +44,23 @@ export const findStyleType = function findStyleType(styles) {
   return null;
 };
 
+// This returns the last styleRule flagged as header. In other words, if there are 3 styleRules 
+// an all of them have header=true, then the last one will be returned and set on the icon legend.
+// If there is only one styleRule that will be used as header icon.
+// If there are more than on styleRule but none of them has header flag, then null is returned meaning that list_icon will set as header icon
 export const findHeaderStyle = function findHeaderStyle(styleRules) {
   if (styleRules.length === 1) {
+    const icons = styleRules[0].filter(sr => sr.icon);
+    if (icons && icons.length && icons[0].extendedLegend) {
+      return null;
+    }
     return styleRules[0];
   }
   return styleRules.reduce((prev, styleRule) => {
-    if (styleRule.filter(style => style.header).length) {
+    const headerItems = styleRule.filter(style => {
+      return style.header
+    });
+    if (headerItems.length) {
       return styleRule;
     }
     return prev;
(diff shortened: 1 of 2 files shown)


tonnyandersson:
Works fine, @imantahriri!

A few things, though:

- There are a few linter errors. I can fix them if you want.
- Perhaps the _extendedLegend_ parameter should be renamed. I know you've already changed the name from _legendIsMultiRow_, but I feel it is still a bit misleading. I mean, the legend isn't really extended in any way. Perhaps _hasLegendGraphic_ or _isThematic_ would be more accurate.

What do you think, @daRasmussen , @markus659 , @jokd ?

You can use https://karta.sundsvall.se/bxkarta/index.json for testing. Just replace all occurrences of _legendIsMultiRow_ with _extendedLegend_.

imantahriri:
Cool.

- Yes, that was what we agreed on. I actually should only have pushed the branch.
- Sure, just tell me what it should be and I fix it.

markus659:
Thanks for the PR. I tried it with the config you provided @tonnyandersson, but some of the images are still very much at such a small size, making the texts contained in the images illegible, is this expected behaviour (fastighetsgränser for instance)?

src/utils/legendmaker.js
```diff
@@ -109,15 +120,28 @@ export const renderLegendItem = function renderLegendItem(svgIcon, label = '') {
           </li>`;
 };
 
+export const renderExtendedLegendItem = function renderExtendedLegendItem(extendedLegendItem) {
```
markus659: Could this function be merged with renderLegendItem instead (so that there is just one function)? They are quite similar.

tonnyandersson:
@markus659 You are right about the image sizes. I can't say it's the _preferred_ behaviour, but that is what you get when you try to squeeze a larger legend graphic into a legend with a fixed size. We'll have to deal with this somehow. I have come up with three potential solutions for this:

- Set a max size requirement for legend graphics (not very practical)
- Resize the whole legend to fit the legend graphics in its native size (not very pretty)
- Implement a "zoom" function that would open the legend graphics in a temporary popup (or something) on click or hover

Thoughts?

imantahriri:
@tonnyandersson @markus659 
I agree that the first one is not very practical and the second one not very pretty. And the third one seems like an overkill to me!

tonnyandersson:
Any other suggestions, @imantahriri ?

imantahriri:
@tonnyandersson Not really, sorry! :(

src/utils/legendmaker.js
markus659: I was thinking you could set the lines that are changed to a variable instead, so that the variable would have the correct value when the function is called, or set it is a parameter perhaps.

markus659:
Setting a max size would be a viable option to me, although maybe not the optimal solution. It is difficult, because you are trying to fit an image in a space that won't really allow it, so maybe you should start there.

I think if the zoom function is implemented in way that looks good it might work. Hovering over the item you would get an enlarged version. I do think this needs to be solved. 

Is there an issue related to this PR btw?

tonnyandersson:
Yes, there's an old issue related to this PR - issue #394.

@imantahriri, can you please reference the issue above in your first comment?

tonnyandersson:
I'm working on a fix for the size issue mentioned above. Got some fine tuning left to do, but besides that, I think it's pretty much done.

Despite that, I would like to see this PR merged first. When merged, I will submit a new issue regarding the size thing.

Everyone okay with that?

jokd:
Sure. Haven't tested it though ;)

src/utils/legendmaker.js
tonnyandersson: @markus659, @imantahriri, what's your conclusion?

imantahriri: @tonnyandersson I think it is good as is.

PR #615, opened 2019-05-07
Add button to hide splash

### LINKED ISSUE #591: Option to hide Splash
It would be nice to have an option in Splash to hide it the next time a user visit the map. If the splash text is changed it will automatically be visible again.

Namnlös

markus659:
This fixes #591. It is configured like so:

```json
{
  "name": "splash",
  "options": {
    "url": "test.html",
    "hideButton": {
      "visible": true,
      "hideText": "Visa inte fler gånger",
      "confirmText": "Vill du verkligen inte se den här informationen igen?"
    }
  }
}
```

The only thing needed is the "visible" option, the text configurations are optional.

### DIFF
--- a/src/controls/splash.js
+++ b/src/controls/splash.js
@@ -1,33 +1,97 @@
-import { Component, Modal } from '../ui';
+import { Component, Button, Modal } from '../ui';
 
 const Splash = function Splash(options = {}) {
   const defaultTitle = 'Om kartan';
   const defaultContent = '';
   const cls = 'o-splash';
   let viewer;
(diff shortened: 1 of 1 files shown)


jokd:
Nice! Is it possible to somehow override the users choice of hiding the splash if you for instance change the content and want it to be shown again?

markus659:
Yep, that's how it works, as long as the content stays the same it will remain hidden. If the content is changed, it will display again. The user is just making a choice to not see the current information anymore, as soon as there is new content, it will be displayed again.

Issue #606, opened 2019-04-24
Files in conf folder must be removed


imantahriri:
Default controls are initialized in Origo.js now and conf/origoConfig.js is no longer in use.
Controls are initialized in src/controls.js and conf/origoControls.js is no longer in use.

Those files in conf folder are just making confusion and serve no purpose.

daRasmussen:
Agree!

PR #616, opened 2019-05-09
Removed redundant conf folder

tonnyandersson:
Removed conf folder since the files are no longer in use. Solves #606.


markus659:
Looks good to me.

tonnyandersson:
Also removed controlinitialiser.js since controls are now initialised in origo.js.

Issue #395, opened 2018-01-18
Extended share map function


tonnyandersson:
Function to share the map with current settings (zoom, center, layers, layer visibility etc.). Map state is saved to a folder using OrigoServer. The returned URL contains an UUID referring to the saved map state.

Based on an older version of the Origo share map function and further developed for the Open eMap version of Origo.

PR #618, opened 2019-05-10
Extended sharemap

tonnyandersson:
This PR provides extended share map functionality. Solves issue #395.

Main features:

- Ability to save map state to server, including drawn (requires [draw-plugin](https://github.com/origo-map/draw-plugin)) features
- Provides short and discrete share map permalinks

These features require a web service capable of receiving a JSON string, store it with an ID and return the same JSON string upon request with the given ID.

**Configuration**
```
{
  "name": "sharemap",
  "options": {
    "storeMethod": "saveStateToServer",
    "serviceEndpoint": "URL-to-web-service"
  }
}
```

https://gis.hudiksvall.se/mapstate can be used temporarily as serviceEndpoint URL for testing purposes.


daRasmussen:
Is it possible to provide a full config, for testing?

tonnyandersson:
@daRasmussen Sure. Try [this](https://gis.hudiksvall.se/examples/ext_sharemap.json).

tonnyandersson:
The "extended" part of this feature does indeed require origo-server. I will submit a PR for the necessary parts to the origo-server repo when/if this PR gets accepted.

However, this feature is just an extension of the original share map feature - you can run it perfectly fine without origo-server, the same way it works in Origo today.

I don't think the number of file changes should be a deciding factor regarding plugins. In this case, for example, the permalink features are used extensively, which happens to involve three separate files. Also, since the extended functionality is tightly integrated with the core share map functionality, making a plugin out of this would probably require unreasonable amounts of work.

Had this been a new, rarely used standalone function, I would've agreed with you about making a plugin out of it. But this is nothing more than an extension of an already existing feature, which, to me, does not merit a full plugin overhaul.

src/loadresources.js
```diff
@@ -4,6 +4,27 @@ import getUrl from './utils/geturl';
...
+function getQueryVariable(variable) {
+  const query = window.location.search.substring(1);
+  const vars = query.split('&');
+  for (let i = 0; i < vars.length; i += 1) {
+    const pair = vars[i].split('=');
+    if (decodeURIComponent(pair[0]) === variable) {
+      return decodeURIComponent(pair[1]);
+    }
+  }
+  console.log('Query variable %s not found', variable);
```
daRasmussen: I get this prompt even when I don't use extended sharemap functionality.

src/loadresources.js
```diff
@@ -92,7 +113,29 @@ const loadResources = async function loadResources(mapOptions, config) {
             map.options.map = json;
             map.options.params = urlParams;
             map.options.baseUrl = baseUrl;
-            return map;
+
+            /* We need to find the setting from sharemap to get the serviceEndPoint
```
daRasmussen: Maybe remove these comments?

src/permalink/permalinkparser.js
```diff
@@ -54,5 +59,23 @@ export default {
...
+    }
+    return controls;
+  },
+  controlDraw(drawState) {
+    const features = new GeoJSON().readFeatures(drawState.features);
+    return { features };
   }
 };
+
+// We define the controls parse functions so they can be called from parseControlls funktion
```
daRasmussen: Maybe remove these comments as well?

daRasmussen:
@tonnyandersson good answer. It's ok if you want to merge this but I still think this code is to specific and to me it is overkills a pretty simple problem. But hey it's always nice with new features in Origo  :+1:

tonnyandersson:
@daRasmussen I'll remove the comments and add something that makes the '...not found...' prompt go away.

It may look like overkill, but you have to consider the fact that it provides a bunch of nice features:

- Shorter, more discrete permalinks
- Ability to save features drawn with the draw plugin (without making the visible permalink two miles long)
- Ability to save and store map state on server...
  - ...which also makes it possible to change map state without changing the permalink
- This infrastructure also provides a foundation for further development, for example custom named map states, storing other plugin states and whatnot

tonnyandersson:
@daRasmussen Will you take a quick look at the latest changes before I merge?

PR #629, opened 2019-05-17
added time as input type

jokd:
similar config as date and:
"type": "time"

### DIFF
--- a/src/controls/editor/editform.js
+++ b/src/controls/editor/editform.js
@@ -57,6 +57,18 @@ const createForm = function createForm(obj) {
       }
       el = `<div><label>${label}</label><br><input type="date" id="${id}" placeholder="YYYY-MM-DD" value="${val}"></div>`;
       break;
+    case 'time':
+      if (!val) {
+        if (obj.defaultTime === false) {
+          val = '';
+        } else if (obj.defaultTime) {
+          val = obj.defaultTime;
+        } else {
+          val = isoDate.slice(11, 16);
+        }
+      }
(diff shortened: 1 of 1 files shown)


tonnyandersson:
Does not work very well in IE 11, it acts like a regular text field with no up/down arrow. Works better in Edge, but there you've got a scrollable dropdown thing that could use a little TLC style wise.

Any suggestions for IE?

jokd:
IE? Just ignore it is my suggestion. There are polyfills like https://dan503.github.io/time-input-polyfill/ but I think that the placeholder is enough guidance for those who use unsupported browsers.

tonnyandersson:
@jokd I'm cool with that. Then my only question is whether the placeholder should be "hh:mm" or "tt:mm" by default. Most (all?) of the GUI text is in swedish by default. Maybe this one should follow suit?

jokd:
@tonnyandersson unless we are aiming for world domination I guess we could use swedish but then that should go for all the placeholders.
Either way works for me. Others?

tonnyandersson:
@jokd Consistency is the key to success, they say. I say we should aim to use swedish in all default GUI texts and placeholders, at least until we internationalize Origo.

jokd:
@tonnyandersson ok, I update all placeholders in this file

daRasmussen:
Why not?
`var lang = navigator.language `

jokd:
hmm... I like the thought of it but that is another issue I think. Along with different themes.

daRasmussen:
@jokd I, Agree!

Issue #587, opened 2019-03-18
Lefthand tools top-margin


mulfvik:
There is a little bit of margin at the top on the zoom-buttons that makes it not aligned with search box and menu.

jokd:
You are right, the zoom button container should have no margin-top.

tonnyandersson:
@mulfvik, @jokd Agree with both.

Also, the gap between the measure control and the geopositioning control is wider than the gap between the other controls. In my opinion, the gap should be equally wide between all controls. Perhaps that could be addressed in this issue as well.

![image](https://user-images.githubusercontent.com/15725608/54680401-03fa4400-4b0a-11e9-895d-c3eba918326d.png)

mulfvik:
There is a new PR for this issue  #711

PR #633, opened 2019-05-18
lefthandtools-marginfix, Fix for issue #587

mulfvik:
#587

### DIFF
--- a/scss/ui/_flex.scss
+++ b/scss/ui/_flex.scss
@@ -17,6 +17,7 @@ $flex-grow-all: 2;
   }
   &.column {
     flex-direction: column;
+    margin-top: 0;
   }   
   &.column-reverse {
     flex-direction: column-reverse;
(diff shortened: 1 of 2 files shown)


mulfvik:
@daRasmussen @jokd @tonnyandersson Any thoughts on this ?

scss/ui/_flex.scss
```diff
@@ -69,4 +69,4 @@ $flex-grow-all: 2;
   .align-stretch {
     align-items: stretch;
   }     
-}
\ No newline at end of file
+}
```
daRasmussen: is this change necessary?

mulfvik: @daRasmussen I guess it was a new line at the end issue together with margin change it shows that row 72 was change but it was actually a new row added on row 73 at first commit. 
So to answer your question: yes and no `:)`

PR #639, opened 2019-05-20
Wms style object fix

imantahriri:
styleSettings objects caused the whole application to crash if no style is provided in the index.json.
Related issue: #611


daRasmussen:
Which issue is related to this PR?

imantahriri:
issue #611 which is already merged

PR #671, opened 2019-07-03
Footer update

daRasmussen:
fixes issue 670 and updates footer param logic.


jokd:
What about removing the urlText and just have text? If combined with a url then its a link, otherwise not. In your code you need to use urlText if you want image and text combined even if you have no url, which is not really obvious.

src/components/footer.js
```diff
@@ -9,10 +9,18 @@ export default function Footer(options = {}) {
   return Component({
     render: function render() {
       let middleContent = '';
-      if (data.img) {
+      if (data.img && !data.url && !data.urlText) {
```
daRasmussen: This statement is the same as original could be removed

jokd: Agreed

src/components/footer.js
```diff
@@ -9,10 +9,18 @@ export default function Footer(options = {}) {
...
+        middleContent = `<img src="${data.img}">`;
+      } else if (data.img && data.url && !data.urlText) {
+        middleContent = `<a href="${data.url}"><img src="${data.img}"></a>`;
+      } else if (data.img && !data.url && data.urlText) {
+        middleContent = `<img src="${data.img}">${data.urlText}`;
+      } else if (data.img && data.url && data.urlText) {
+        middleContent = `<img src="${data.img}"><a href="${data.url}">${data.urlText}</a>`;
```
jokd: Should the image be a link as well?

src/components/footer.js
```diff
@@ -9,10 +9,18 @@ export default function Footer(options = {}) {
...
+      } else if (data.img) {
         middleContent = `<img src="${data.img}">`;
       } else if (data.url && data.urlText) {
-        middleContent = `<a href="${data.url}">${data.urlText}</a>">`;
+        middleContent = `<a href="${data.url}">${data.urlText}</a>`;
       } else if (data.text) {
```
jokd: How about replacing text with urlText (or better the other way around)?

src/components/footer.js
daRasmussen: sure, why not 👍

src/components/footer.js
daRasmussen: yeah for sure urlText, although I think this was text is from the original code so not sure if it was intended to be supported.

jokd: However in case you have no url the urlText is a bit misleading

daRasmussen: jupp looking at the docs, that is true. Maybe update docs with an statement that urlText is depended that the url for a link is set.

jokd: yes, and that you cannot combine text with other elements. If you want image and text, you must use urlText.

daRasmussen: Jupp! 

I will referense this conversion in an issue on the api-docs repo.

tonnyandersson:
@daRasmussen I am unable to test this due to some local git issue preventing me from viewing external branches.

However, I agree with @jokd that the name _urlText_ is a bit misleading. I would like to suggest that we use three parameters - _img_, _url_ and _text_.

I can come up with six different scenarios, off the top of my head:

* img - shows an image only
* text - shows text only
* img + text - shows an image and text
* img + url - shows a clickable image
* text + url - shows clickable text
* img + text + url - shows an image with text, both clickable

I think I got them all, right?

This will probably screw things up with old config files, but I think it's worth it because it makes things more intuitive.

jokd:
Closing this since #695 solves the issue

PR #503, opened 2018-09-12
Option to autotoggle subgroups

jokd:
Fixes #245

PR #674, opened 2019-07-09
Autotoggle subgroups

### LINKED ISSUE #245: Submenu could open when activating all layers in submenu.

jokd:
@mulfvik
Fixes #245, rewrote #503
By default you expand/collapse subgroups when ticking the subgroups checkbox. However you can override this by setting `"autoExpand": false` for a certain group.

### DIFF
--- a/src/controls/legend/group.js
+++ b/src/controls/legend/group.js
@@ -16,7 +16,8 @@ const Group = function Group(options = {}, viewer) {
     name,
     parent,
     position = 'top',
-    type = 'group'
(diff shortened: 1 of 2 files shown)


tonnyandersson:
@jokd LGTM. Not sure about having it as the default behaviour, though. But it seems everyone has agreed on that since way back, so...

jokd:
@tonnyandersson I merge this now, making the behaviour configurable can be another issue.

Issue #631, opened 2019-05-17
remove image in editor


jokd:
It is not possible to remove images in the editor form. Also, when uploading a new image the old one is still visible even though it is replaced (as you can see by closing and opening the form)

jokd:
Fixed with #682

PR #682, opened 2019-09-23
Image as attribute

### LINKED ISSUE #678: text problem for editable layer with image
What we would like to do is that we want to use the editor to save an image (as an attribute in database table) for a point layer to our Postgres database through the formula:
  
origo1

For saving images in editing mode, we define the image attribute (“foto”) for the point layer in the configuration json file like this:
		{
	"name": "foto",
	"title": "Ladda upp bild: ",
	"type": "image",
	"image": "foto",
	"maxSize": 800 
		}
The image now is saved in the database as a very long text.  Now it comes to the problem. 

In Orgio 1, when I turn off the editing mode  and click a point from this point layer, the image is not previewed in the information box, the image just appear as text: 
 
image

In order to show preview image for the point layer in the information box, I have to create another map layer for the points and declare the attribute in another way:
	{
		"img": "foto"
		}
 
image

That means I cannot use one single point layer in editing mode och viewing mode in the map. I need to have 2 layers for the same thing in order to show the image. 

In Origo 2, it is even more problematic. 

I want to edit attributes for some existing points while the editing mode is on. I click a point and both the editing formula and the information box in viewing mode come up in the map (this behavior does not exist in Origo 1. In Origo 1, we cannot get any information box for any layer while the editing mode is on). 
 
image

At the same time, if the point has an image, the image in the information box is shown as a very very long text (unlike Origo 1 which has the text length limited in the information box). This mean that when I click a point for editing, first the information box comes up and it extends and moves the map view downwards far away from the point.
  
image

To go around this problem, I have to first make the editable point layer non queryable (so that no information box comes up with a long text) and then create another queryable layer for the same points in order to fulfill the image preview purpose. 

It would be very nice if we can use only one single layer that can show the preview image in editing mode or view mode instead.

jokd:
Fixes #678 
Now (img: {attributename}) will work the same way as (type:image, name:{attributename}) to work in form as well as popup.

### DIFF
--- a/src/controls/editor/edithandler.js
+++ b/src/controls/editor/edithandler.js
@@ -408,20 +408,19 @@ function addImageListener() {
     $(obj.elId).on('change', () => {
       $(`${containerClass} img`).removeClass('o-hidden');
       $(`${containerClass} input[type=button]`).removeClass('o-hidden');
-
-      if (this.files && this.files[0]) {
+      if (event.currentTarget.files && event.currentTarget.files[0]) {
         fileReader.onload = (e) => {
           $(`${containerClass} img`).attr('src', e.target.result);
         };
 
-        fileReader.readAsDataURL(this.files[0]);
+        fileReader.readAsDataURL(event.currentTarget.files[0]);
       }
     });
 
     $(`${containerClass} input[type=button]`).on('click', () => {
       $(obj.elId).attr('value', '');
       $(`${containerClass} img`).addClass('o-hidden');
-      $(this).addClass('o-hidden');
+      $(event.currentTarget).addClass('o-hidden');
     });
   };
 
--- a/src/controls/editor/edithandler.js
+++ b/src/controls/editor/edithandler.js
@@ -405,22 +405,24 @@ function addImageListener() {
   const fn = (obj) => {
     const fileReader = new FileReader();
     const containerClass = `.${obj.elId.slice(1)}`;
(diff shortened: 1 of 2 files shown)


tonnyandersson:
@jokd Unable to test this properly since I don't have an adequate environment set up. Tried with the release test example config but wasn't able to save images, probably due to string length restrictions in the database.

But, from what I could see without saving images, this seems to work fine, with the exception that `"img": "attributename"` does not work in the edit form. But perhaps it's not meant to?

I also found a bug in edithandler.js that throws an error when trying to upload images in the editform.

https://github.com/origo-map/origo/blob/24225fdddb1c1d46e63b31417e43c492a4736331/src/controls/editor/edithandler.js#L408-L419

`this` (as in `this.files`) is not a thing in arrow functions and should be replaced with `event.currentTarget` or the arrow function should be replaced entirely with a standard function. Perhaps a completely different issue?

jokd:
@tonnyandersson Added your fix so this PR will solve #631 as well

tonnyandersson:
@jokd Oh, I missed one thing. The `event` variables should be declared as arguments to the functions before being used.

jokd:
@tonnyandersson Please push a fix

tonnyandersson:
Ain't letting you off the hook just yet. :)

Please note that there are two different events here (actually, there's three but `filereader.onload` is not affected by this). I don't think they should refer to the same event target.

image

PR #695, opened 2019-10-02
Fix for footer

### LINKED ISSUE #670: Footer not showing img/url
Use of "footer" under pageSettings in config file has no effect.

"footer": {
  "img": "img/png/logo.png",
  "url" : "https://github.com/origo-map/origo",
  "urlText": "Origo"
}

Using same config file with older version origo version works.

jokd:
Fixes #670 , replaces #671


tonnyandersson:
Works like a charm, @jokd! 

One thing, though. I would prefer if the linked page opens in a new tab. Besides that, this one is good to go.

tonnyandersson:
By the way, are the changes to package-lock.json necessary for this PR?

PR #696, opened 2019-10-02
Added number and hidden as input types

jokd:
This PR is another way to solve #552 and introduces the number and hidden input types for the editor form. 
Along the hidden type it would be nice to introduce some default values/functions to populate the field. Eg todays date.

### DIFF
--- a/src/controls/editor/editform.js
+++ b/src/controls/editor/editform.js
@@ -6,26 +6,28 @@ const createForm = function createForm(obj) {
   const label = obj.title;
   let val = obj.isVisible && obj.val != null ? obj.val : '';
   const type = obj.type;
-  const maxLength = obj.maxLength ? ` maxlength="${obj.maxLength}" ` : '';
+  const maxLength = obj.maxLength ? ` maxlength="${obj.maxLength}"` : '';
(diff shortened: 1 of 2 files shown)


tonnyandersson:
@jokd Shouldn't the "hidden" attribute be hidden in the featureinfo popup as well?

image

image

jokd:
Well, not sure about that. I guess it depends, perhaps make it configurable? Or we add the "readonly" property for input types that should be autopopulated and shown and keep the hidden field always hidden?

tonnyandersson:
I just don't like the inconsistency in having something invisible in one place and visible in another. Perhaps the latter then, with "readonly" as a slightly greyed out, non-editable field and "hidden" as really hidden?

jokd:
@tonnyandersson Now its totally hidden, I hope

tonnyandersson:
@jokd Resolve the conflicts and I think we're good to go.

And please provide a simple config example next time. Doesn't have to be a complete config file, just a few lines showing the relevant settings. Guessing takes time... :)

jokd:
@tonnyandersson Sure, sorry about that

PR #697, opened 2019-10-02
Fixes #693

### LINKED ISSUE #693: No group for layers using drag'n'drop
Layers using drag'n'drop doesn't belong to a group. With deafult or option "groupTitle".

dragndrop_problem

jokd:
Solves issue where drag-and-dropped layers always turned up in the root of the legend

### DIFF
--- a/src/controls/draganddrop.js
+++ b/src/controls/draganddrop.js
@@ -8,7 +8,7 @@ import VectorSource from 'ol/source/Vector';
 import VectorLayer from 'ol/layer/Vector';
 import { Component } from '../ui';
 
-const DragAndDrop = function DragAndDrop() {
+const DragAndDrop = function DragAndDrop(options = {}) {
   let viewer;
 
   return Component({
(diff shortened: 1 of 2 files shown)


jokd:
This PR introduces the groupName option, If set it will add the layer to the group if it exists, otherwise it will create the group. If set, groupTitle will be used when creating a new group.
Configuration:
{
      "name": "draganddrop",
      "options": {
        "groupTitle": "Mina lager", //defaults to Egna lager
        "groupName": "mina-lager" //defaults to egna-lager
      }
    }

tonnyandersson:
Works really well!

But!

The drag-and-drop group should, preferably, not be created/visible from the start, but rather be created/visible when the first drag-and-drop layer is added to the map. The group should also be deleted/hidden when it is empty.

jokd:
I can agree on the former and have changed the behaviour. Removing empty groups takes a bit more effort and can be addressed in another issue I think.

jokd:
@tonnyandersson Would removing the group when empty be the default behaviour when removing layers, or should this just apply to draganddrop-layers? I suggest the former, for instance when using the layermanager it makes sense.

tonnyandersson:
@jokd I agree, removing empty groups should be the default behaviour. I can't see why you'd want a bunch of empty groups in your legend.

tonnyandersson:
LGTM

PR #698, opened 2019-10-04
Fixed @center coordinate transformation bug

### LINKED ISSUE #687: Attribute option html @center coordinate transformation does not seem to happen
1. Get and install lastest origo, master branch
2. Edit the default index.json so that origo-cities defines the info popup with the 'html' option and center coordinates transformation as suggested in https://origo-map.github.io/api-documentation/latest/#attributes , for instance: 

`    "layers": [
        {
            "name": "origo-cities",
            "title": "Origokommuner",
            "group": "root",
            "style": "origo-logo",
            "type": "GEOJSON",
            "attributes": [
                {
                    "html": "<p>For more information - {{@center(EPSG:4326,reverse)}}</p>"
                }
            ],
            "visible": true
        },`
3. npm run start
4. Info click an Origo hotspot

 _Expected_:  lat/lon or planar coordinates in whatever system you specced at 'EPSG:..'
(from the epsg:3857 in the data/origo-cities-3857.geojson )

_Shown_: epsg:3857, reversed

In my current laboration environ src/geometry/getcenter.js never appears to reach what's inside
`if (destination && map)` 
as map is undefined. That might be a user error but as far as I'm aware I'm running a default application and trying out an option as detailed in the latest documentation, please advice if not :) 

The problem currently also manifests itself in https://www.malardalskartan.se/turism and karta.strangnas.se  - badplatser   -- "Hitta hit"  

Thanks for getting to the end of this rather long report.

tonnyandersson:
Fixes #687.

Can be tested using the default config while adding the snippet below as an attribute to the 'origo-cities' layer.

```
{
  "html": "<a href='https://www.google.se/maps/dir//{{@center(EPSG:4326,reverse)}}/@{{@center(EPSG:4326,reverse)}}' target='_blank' style='float: right'>Directions &raquo;</a>"
}
```

This PR is a quick fix to solve the issue mentioned above. But it only works if two arguments are passed to the @&#x200B;center helper function. E.g. 'EPSG:4326' and 'reverse' in the snippet above. If the 'reverse' argument is omitted, an empty argument must be passed to the function, like this (note the trailing comma separator):

```
@center(EPSG:4326,)
```

This is, in part, because of the way the searchAndReplace function is written, including a helper function call with flexible number of arguments. There are a few ways around this, but I found it hard to come up with something pretty that doesn't require rewriting everything related to the searchAndReplace function.

Suggestions are welcome.

### DIFF
--- a/src/geometry/getcenter.js
+++ b/src/geometry/getcenter.js
@@ -7,7 +7,6 @@ export default function getCenter(geometryIn, destination, axisOrientation, map)
     geometry.transform(map.getView().getProjection(), destination);
   }
 
-
   const type = geometry.getType();
   let center;
   switch (type) {
--- a/src/geometry/getcenter.js
+++ b/src/geometry/getcenter.js
@@ -1,5 +1,3 @@
-import viewer from '../viewer';
-
 export default function getCenter(geometryIn, destination, axisOrientation, map) {
   const geometry = geometryIn.clone();
 
--- a/src/utils/replacer.js
+++ b/src/utils/replacer.js
@@ -13,7 +13,7 @@ const replacer = function replacer() {
     return ['', str];
   }
 
-  function searchAndReplace(name, obj) {
+  function searchAndReplace(name, obj, map) {
     const regex = new RegExp(`${start}(.*?)${end}`, 'g');
     const matches = regex.exec(name);
     if (matches) {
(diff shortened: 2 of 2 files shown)


jokd:
I can live with that fix. But for clarity, perhaps the arguments (documentation-wise) should be reverse (or false?) and true instead of empty?

tonnyandersson:
@jokd Definitely! I'll look into it.

tonnyandersson:
@jokd I can update the documentation specifying ```default``` or ```reverse``` as required arguments. If that is a passable solution, there's nothing that needs to be added to this PR.

jokd:
Lets get this merged then.

Issue #692, opened 2019-09-27
Clear button in the legend


asemoller:
The ability to turn off all visible layers at once.

Grammostola:
Hi, we have one of those buttons: https://github.com/Eskilstuna-kommun/EK-extern/commit/f69894967d8426ae04b10ad870889e4ef1577c20 , I haven't investigated how compatible it is with the core yet but it is a UI component at any rate and does not look overly complicated. Can try attaching it to a core Origo on Tuesday if no one has commented to the contrary here before then : )

asemoller:
Nice @Grammostola! looking forward to it :+1:

Grammostola:
It introduces a new control component and from some preliminary testing doesn't seem tied to any variations of the core that we (Eskilstuna) are using. It isn't an on/off button but merely an 'off'. The commit does alter a few core files(overlays - menu width) and legend.js (the new button) and as it currently looks doesn't turn off layers that are set to visible in the config.json / from the start. I assume what you are looking for is 'Turn off all visible layers (except backgrounds) at once' and that's not hard to change.
It currently has a paper bin icon, see the attached screenshot. 
![bin](https://user-images.githubusercontent.com/5138906/65967912-b2405800-e462-11e9-9cec-2845ad3a0342.PNG) 
which could of course also be altered. 
Further comments may advice whether this might after some possible alterations be the subject of a pr or whether a dissimilar solution is preferred.

PR #700, opened 2019-10-08
Implements 'turn-off-all-layers'-functionality

Grammostola:
Rough draft proposal for issue  #692 meant to serve as a basis for discussion : ) Here's an essay to get us started:

- Currently the Layermanager blue add button appears on the same toolbar row as the background maps and as this Turnofflayers button, an alternative is shown in attached image one (Eskilstuna's current version) (where the responsibility of creating the button and accompanying text is moved to the Layermanager plugin from legend.js) (
eskilstuna
) (I've replaced the paperbin icon in this PR). What this means is that fewer background maps fit before the legend pane starts to look funny: 
4bm_turnoff_lm
 (_but the number of background maps / controls before the panel starts looking funny is potentially a separate issue_ . Altering this src/legend/overlays.js row:
`  const style = dom.createStyle(Object.assign({}, { width: '220px', height: '100%', 'min-width': '220px' }, styleSettings));` to a greater fixed width increases the width of the legend pane and allows more items in the background toolbar row but it is a compromise. "Auto" makes the legend pane resize when legends are clicked which can be a bit jarring.)

 It also means that I've replaced the "addControl" property of legend.js with addLayermanagerControl and addTurnoffLayersControl properties. 

-If merely one background map is employed the toolbar looks potentially funny too: 
1bm_turnoff_lm

-Currently the addTurnoffLayersControl: true property is as such set in index.json for the legend control; the control is also included on its own. This also means that if this property is true for the legend control but the control isn't included too the turnofflayers-button is present but doesn't do anything visible (same as for the Layermanager button if that plugin isn't present)

-'secondDivider' is questionable

-If there happens to be four background maps and no Layermanager button on the same row things look rather swell:
4bm_turnoff

(This is a slightly reworked version of the original control developed for Eskilstuna's spork: https://github.com/Eskilstuna-kommun/EK-extern/commit/f69894967d8426ae04b10ad870889e4ef1577c20 by Bimane1900 and Zakuto3 )


asemoller:
I haven´t actually test this yet but from your comment and images it looks very good @Grammostola! This is what we're looking for.

I personally like the "visible_off" icon better and the placement was my other concern but it seems like you solved that (?). 

Only one question remains for us: whether or not all visible layers except background map(s) should be turned off or if the function returns the map to the default setting.
For the everyday user the latter suggestion is probably more user-friendly and less confusing even though I can see use in a function that actually turns off all visible layers regardless the default settings.

CBEskilstuna:
lager

Grammostola:
@asemoller Good : ) It was discussed today and a 'restore to map default' would be more of a reset, with perhaps a different icon, so this PR will be about a function that turns of all visible layers except the background one. Will provide some new pictures when ready, one thing that might be commented upon before is whether the "tools row" above the background maps row, as seen above, would fill from the left, the right, or try to push 'add layers' to the right and 'turn off layers' to the left (and when a new tool comes along we'll decide then how to proceed).

asemoller:
Ok, so you changed the behaviour for the button from your eskilstuna fork? Either way works for us.

I think that since "Layer Manager" is a plugin and the button for this PR is PR Accepted I prefer that the "Clear button" is on the same toolbar row as the "background maps" and in the same column as the "layers visible button" to the right (see in image 1 and 4 above).
I can understand the problem with having more than three "background maps", but that can be solved by just changing settings for the width in the config. Either you have many background maps or you have one function button, if you want both - make the legend larger.

The "Layer Manager" could either be as is (see image 1) or as a button in the same row as the "Clear button". 

Maybe there should be a new row with function buttons but that will look odd if you only have one background map and one function button. When there are more than two function buttons fighting for the same place, we should solve that in another issue as you mentioned.

Regardless where you put the button, I really like the function :).

This is what I prefer (with LM) using "width": "280px" for the legend:

image

Without LM:
image

jokd:
@Grammostola It looks nice and works well, my only concern is the placement. 
I think that there is a risk that the not visible-icon in the same spot as the menu opener can be interpreted as a close button. I'd prefer if we in the lower right had a close button like in the upper menu and perhaps moved the clear button to the top like this (pardon my lack of Paint-skills):
image

tonnyandersson:
@jokd I think it would look a little bit cleaner if the divider is left out when there's only one item in that top (or bottom) row.

image
image

asemoller:
You have a point @jokd!
If you put a close button in the legend, we ought to remove the arrow at the top of the legend as it would then be redundant. 
Having a "close button" instead of the arrow is preferable as this is a well established icon language.
(One idea is maybe keep the upper arrow, but change its functionality. Maybe to collapse the legend down to background maps? this would of course be another issue.)

Another issue to consider might be adressing the inconsistencies of the legend design (such as mixtures of square and round buttons or left vs right positioning of arrows).

jokd:
@asemoller I have a PR for the close button, #704 . I agree on the inconsistencies of the legend design and welcome a discussion.

Grammostola:
I hear you, the eye aligned horizontally with the layer on/off looks good. Hm, everything except the background toggle group is handled in an legend/overlays component, at a first glance one could implement the turnofflayers-button there but it may not be optimal if one wishes specify its inclusion as an option for the legend control in the json (haven't tested yet). ( Speaking of specifying things in the json, @asemoller - are you changing the width of the legend via the style option? ) I'll investigate further and maybe we can discuss this on Wednesday : )

asemoller:
@Grammostola - yes, like this:
  `   "name": "legend", "options": {"expanded": false, "contentStyle": {"width": "280px"}}`

@jokd, @Grammostola - I guess we can discuss this on Wednesday then :)

Grammostola:
I'll try to show an example early next week. One visual go-ya is that the scrollbar, which we agreed should not scroll the eye, pushes the overlay with the layer list to the left so if the eye is aligned with the turn off/on layers buttons from the start it isn't after the scrollbar appears. But I'll try and make a video : )

jokd:
Tried it and I think it looks nice!

Grammostola:
Current state:

inget_dera
layermanager_
turnofflayers_layermanager_
turnofflayers_

There's no text (is text preferrable?) and the separator line between the root group and the background bar is missing if no tools are enabled. (it collided with the close button when I enabled it). It doesn't precisely look bad, merely a bit inconsistent. Making the legend wider via Åse's control style option the tool buttons are still aligned with their above UI elements.

jokd:
I think I prefer if the group title bar doesnt have a border-bottom when expanded as Layers2 in this example, compared to the group Layers:
image
That way it is easier to see that the layers belong to the group

Grammostola:
Hi, right. Removing the border altogether from the collapseheader is the easy way: 
image

I don't think we're doing the easy way here though so will find a way to deal with the collapseheader component's expand method when employed by group.js. Please stand by..

Grammostola:
@jokd Hi, the separator line for group titles has been altered, see the movie below for that and how the turnofflayers button works currently.
grupper_släcka

-Group titles now have no separator line when expanded
-Subgroup titles now have no separator line when expanded. (By default they have no separator line when not expanded, this remains unchanged)

(I tinkered a bit with putting the bottom separator line in class collapse and collapse-container but didn't find an immediate solution there. This works, anyways :)  )

jokd:
@Grammostola I like it this way!

PR #704, opened 2019-10-14
Adds close button

### LINKED ISSUE #703: Close button in the legend
It would be nice to be able to close the legend with a button in the lower right.

jokd:
Fixes #703
Adds a close button in the lower right, similar to the one in the map menu.
image

### DIFF
--- a/src/controls/legend.js
+++ b/src/controls/legend.js
@@ -20,6 +20,7 @@ const Legend = function Legend(options = {}) {
   const backgroundLayerButtons = [];
   let toggleGroup;
   let layerSwitcherEl;
+  let closeButton;
   let layerButton;
   let layerButtonEl;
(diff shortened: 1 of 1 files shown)


johnnyblasta:
The alignment of the icon bugs me a bit. Would perhaps prefer aligned at top with layer icons and that the icon could be a bit smaller.

jokd:
@johnnyblasta I wanted the same behaviour as the mapmenu, as well as positioned at the same spot as the layerbutton so you can toggle the legend without moving the cursor. But I can agree it looks a bit misaligned the way it is. Perhaps we should rework the mapmenu as well?

tonnyandersson:
LGTM.

johnnyblasta:
LGTM

PR #708, opened 2019-10-15
Fixes #127, layer transparency

### LINKED ISSUE #127: Set transparency on layer
Option for user to set transparency on layer.

jokd:
Idea for solving #127, adds a slider in the layer tab. No configuration at this moment so it is shown for every layer.
image

### DIFF
--- a/scss/ui/_input.scss
+++ b/scss/ui/_input.scss
@@ -57,3 +57,92 @@ input[type=button] {
 input[type=button]:hover {
   background-color: $border-color;
 }
+
+// range
+input[type=range] {
+  height: 24px;
+  width:100%;
+  -webkit-appearance: none;
+  border:0;
+  padding:0;
+  margin:0;
+}
+input[type=range]:focus {
+  outline: none;
+}
+input[type=range]::-webkit-slider-runnable-track {
+  height: 4px;
+  cursor: pointer;
+  animate: 0.2s;
+  box-shadow: 1px 1px 5px $grey-dark;
+  background: $primary-color;
+  border-radius: 3px;
+}
+input[type=range]::-webkit-slider-thumb {
+  box-shadow: 1px 1px 4px $grey-dark;
+  border: 1px solid $grey-lightest;
+  height: 16px;
+  width: 16px;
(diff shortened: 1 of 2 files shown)


tonnyandersson:
Perfect! But it should definitely be a layer setting, like `"transparenceControl": true`.

And **PLEASE** - for the love of \*insert deity of personal preference\* - give it some Origo-esque CSS love!

jokd:
@tonnyandersson Please elaborate your criticism. The native slider is beautiful in its simplicity. It even works on touch screens sometimes.

tonnyandersson:
If it works on touch screens _sometimes_ - I'm convinced! Just give it a splash of the OrigoBlu&trade; and it's good to go!

All jokes aside, I was thinking about something along these lines (pun very much intended):

image

jokd:
@tonnyandersson What should be the default behaviour? I vote for true.
OK with this: ?
image

jokd:
set opacityControl: false on a layer to deactivate the slider

tonnyandersson:
@jokd Personally, I think it would look better without the bar shadow, like in the example. I also think it looks cleaner if it is deactivated by default, but then most people won't use the feature because they won't know it exists.

So, drop the bar shadow (but keep the button shadow) and make it activated by default. My two cents.

jokd:
@tonnyandersson I dropped the box-shadow:
image

tonnyandersson:
Three thumbs up regarding the looks! Haven't tested it yet, though, but it looks great.

tonnyandersson:
Works fine in Chrome and Edge, but not in IE11 (surprise!). Input event is not triggered. Other than that, it's good to go.

jokd:
@tonnyandersson Haven´t we dropped support for IE11?

tonnyandersson:
@jokd I guess. Feels weird, though. Your call if you wanna try and fix it. I'll approve this.

jokd:
@tonnyandersson I suppose you will have to listen to _change_ event rather than _input_ to make it work in IE11. This will require either an IE exception or a change in behaviour where you must release the slider to trigger the event.

Issue #712, opened 2019-10-17
Window reload pressing enter in text input


mulfvik:
In attribute editform when hitting enter key and having cursor active in text field input, the window reloads.
Only tested when running on webpack dev server.

PR #717, opened 2019-10-23
Enterkey returns false

mulfvik:
Fix for #712


jokd:
@mulfvik When using textarea you sometimes want to be able to use Enter key. How about just removing/replacing the form tag since we anyway doesn't submit the form? Or do we?

mulfvik:
Yes, we can do that. The pressed return is stored as a whitespace in db and it is rerendered that way in attribute editor. Either way is fine for me.

PR #719, opened 2019-10-24
Makes it possible to add new attribute types

### LINKED ISSUE #718: Changes necessary for elevation-profile-plugin
In order to dynamically add new attribute types the getattributes function needs to be refactored.

jokd:
Fixes #718 - Makes it possible to add new attribute types


tonnyandersson:
Works well with the elevation profile plugin. LGTM.

PR #725, opened 2019-10-29
Added Scalepicker control

### LINKED ISSUE #393: Scale picker

tonnyandersson:
Fixes #393, replaces #481 

Adds a Scalepicker control to the left footer toolbox. The placement of the control, in relation to other controls in the same toolbox, is determined by the order of the controls in the config.

Simple configuration example:

```
{
  "name": "scalepicker"
}
```

Configuration example with options:

```
{
  "name": "scalepicker",
  "options": {
    "buttonPrefix": "Scale: ",
    "listItemPrefix": "Scale: "
  }
}
```


jokd:
Works well, but how about a subtle animation (similar to when you use the scrollwheel) when changing the zoom level?

tonnyandersson:
@jokd Definitely!

PR #736, opened 2019-11-03
Added class

### LINKED ISSUE #735: Map menu and legend on hover
When hovering the buttons should change color similar to the tools on the left hand side.

jokd:
Fixes #735


tonnyandersson:
There's a linter error on row 179 in legend.js. Not caused by these particular changes but could easily be fixed while you're at it.

The rest is fine. LGTM.

PR #737, opened 2019-11-03
Added tooltips to buttons

### LINKED ISSUE #583: Tooltips won't show
When hovering buttons, no tooltip is shown.

jokd:
Fixes #583 
Used by externalurl, measure and map menu at the moment.


tonnyandersson:
@jokd tooltipPlacement should be changed to 'east' for externalUrl as well, I think.

PR #742, opened 2019-11-07
Adds startExtent option

### LINKED ISSUE #741: Set maps start extent instead of zoomlevel
It would be nice to be able to set the maps start extent instead of using zoomlevel and center coordinate. This way it will show approximately the same extent no matter what screen size the user has.

jokd:
Fixes #741


tonnyandersson:
LGTM.

PR #743, opened 2019-11-07
Adds projection option for wfs and geojson

jokd:
Fixes # 734
Adds projection option for wfs and geojson.
Simply set "projection":"EPSG:4326" to request features in another reference system and Origo will handle the transformation. The proj4Defs has to be configured in index.json unless it's EPSG:4326 or 3857.

### DIFF
--- a/src/layer/geojson.js
+++ b/src/layer/geojson.js
@@ -7,20 +7,21 @@ function createSource(options) {
   const vectorSource = new VectorSource({
     attributions: options.attribution,
     loader() {
-      fetch(options.url).then(response => response.json()).then(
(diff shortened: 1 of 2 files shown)


tonnyandersson:
@jokd Works like a charm! But please remove the console.log() statement on line 19 in wfs.js.

MattiasSp:
How wonderful! Now Origo can read GeoJSON from a QGIS Server WFS out of the box :+1:  (since QGIS Server only serves GeoJSON in the SRS EPSG:4326 that is defined by the standard) :champagne:

Issue #323, opened 2017-10-11
Validation in editor


mulfvik:
Add validation handling to attribute input in editor, i.e if an email is not containing '@' return "Enter a proper email-address please"

PR #746, opened 2019-11-11
Validate input

mulfvik:
Address issue #323


jokd:
Are input types "decimal" and "integer" valid html? Otherwise type should be "number". Setting step="1" will only allow integers while step="any" will allow integers as well as decimal values.

mulfvik:
@jokd I guess its not not valid html but it could be hard to read so I will change that.

jokd:
Seems to do the job, can't see why we wouldn't merge this.

mulfvik:
Just one thing, one should be able to use multiple cases of types. Will handle that in next coming commit.

johnnyblasta:
Have tested the latest fix for multiple of same type and it LGTM.

PR #751, opened 2019-11-14
Markus' search suggestions fix

### LINKED ISSUE #425: Search suggestion bug

Grammostola:
Proposes to fix #425 , same as old PR #426, as agreed. (ESlint doesn't like a conditional return statement; I can't tell that this causes a functional issue however.) Movies:

search_pre_Markus
search_post_Markus

Tested with V:ås external map's search back end, clicking hits does not pan and zoom by default neither with the current search nor the slightly altered version proposed here, perhaps a trivial thing, may be that someone else is interested in testing.

Edit. Yeah, two movies of unequal lenghts autoplaying right next to each other is a bit annoying

### DIFF
--- a/src/controls/search.js
+++ b/src/controls/search.js
@@ -293,8 +293,8 @@ const Search = function Search(options = {}) {
       sort: false,
       maxItems: limit,
       item: renderList,
-      filter(suggestion, input) {
-        if (suggestion.value.toLowerCase().includes(input.toLowerCase())) {
+      filter(suggestion, userInput) {
+        if (suggestion.value.toLowerCase().includes(userInput.toLowerCase())) {
           return suggestion.value;
         }
         return false;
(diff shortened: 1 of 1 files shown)


jokd:
Works well, however the linter complains about input being redeclared on row 296 so I suggest you rename it. After that we can merge this.

Issue #754, opened 2019-11-25
"Support for older browsers"


mulfvik:
We can remove this text in README

PR #755, opened 2019-11-25
Removes support text

mulfvik:
Issue #754 
Removes the text "support of old browsers" in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,6 @@ Origo is a web mapping framework. It is based on the [OpenLayers library](https:
 ## Features
 
  * Responsive design
- * Support for older browsers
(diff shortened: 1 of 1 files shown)


asemoller:
Have we really dropped the support for ie11? Unfortunately it's still a very common browser. If we have dropped it, maybe we should remove text in the developing.md as well?

_"Origo shall support the most commonly used web browsers such as Internet Explorer 10-11, Edge, FireFox, Chrome and Safari."_

mulfvik:
The support for "older" browsers are dropped as far as I know. Its a good question and we should discuss it in arbetsgruppen what browsers is "common browsers" in this case.

PR #761, opened 2019-11-27
changed to advanced print control

### LINKED ISSUE #136: Control for client side printing
Control for client side printing with advanced options to set scales, templates and formats.

afogelberg:
Fixes #136 

Can be tried out here: https://maps.afogelberg.ninja/origo-print/


jokd:
Works great! One thought: Would it be possible to shrink the preview to fit the browser window? For instance when I select A3 as format it stretches outside what my screen presents. Especially in portrait mode.
Besides that, really nice!

afogelberg:
@jokd that will be a bit tricky to keep the correct resolution and give the correct preview. It can be done but with some effort.

PR #775, opened 2019-12-12
Resizable on o-modal class

mulfvik:


jokd:
Should this somehow be optional? Also there are no support in IE/Edge, do we bother?

mulfvik:
Would it be good to have for #777 ? Else no bother

jokd:
In that case it could be inline when creating the modal and the content is an iframe?
Its a nice feature indeed.

jokd:
@mulfvik Added your solution inline when creating the modal with iframe. Perhaps thats the only situation where you want to be able to resize the modal?

Issue #774, opened 2019-12-12
Resizable modal


mulfvik:
The modal could be resizable or scrollable. Modal should be scrollable according to #28 but it's not behaving that way.

Example on resizable modal #775

PR #777, opened 2019-12-13
Modal iframe

### LINKED ISSUE #381: Full screen modal with iframe

jokd:
Fixes #381 
Add target: modal or modal-full to open links in modal window instead of a new window. Used along with url. 
Example:

```
{
  "name": "name",
  "url": "url",
  "target": "modal-full"
}
```

### DIFF
--- a/scss/_modal.scss
+++ b/scss/_modal.scss
@@ -14,7 +14,7 @@
 .o-modal {
   background-color: white;
   border-radius: 6px;
-	bottom: auto;
+  bottom: auto;
   -webkit-box-shadow: 0 2px 8px 1px rgba(0,0,0,0.45);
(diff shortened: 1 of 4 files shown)


mulfvik:
Looks good, I close #774 #775.
If use config like this 
`{
  "url": "url",
  "html": "Läs mer &raquo; ",
  "target": "modal-full"
}`
the html value is showing in the modal top but it migth be nice if it always shows the url?

jokd:
@mulfvik Im sure there are cases where you want this behaviour but also the opposite. I guess we can make it configurable.

jokd:
@mulfvik Please have another look

mulfvik:
Lookin good @jokd

PR #781, opened 2019-12-13
Removed width in scalepicker

### LINKED ISSUE #780: Scale picker problem with character limit
When using the scale picker, the presentation of the chosen scale becomes abbreviated when string becomes too long. This becomes very clear when implementing the option "buttonPrefix": "Scale: ". I.e: 1:16000 is presented as 1:16 and the right arrow moves down a bit.

image

jokd:
Fixes #780 
Partly fixes the issue by removing the width for scalepicker. However, fitting scalepicker and position in the left side of the footer can be an issue when working on a smaller browser window/screen. The footer may need an overhaul.

### DIFF
--- a/scss/_scalepicker.scss
+++ b/scss/_scalepicker.scss
@@ -1,4 +1,5 @@
 .o-scalepicker {
+  min-width: 110px;
   margin-right: 5px;
   height: 1.25rem;
 }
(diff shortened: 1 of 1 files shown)


jokd:
Is this ok @tonnyandersson or do we lose something when dropping the fixed width?

tonnyandersson:
If there's a wide span between the smallest and the largest available scales and the width is not set, the dropdown tab will sometimes become too small when picking the largest scale (shortest string). See image below.

Perhaps if we change the `width` parameter to `min-width` instead of removing it?

image

jokd:
That works, I pushed the fix

PR #800, opened 2020-01-21
Removed collapse button

### LINKED ISSUE #772: Close layer menu only with X button
image

jokd:
Fixes #772
Removed the button and collapse function from the header since we now have a close button.

### DIFF
--- a/src/controls/legend/overlays.js
+++ b/src/controls/legend/overlays.js
@@ -73,10 +73,6 @@ const Overlays = function Overlays(options) {
   });
 
   const collapseHeader = Component({
-    onInit() {
-    },
-    onRender() {
-    },
     render() {
       const headerCls = 'flex row grow no-shrink justify-center align-center collapse-header';
       return `<div id="${this.getId()}" class="${headerCls}" style="height: 0.5rem;">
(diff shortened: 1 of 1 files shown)


Grammostola:
Hi, an empty header works well, efficient : ) If I may suggest a tiny tweak it would be somewhat reduce the area, height more precisely, of the header, perhaps with style="height: .5rem;" instead of 1.5rem:

pointfive
onepointfive

tonnyandersson:
Looks good, @jokd.

One thing, though, Is it possible to remove the empty `onInit()` and `onRender()` functions from the collapseHeader component without breaking anything? Empty functions look... unfinished.

jokd:
It is possible, but I think the components are supposed to have these functions. The functions will be called when creating the component no matter what.

tonnyandersson:
@jokd Actually, I don't think a component *must* have those functions. I think they are only needed if you want something specific done on init or on render. An 'init' event will be triggered when the component is created, no matter what. But if there's no `onInit()` function specified, nothing in particular will happen. And that's fine, if nothing in particular is supposed to happen.

jokd:
Removed!

Issue #808, opened 2020-02-05
Button has rounded children


mulfvik:

mulfvik:
This makes background maps buttons squared

PR #809, opened 2020-02-05
No roundness on button child

mulfvik:
Issue #808


jokd:
I think you need to explicitely point out the icons in the toolbar container, or else you will alter at least the background maps:
image

mulfvik:
Been looking in to this and can't find any solution that does'nt need adding some kind of configuration for the users and for thats a little bit overkill. My opinion is that for this one case of edge nudging the symbol, is fine and if the problem pops up in future scenarios we can come back to this then.

jokd:
OK to close then?

Issue #785, opened 2019-12-18
Enhancement for the printing control


jokd:
It would be nice to:
- [ ] - See the scale
- [ ] - Be able to set the scale
- [ ] - Get higher resolution
- [ ] - Have a legend
- [ ] - Optional north arrow
- [ ] - Date/time for creation
- [ ] - Preview fits to screen size

tonnyandersson:
It is possible to obtain a higher resolution just by passing a scale parameter to html2canvas. It won't make a huge difference, but it gives a noticable bump in overall image quality. As one might expect, increasing the scale has the greatest impact on regular HTML elements like texts and logos, but it will affect the map element as well. See comparison below.


![image](https://user-images.githubusercontent.com/15725608/71165083-873ef680-2250-11ea-9aa7-b4e5f0cfe0cb.png)
*Default scale (150% zoom)*


![image](https://user-images.githubusercontent.com/15725608/71165125-99b93000-2250-11ea-9307-02614d19f20e.png)
*4x scale (150% zoom)*

Please note that the difference in basemap quality is slightly (slightly) more distinct in an actual print.

This is a quick and dirty way to achieve a slightly better image quality. However, to obtain a higher resolution on the map element as well, and thus greatly improve the overall export quality, would require a bit more work than just adding a simple parameter.

jokd:
How is file size affected?

tonnyandersson:
Quite a bit. At default scale, the export above, originally an A4 PDF, weighs in at 451 kB. At 4x scale, the file size is 2447 kB.

jokd:
I can live with that if the vector features are crispy and clear, are they?
What do 2x looks like?

tonnyandersson:
Added a 2x screenshot to my first comment. 1080 kB as A4 PDF. And no, vector features are not crispy and clear since such features are a part of the map canvas, just like the basemap.

jokd:
4x gets my vote. Could it be an option for the control?
Would be nice to get this in before the release.

tonnyandersson:
It's a whole lot faster to just hardcode it. It's a quick fix, a temporary workaround. No point in putting too much time into it. File sizes and export/download times are acceptable, even for A3 exports.

But that's just, like, my opinion.

jokd:
Then I think we should set it to 4x. At least for now.

jokd:
Overlays are rendered poorly:
[...]

PR #817, opened 2020-02-19
Fixes print bug when height>width

### LINKED ISSUE #784: Print issues when width>height
When using custom format the resulting pdf is distorted if the width is greater than the height.
image

jokd:
Fixes #784
Further improvement would be to remove the portrait/landscape orientation buttons when setting custom sizes since they are confusing.

### DIFF
--- a/src/controls/print/print-component.js
+++ b/src/controls/print/print-component.js
@@ -134,15 +134,27 @@ const PrintComponent = function PrintComponent(options = {}) {
       });
     },
     async downloadPDF() {
+      let height;
+      let width;
+      let pdfOrientation;
+      if (sizes[size][1] > sizes[size][0]) {
+        height = sizes[size][0];
+        width = sizes[size][1];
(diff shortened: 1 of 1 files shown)


tonnyandersson:
@jokd I agree that the portrait/landscape buttons should be removed OR be connected to the heigth/width sliders so that p/l mode changes automatically when changing h/w, and vice versa. The latter would require more work, though, and may not be particularly useful. So I vote for hiding p/l when a custom size is set.

I think, however, that it should be done in this PR. No point in waiting. Confusing AF, right now.

jokd:
This fixes the bug that creates corrupt pdfs. Removing the toolbar can be added as an enhancement in the #785 issue. Is that ok?

tonnyandersson:
I guess I can live with that.

tonnyandersson:
Works well, PDF:s are no longer distorted.

PR #822, opened 2020-02-21
Update agstransaction.js

### LINKED ISSUE #821: agstransaction needs an update
agstransaction.js is missing a viewer parameter since just importing viewer is no longer adequate. Also, later versions of ArcGIS Server sends insert, update and delete responses as JSON objects, rather than JSON text. Because of this, the response must be handled differently depending och AGS version.

tonnyandersson:
Fixes #821.


jokd:
Unable to test this but codewise it looks good and ok to merge unless someone disagree

Issue #825, opened 2020-02-27
Outdated dependencies


daRasmussen:
Currently there are a lot of outdated dependencies and it would be nice if these were updated. I suggest following updates (brief not detailed list): 

**MAJOR UPDATE**
- acorn
- core-js
- whatwg-fetch
- babel-eslint
- compression-webpack-plugin
- copy-webpack-plugin
- eslint
- eslint-config-airbnb-base
- npm-run-all

** MINOR/PATCH UPDATE ** 
- babel/polyfill
- awesomeplete
- jquery
- proj4
- babel/core
- babel/plugin-transform-runtime
- babel/preset-env
- eslint-plugin-import
- node-sass
- uglify-js
- uglifyjs-webpack-plugin
- webpack
- webpack-cli
- webpack-dev-server

** REMOVE **
- babel/runtime-corejs2

** ADD ** 
- babel/runtime
- babel/runtime-corejs3
- regenerator-runtime

**Summary**
Biggest changes are core-js and whatwg-fetch upgrade.

PR #826, opened 2020-02-27
Fixes Outdated dependencies ISSUE #825

daRasmussen:
This PR would resolve outdated dependencies issues, it has been tested 2 times. Feel free to test and leave a comment.


daRasmussen:
This PR NEEDS a update, since these changes would also update eslint rules and airbnb which leads to an update of code to match new lint standard. I suggest that these are fixed in an different issue/pr.

daRasmussen:
This PR also fixes Issue #829

A config can be found [here](http://kartor.vasteras.se/sam/vas.json)

jokd:
Everything seems to work well

tonnyandersson:
LGTM2, @daRasmussen . But next time, please try to split the stuff into several PR's. I think it is bad practice to fix so many unrelated issues in the same PR.

daRasmussen:
@tonnyandersson Yes, will do in the future!

PR #838, opened 2020-03-04
Added WMTS style option

### LINKED ISSUE #619: WMTS style parameter should be made into a config option
WMTS source style parameter is hardcoded as 'default' by default. This means the user must provide a style named 'default' for the WMTS service, or hardcode a different parameter value. This parameter should instead be made into a configurable parameter, with an empty string as default.

https://github.com/origo-map/origo/blob/5ebf8c9e5cae1323cc472dfa3a03164a980c9b65/src/layer/wmts.js#L7-L24

tonnyandersson:
Fixes #619.

Can be tested with [test map config](https://www.malardalskartan.se/origo/v2.0.1/test.json), or any other config with WMTS layers.


jokd:
Dont forget the documentation ;)

tonnyandersson:
Oh, I won't. Promise.

PR #843, opened 2020-03-04
Hide mapmenu on mapclick

### LINKED ISSUE #179: Hide mapmenu on mobile when clicking the map
When mapmenu is expanded on a mobile device and you want to hide it, you should be able to click the map to do it. Both Google maps and Eniro has that implementation. What do you think of this, is it something that we want to implement in Origo?

tonnyandersson:
Fixes #179.

Adds the option `"autoHide"` to the mapmenu control. It has three different settings: `"always"`. `"mobile"` and `"never"` (default).

`"always"` always closes the mapmenu when clicking the map.
`"mobile"` hides the mapmenu when clicking the map if the map size is `m`, `s` or `xs` (see breakpoints in origo.js)
`"never"` never closes the mapmenu when clicking the map.

This function could easily be added to the legend as well.

### DIFF
--- a/src/controls/legend.js
+++ b/src/controls/legend.js
@@ -6,6 +6,7 @@ const Legend = function Legend(options = {}) {
   const {
     cls: clsSettings = '',
     style: styleSettings = {},
+    autoHide = 'never',
     expanded = true,
     contentCls,
     contentStyle,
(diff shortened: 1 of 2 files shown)


tonnyandersson:
This awesome feature is now available for the legend also!

johnnyblasta:
That works too

PR #844, opened 2020-03-04
Makes it possible to use static urls as attributes

### LINKED ISSUE #660: Static url as attribute
As now you can only use an attribute as url value. It would be nice to be able to use a static url as well.

jokd:
Fixes #660 
If url-value is not an attribute it checks if its a valid url
```
      "attributes": [
          {
            "title": "Some title",
            "url": "//www.static.url/that/is?true=yes",
            "target": "modal"
          }
      ],
```

### DIFF
--- a/src/getattributes.js
+++ b/src/getattributes.js
@@ -1,6 +1,6 @@
 import featureinfotemplates from './featureinfotemplates';
-import replacer from '../src/utils/replacer';
-import isUrl from '../src/utils/isurl';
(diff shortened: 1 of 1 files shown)


tonnyandersson:
Gettin' some linter errors because of the new linting regime that came with #826. Did you update your Atom linter add-ons?

Other than that... LGTM.

jokd:
Missed that update^^

Issue #856, opened 2020-03-10
Text overflow in Legend


daRasmussen:
If the title of the layer in the legend is too long, origo handles overflow like: 

![image](https://user-images.githubusercontent.com/33061427/76298928-813e3c00-62ba-11ea-9bd7-d8e7772cba44.png)

Wish is pretty nice, but would like the default behavior to be: 

![image](https://user-images.githubusercontent.com/33061427/76299191-f0b42b80-62ba-11ea-9105-61a92143bcca.png)

PR #857, opened 2020-03-10
Overflow fix

daRasmussen:
fixes issue #856


tonnyandersson:
Tested with a long title and got some cool legend flutters when moving the mouse around. The length of the title does not seem to be the problem, but rather how it's made up. The flutter only occurs when there are lines that are like a pixel or so from being broken into two lines.

Oh, and the flutter stops if you remove the position control.

flutter

tonnyandersson:
The fluttering should probably be treated as a separate issue, since it seems to be part of a larger problem, involving the position control and whatnot.

Issue #516, opened 2018-11-19
Feature information


asemoller:
A new way of presenting feature information from multiple selected objects within the same layer as well as several active layers.

A mockup of gui is forthcoming.

Issue #744, opened 2019-11-07
Selection manager


asemoller:
Selection manager is a module that recieves all selected objects from different tools in Origo and sends those objects on to the new bar/information window. All tools that in any way select one or more objects, such as default selection, multiselection and search, sends the objects on to the selection manager that in turn handles all logic concerning what to do with the selection.
It also handles properties from the jsonfile (config) such as name, geometry, attributes etc.

As of today, the only "receiver" ~~thing~~ that is functional with the selection manager is the new bar/information window (that is to be included in the core). Selection manager works side by side with current solution, which is not optimal. The selection manager should eventually replace the current solution.

Two images that describes present vs future.

EDIT: Updated image (present):

![sm_present2](https://user-images.githubusercontent.com/17123002/68571826-345d6c80-0464-11ea-851a-19af6eb53e76.PNG)

![sm_future](https://user-images.githubusercontent.com/17123002/68388585-4dfb6d00-0161-11ea-9100-a1cd1ab81861.PNG)

asemoller:
Please note that the image "Present" have been updated.

asemoller:
Can we close this now?

PR #861, opened 2020-03-17
PR for Infowindow and selection manager 

imantahriri:
Infowindow is the given name for an alternative tool to show the information about one or more selected feature(s) on the map. 

one small change in the index.json file is required as `overlay: true / false` should be changed to `infowindow: infowindow / overlay / sidebar`.

Choosing infowindow lets user to see all the selected features and highlight or deselect each one of them. Besides, features will be shown in different groups, each group representing a layer or a group-layer. 

All selected features can be sent to an external service. Features of layers marked with "simpleExport" in the config will be sent to a simple service in origo-server (comes in a seperate PR) and features of layers marked with "layerSpecificExport" can be sent to any given external service to create an ecxel or pdf list of features or any other possible analysis .

A "Selection manager" is associated with this tool that has the responsibility of storing selected features, adding features to them, removing from them and other managerial logic. Infowindow is the representational part of.

The main purpose of the Selection manager is to be central brain of the application regarding selecting/deselecting features and all other tools like multi-selector and search etc should simply deliver an item/several items and manager should do the rest.

Note that selectionmanager module IS but infowindow module IS NOT converted to Origo 2 format.


asemoller:
This pr fixes issue #516 and #744.

A config with all settings are in progress.

asemoller:
Se this [config](https://karta.sigtuna.se/arbmat/pr_861/index.json).

What does "All checks have failed" mean @daRasmussen?
What can we do to fix it?

daRasmussen:
@asemoller checking

daRasmussen:
@asemoller [Hope this helps](https://github.com/origo-map/origo/pull/863)

jokd:
Tested and works as expected besides the behaviour of clicking the row/arrow in the infowindow. I suggest clicking a row should always select the feature (and perhaps expand the rows info/collapse the other rows), and the arrow should just toggle the info (expanded/collapsed). See attached gif.
I haven't tested the export functions.
Besides that I get some linter errors in Atom that should be taken care of.
gif

daRasmussen:
@asemoller @imantahriri All checks have passed

tonnyandersson:
Clicking around in the infowindow seems to work fine now but I'll let @jokd have the last word on that since he brought it up.

Exporting to Excel works like a charm, except for one thing. There seems to be some inconsistencies regarding how the geometry attribute is handled at export depending on layer type. Or rather, depending on *map server* type. The function is hardcoded to delete the `geom` attribute, but "geom" is not a standard geometry attribute name. ESRI services, for instance, uses the name "geometry".

I suggest using the `feature.getGeometryName()` function to obtain the actual geometry name before removing the property, instead of using the hardcoded `obj.geom`.

imantahriri:
> Clicking around in the infowindow seems to work fine now but I'll let @jokd have the last word on that since he brought it up.
> 
> Exporting to Excel works like a charm, except for one thing. There seems to be some inconsistencies regarding how the geometry attribute is handled at export depending on layer type. Or rather, depending on _map server_ type. The function is hardcoded to delete the `geom` attribute, but "geom" is not a standard geometry attribute name. ESRI services, for instance, uses the name "geometry".
> 
> I suggest using the `feature.getGeometryName()` function to obtain the actual geometry name before removing the property, instead of using the hardcoded `obj.geom`.

That was a really good point @tonnyandersson, fixed now.

jokd:
@imantahriri There are still linter issues in eg featureinfo.js and getfeatureinfo.js

jokd:
LGTM

PR #865, opened 2020-03-20
Require list items contents to add them to the feature info popup DOM

### LINKED ISSUE #864: List elements of empty attributes are still added to the feature info popup, together with any custom css classes
In Origo the normal behaviour of the feature info popup window is to not show empty attributes. 

However, they are still added as empty HTML `<li>` elements (to the DOM). A custom CSS class specified with the `cls` attribute parameter is also still applied to the `<li>` element, which causes any visual CSS styles to still show up in the feature info popup window, even when the attribute itself is empty.

**To Reproduce**
1. Specify custom attributes on a layer, e.g.
```json
"layers": [
  {
    "name": "layername",
    "etc": "blabla",
    "attributes": [
      {
        "name": "myattribute",
        "title": "My attribute: ",
        "cls": "o-custom-style"
      }
    ]
  }
]
```
2. Add some visible CSS styling to the custom class `o-custom-style` in `style.css`, e.g.
```javascript
.o-custom-style { 
  border-style: solid; }
```
3. Click on a feature where the `myattribute` field is empty.
4. See the visible CSS style applied to the empty `<li>` element.

**Expected behavior**
Empty attributes should not show up, and neither should their custom CSS styles.

**Desktop:**
 - OS: Windows 7
 - Browser: Chrome
 - Origo version: 2.1.0

MattiasSp:
Fixes #864.

Another condition was added for featureinfo popup list elements to be added to the DOM, requiring that the list item has contents (`.innerHTML.length > 0`).

### DIFF
--- a/src/getattributes.js
+++ b/src/getattributes.js
@@ -143,7 +143,7 @@ function getAttributes(feature, layer, map) {
           } else {
             val = customAttribute(feature, attribute, attributes, map);
           }
-          if (val instanceof HTMLLIElement) {
+          if (val instanceof HTMLLIElement && val.innerHTML.length > 0) {
             ulList.appendChild(val);
           }
         }


MattiasSp:
I suspect that the Travis issue is related to this temporary Mac build problem: https://www.traviscistatus.com/incidents/7mbxh26jk3sl

See also this community discussion: https://travis-ci.community/t/macos-build-an-error-occurred-while-generating-the-build-script/7684/7

Just restarting the Travis build would hopefully work this time @daRasmussen :)

daRasmussen:
@MattiasSp thanks for information, I restarted the check hopefully it will pass.

PR #869, opened 2020-03-24
New `groupSuggestions` option, and group titles same as in feature info popups

### LINKED ISSUE #868: Possibility to group Search control suggestions without defining map layers
When configuring the Search control, defining the `layerNameAttribute`  option causes the awesomplete search suggestions to be grouped by layer, but [group titles are fetched from map layer titles](https://github.com/origo-map/origo/blob/7e0d00ae42303514e88615890f16bf7d13dc0d32/src/controls/search.js#L240) of defined map layers only, making it impossible to group suggestions when searching purely against services such as origo-server.

It would simplify the configuration and use of the Search control if the handling of titles in both suggestion groups and in feature info popups were made more explicit and consistent.

**Proposed solution**
The grouping of search suggestions should stop relying on the `layerNameAttribute` option and could instead be turned on with its own (boolean) Search control option: `groupSuggestions`.

The titles used for suggestion groups should by default be the same as the titles in the feature info popup, and should use the `title`, `titleAttribute` or map layer titles similarly. If the need arises, it would be easy to later add e.g. `suggestionTitle` and `suggestionTitleAttribute` to make it possible to use different titles for the suggestion groups and the feature info popups.

**Alternatives considered**
It would be possible to keep using the `layerNameAttribute` for turning the suggestion grouping on, and if using it with `titleAttribute` [(Option 3)](https://origo-map.github.io/origo-documentation/latest/#search-control), the awesomplete handler could try to fetch the title using the `titleAttribute` if there is no matching map layer defined. This would require fewer code changes and would avoid adding new Search control options, which are already quite many.

But, in my opinion, using `layerNameAttribute` to activate suggestion grouping is rather unintuitive (and more importantly [undocumented](https://github.com/origo-map/origo-documentation/issues/88)) and being more explicit about what options/effects are active in a control is better.

MattiasSp:
Fixes #868.

A `groupSuggestions` option is added for the Search control which replaces `layerNameAttribute` in controlling whether or not to group suggestions.

The titles for suggestion groups are now fetched the same way as the titles for feature info popups, providing more consistency and flexibility, such as the possibility to add suggestion group titles for single layer searches or searches without defined map layers.

### DIFF
--- a/src/controls/search.js
+++ b/src/controls/search.js
@@ -31,6 +31,7 @@ const Search = function Search(options = {}) {
     title,
     titleAttribute,
     contentAttribute,
+    groupSuggestions,
     includeSearchableLayers,
     searchableDefault,
     maxZoomLevel,
@@ -233,13 +234,23 @@ const Search = function Search(options = {}) {
     const group = {};
     const ids = Object.keys(data);
     ids.forEach((id) => {
-      const item = data[id];
-      const type = item[layerNameAttribute];
-      if (type in group === false) {
-        group[type] = [];
-        item.header = viewer.getLayer(type).get('title');
+      const item = data[id];      
+      const typeTitle =
+        (layerNameAttribute && idAttribute) ? viewer.getLayer(item[layerNameAttribute]).get('title') :
+        (geometryAttribute && layerName) ? viewer.getLayer(item[layerName]).get('title') :
+        (titleAttribute && contentAttribute && geometryAttribute) ? item[titleAttribute] :
+        (geometryAttribute && title) ? title :
+        (easting && northing && title) ? title :
+        undefined;
+      if (typeTitle && typeTitle in group === false) {
+        group[typeTitle] = [];
+        item.header = typeTitle;
       }
-      group[type].push(item);
+      if (typeTitle) {
+        group[typeTitle].push(item);
+      } else if (id == 0) {
+        console.log('Search options are missing');
+      }
     });
     return group;
   }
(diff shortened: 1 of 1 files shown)


src/controls/search.js
```diff
@@ -233,13 +234,23 @@ const Search = function Search(options = {}) {
...
     const ids = Object.keys(data);
     ids.forEach((id) => {
-      const item = data[id];
-      const type = item[layerNameAttribute];
-      if (type in group === false) {
-        group[type] = [];
-        item.header = viewer.getLayer(type).get('title');
+      const item = data[id];      
+      const typeTitle =
+        (layerNameAttribute && idAttribute) ? viewer.getLayer(item[layerNameAttribute]).get('title') :
```
daRasmussen: Could this if statement be refactored?

daRasmussen:
@MattiasSp This is a nice addition although it would be nice if you could supply a config for testing.

src/controls/search.js
MattiasSp: Thanks for the review! I have pushed a commit refactoring the ternary assignment. Please take another look :)

MattiasSp:
> @MattiasSp This is a nice addition although it would be nice if you could supply a config for testing.

A config is in the works.

MattiasSp:
> @MattiasSp This is a nice addition although it would be nice if you could supply a config for testing.
> 
> A config is in the works.

Now, after a fix, here's a [config](https://karta.haninge.se/devmaps/PR869/index.json).

MattiasSp:
Something to note:

This PR will require users with the search control options `layerNameAttribute` and `idAttribute` in their configs to also add the new `groupSuggestions: true` if they wish keep the same behaviour.

MattiasSp:
Added a PR for accompanying docs: [origo-documentation PR #97](https://github.com/origo-map/origo-documentation/pull/97)

PR #889, opened 2020-04-02
Error message read from config

imantahriri:
Fixing one little thing that was missing in PR for Infowindow. Plus fixing Linter problem in a nother file related to that PR.

Fixes issue #890


tonnyandersson:
LGTM. Perhaps the default error messages should be in english, but since a lot of other default/hardcoded messages in Origo are in swedish, I'll let this one slide.

PR #905, opened 2020-04-16
Value zero attribute fix

### LINKED ISSUE #547: Value 0 is interpreted as undefined

johnnyblasta:
fixes #547 

Example of 0 value in attribute:
image

Example of 0 value in attribute for popup:
image


tonnyandersson:
Seems to work as intended.

Issue #902, opened 2020-04-14
Disabling linting rules for next line only


mulfvik:
**Is your feature request related to a problem? Please describe.**
Linting problems resolved by "disable next line" could affect code in a negative way. 

**Describe the solution you'd like**
It should instead be configured to ignore a rule for the whole project otherwise the code should be refactored. 
(To reasonable extent)

PR #906, opened 2020-04-17
Linting rules first stop

mulfvik:
Addresses issue #902


src/utils/generateuuid.js
```diff
@@ -8,8 +9,10 @@ const generateUUID = function generateUUID() {
     d += performance.now(); // use high-precision timer if available
   }
   const uuid = 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, (c) => {
+    // eslint-disable-next-line no-bitwise
     const r = (d + Math.random() * 16) % 16 | 0;
     d = Math.floor(d / 16);
+    // eslint-disable-next-line no-bitwise
     return (c === 'x' ? r : (r & 0x3 | 0x8)).toString(16);
```
daRasmussen, line 16: There is a better solution to this rather than no-bitwise.
(c === 'x' ? r : ((r && 0x3) || 0x8)).toString(16);

src/utils/validate.js
```diff
@@ -26,7 +26,7 @@ validate.email = (email) => {
 
 validate.url = (url) => {
   if (url) {
-    const regex = /^(?:www|http(s)?:\/\/)+[\w\-\._~:/?#[\]@!\$&'\(\)\*\+,;=.]+$/i;
+    const regex = /^(?:(?:https?|ftp):\/\/)?(?:(?!(?:10|127)(?:\.\d{1,3}){3})(?!(?:169\.254|192\.168)(?:\.\d{1,3}){2})(?!172\.(?:1[6-9]|2\d|3[0-1])(?:\.\d{1,3}){2})(?:[1-9]\d?|1\d\d|2[01]\d|22[0-3])(?:\.(?:1?\d{1,2}|2[0-4]\d|25[0-5])){2}(?:\.(?:[1-9]\d?|1\d\d|2[0-4]\d|25[0-4]))|(?:(?:[a-z\u00a1-\uffff0-9]-*)*[a-z\u00a1-\uffff0-9]+)(?:\.(?:[a-z\u00a1-\uffff0-9]-*)*[a-z\u00a1-\uffff0-9]+)*(?:\.(?:[a-z\u00a1-\uffff]{2,})))(?::\d{2,5})?(?:\/\S*)?$/;
```
daRasmussen, line 29: This is a glorious regex but ftp, and not sftp? 
And I would refactor this part a bit, for readability.

daRasmussen:
One more thing in a future PR it would be nice if eslintrc was moved to package.json

src/utils/validate.js
mulfvik: Hahah, I can not take any credit for that.

src/utils/generateuuid.js
mulfvik: This is correct according to [spec](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_Operators)

tonnyandersson:
@mulfvik Removed *self-import* from permalinkparser.js.

mulfvik:
@tonnyandersson Ok for merge?

PR #917, opened 2020-04-27
Share map with clustered features

### LINKED ISSUE #910: Share map link to marked object
**Describe the bug**
When an object is marked the share map link does not pan/zoom to that object.

**To Reproduce**
Mark an object, press share map control and open the url i new window.

**Expected behavior**
The map should pan/zoom to the object that was marked before getting the link.

**Additional context**
Dont find any documentation for this feature.

mulfvik:
Fixes #910

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -411,22 +411,17 @@ const Viewer = function Viewer(targetOption, options = {}) {
         const featureId = urlParams.feature;
         const layerName = featureId.split('.')[0];
         const layer = getLayer(layerName);
+        const vectorSource = layer.getSource();
+        const clusterSource = layer.getSource().source;
+
         if (layer) {
           layer.once('postrender', () => {
-            let feature;
             const type = layer.get('type');
-            feature = layer.getSource().getFeatureById(featureId);
-            if (type === 'WFS') {
+            let feature;
+            if (vectorSource.constructor.name === 'Cluster') {
+              feature = clusterSource.getFeatureById(featureId);
+            } else if (vectorSource.constructor.name === 'VectorSource' && type !== 'TOPOJSON') {
               feature = layer.getSource().getFeatureById(featureId);
-            } else {
-              const id = featureId.split('.')[1];
-              let origin = layer.getSource();
-              feature = origin.getFeatureById(id);
-              // feature has no id it is not found it maybe a cluster, therefore try again.
-              if (feature === null && type !== 'TOPOJSON') {
-                origin = origin.getSource();
-                feature = origin.getFeatureById(id);
-              }
             }
             if (feature) {
               const obj = {};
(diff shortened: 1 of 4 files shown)


jokd:
Tested and the map still doesn't pan to the feature. Tested on Edge Chromium with the origo-cities geojson as cluster layer.

mulfvik:
> Tested and the map still doesn't pan to the feature. Tested on Edge Chromium with the origo-cities geojson as cluster layer.

Only tested with WFS so my bad, looking in to it.

mulfvik:
Found one more bug, when infowindow is infowindow the zoom to object wont work at all. No id in url is present

src/viewer.js
```diff
@@ -411,17 +411,21 @@ const Viewer = function Viewer(targetOption, options = {}) {
...
-            if (type === 'WFS') {
+            if (clusterSource && type === 'WFS') {
+              feature = clusterSource.getFeatureById(featureId);
+            } else if (type === 'WFS') {
               feature = layer.getSource().getFeatureById(featureId);
             } else {
               const id = featureId.split('.')[1];
               let origin = layer.getSource();
               feature = origin.getFeatureById(id);
+              feature = layer.getSource().getFeatureById(featureId);
```
jokd, line 428: I think this row just messes things up.

mulfvik: Refactored this and came up with that topojson has problems with this feature
             `if (type === 'WFS' && clusterSource) {
              feature = clusterSource.getFeatureById(featureId);
              console.log(feature)

            } else if (type === 'WFS') {
              feature = layer.getSource().getFeatureById(featureId);
              console.log(feature)

            } else if (type === 'GEOJSON' && clusterSource) {
              feature = clusterSource.getFeatureById(id);
              console.log(feature)

            } else if (type === 'GEOJSON') {
              feature = layer.getSource().getFeatureById(id);
              console.log(feature)

            } else if (type === 'TOPOJSON' && clusterSource) {
              feature = clusterSource.getFeatureById(id);
              console.log(feature)

            } else if (type === 'TOPOJSON') {
              feature = layer.getSource().getFeatureById(id);
              console.log(feature)
            }`

jokd: Is that a working solution?

mulfvik: This is an example where WFS and GEOJSON is returning a feature but TOPOJSON returning null. Not sure why, the getFeatureById is pretty straight forward..

jokd: could it be your layer source?

mulfvik: Yes, Im confused where in the prototype the feature is found..

jokd: I tried the socken.json topojson file and that file seems to lack IDs. The id in the permalink is perhaps the OL id.

mulfvik: If i share Övik (id 16) in geojson. Permalink is referencing origo-cities.7. I **thought** getFeatureById was straight forward, I was wrong.
Testing on fme converted origo-cities.geojson -> origo-cities.json and in topojson i get id 94

mulfvik: So, topojson object id should not be nested inside properties. There is nothing wrong with the code as far as I can se.

PR #918, opened 2020-04-29
Adds created datetime label for print control

### LINKED ISSUE #916: Print - Date/time for creation
It would be nice to see the date/time for creation of printed map.

The date/time for creation could be written out in the lower right corner. It should be selectable in the control settings if it is shown or not. Default value should be assigned in the config.

johnnyblasta:
Closes #916 

There is a fault I'm having trouble finding the cause for. First open the print control and deselect the "Använd marginaler". Then close the print control and open it again. The result is a white border at the bottom that shouldn't be there, see image.
image

### DIFF
--- a/scss/_print.scss
+++ b/scss/_print.scss
@@ -4,7 +4,7 @@ $print-control-border-radius: $radius-small - 0.0625rem;
 $print-control-font-size: $font-size-smallest;
 $print-control-padding: .1875rem;
 $print-control-z-index: $z-middle;
-$print-margin: 10mm 15mm;
+$print-margin: 10mm 15mm 5mm;
 
 $print-scale-line-color: $grey-darker;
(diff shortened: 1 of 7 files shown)


johnnyblasta:
The problem white border at the bottom is now fixed.

tonnyandersson:
Perhaps this could have been made in a slightly cleaner way, without all the adding and removing of classes in print-component.js. Kinda like the margin, title and description controls. Would that require a lot of work?

johnnyblasta:
Somewhat cleaner implementation is committed.

tonnyandersson:
@johnnyblasta Oh, I forgot. The "Created..." text should be made into an option and should not be hardcoded.

johnnyblasta:
Even cleaner implementation, but without same height on top and bottom margin when created is inserted. Removed an accidentally inserted multiple instance of description.

PR #927, opened 2020-05-12
Tries to check whether mapStateID is employed for embedded map when f…

### LINKED ISSUE #912: Fullscreen never minds a sharemap: storeState: saveStateToServer
**Describe the bug**
The Fullscreen function in an embedded map will open a new browser window/tab and calculate the map state with the hash method instead of using the mapStateId the map was embedded with

**To Reproduce**
-Have Origo server ready to save map states. 
-Set index.json for the Origo client as such:     ` {
      "name": "sharemap",
      "options" : {
        "serviceEndpoint": "https://we_have_domain/mapstate",
        "storeMethod": "saveStateToServer"
      }
    }, `
-Save a map state by clicking Share in the Origo client
-Embed shared (mapState) link someplace, I've only tried as an iframe. The Origo client then becomes aware that it has been embedded and shows the Fullscreen button.
-Click/push the fullscreen button
-Saved map will open in new browser window/tab as a hash save URL

**Expected behavior**
-Saved map opens in new browser window as it was embedded (Ori-go map/?mapStateId=6aab6501-02e2-4937-a6ed-06c8cdb90600)

**Additional context**
-This is currently unlikely to cause functional issues but it seems a bit inconsistent
-permalink.js currently looks ready to receive a mapStateId but fullscreen never sends it
-There may be special considerations in other methods of embedding than iFrame that I am unaware of. If so, please mention it below

Grammostola:
…ullscreen is activated and if so uses that (url) when opening a new tab/window.

Fixes #912 

This is a rather simple fix that seems to work for (our) iframes at the very least, i.e if the embedded map url contains a ?mapstate= then that url is used when clicking the fullscreen button, else the hash is calculated and used. The only method to ascertain a current mapStateID that I managed to find was getQueryVariable(  ) in loadresources. It relies on window.location too and I was hesitant to add a reference to loadresources in fullscreen. 

Feel free to test/comment.

### DIFF
--- a/src/controls/fullscreen.js
+++ b/src/controls/fullscreen.js
@@ -13,8 +13,8 @@ const Fullscreen = function Fullscreen(options = {}) {
   const goFullScreen = function goFullScreen() {
     let url = '';
     if (window.location.href.indexOf('?mapStateId=') > 0) {
-      url =  window.location.href;      
-    } else {url = permalink.getPermalink(viewer); }
+      url = window.location.href;
+    } else { url = permalink.getPermalink(viewer); }
     window.open(url);
   };
 
--- a/src/controls/fullscreen.js
+++ b/src/controls/fullscreen.js
@@ -11,7 +11,10 @@ const Fullscreen = function Fullscreen(options = {}) {
   let fullscreenButton;
 
   const goFullScreen = function goFullScreen() {
-    const url = permalink.getPermalink(viewer);
+    let url = '';
+    if (window.location.href.indexOf('?mapStateId=') > 0) {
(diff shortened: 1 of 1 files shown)


tonnyandersson:
Seems to work just fine. Just a few minor specks of lint to be removed.

tonnyandersson:
LGTM. No really a fan of using hardcoded values (e.g. `'?mapStateId='`) defined in different modules as part of a condition, mainly because of the risk of values being changed without notice. But since the risk is low in this case, and because I don't have a better solution at hand, I'll approve this. It's clean and it works.

PR #929, opened 2020-05-15
Added support for styling with mbstyle

### LINKED ISSUE #805: Support for Mapbox style specification
I am adding support for Mapbox style specification by integrating [ol-mapbox-style](https://github.com/openlayers/ol-mapbox-style),  primarily for styling vector tile layers but could also be used to add layers and configure the map.
Should this be part of Origo map or a plugin? Or perhaps introduce the concept of native plugins/extensions that can be turned on/off in build time similar to controls.

The library will add ~250kB to the build size.

jokd:
Fixes #805 - Styling with mapbox style file
Added the concept of extensions similar to controls but not part of the UI. To make the extension part of the build make sure it is in extension.js and to use it in the map activate it in the config:
`  "extensions": [
      {
        "name": "Olms"
      }
  ],`
Config to use Lantmäteriets open vt service is attached.

[lmvt.json.txt](https://github.com/origo-map/origo/files/4633524/lmvt.json.txt)


jokd:
Activating the extension adds ~250kb to the build.

jokd:
...or do you think this should be a plugin instead?

johnnyblasta:
Works for me now. I had missed uncommenting in extensions.js
Would be nice to test with other style too, to be meticulous.

PR #935, opened 2020-05-27
Introduces a positional memory property for the legend's scrollbar

### LINKED ISSUE #691: Returning from "layer info" to the same position in the legend
Ability for the legend to retain scroll position when returning from "layer info" (clicking on the layer name). As it is now, when returning after choosing "layer info" it leaves you starting at the top of the legend and not where you were.

Grammostola:
First suggestion, rough draft .. 
fixes #691 , kind of : ) 

This is how it currently is:
pre_position

Compared to what this patch does:
post_position


jokd:
Works well I think

jokd:
OK to merge?

Issue #939, opened 2020-05-27
Search not working in IE11


johnnyblasta:
**Describe the bug**
Search not working in IE11

**To Reproduce**
1. Go to https://www.malardalskartan.se/origo/v2.2.0/test.html
2. Search for 'karl'
3. No result

**Expected behavior**
Should have returned matches.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser IE
 - Version 11

**Additional context**
Errors in console

mulfvik:
[https://stackoverflow.com/questions/30571488/windows-resize-trigger-event-is-not-working-in-ie11](https://stackoverflow.com/questions/30571488/windows-resize-trigger-event-is-not-working-in-ie11)

jokd:
Fixed with #940

PR #940, opened 2020-05-27
IE not like includes

mulfvik:
This fixes the issue #939 but there are still errors. This doesnt add too much code and awaits no Origo support for IE


tonnyandersson:
Isn't there a polyfill for `includes`?

https://github.com/origo-map/origo/blob/master/src/utils/polyfill.js

mulfvik:
When importing babel/polyfill in polyfill.js its fine. I dont know if its supposed to be imported this way though?

jokd:
We need polyfill for custom events as well

jokd:
Nevermind, polyfill was overkill. I pushed a minor fix.

Issue #941, opened 2020-05-28
"infowindow" - move "ExportedFileName" to "ExportUrls"


asemoller:
**Is your feature request related to a problem? Please describe.**
Currently the filename for layer specific export in infowindow is set per layer which makes it impossible to set multiple exports with different file extensions within the same layer.

**Describe the solution you'd like**
Move "ExportedFileName" to "ExportUrls" for LayerSpecificExport.

jokd:
Closed with #942

PR #942, opened 2020-05-28
Moved exportFileName and some small fixes

asemoller:
Made it possible to have multiple exports for a layer with different file extensions.
Fixes issue #941 .

Also made some changes for error message, filename and button text.

### DIFF
--- a/src/infowindow.js
+++ b/src/infowindow.js
@@ -187,8 +187,8 @@ function createToaster(status, message) {
   const toaster = document.createElement('div');
   toaster.style.fontSize = '12px';
   if (!message) {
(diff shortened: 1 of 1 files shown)


jokd:
Isn't a default value for filename a good idea? What will the file be named if the option is omitted?

asemoller:
I agree that default values are to prefer.

In this case we can send data to any service that returns something. What it returns and in what format can differ. If the service returns anything as a file, then the file must be defined by its file extension. This will then negate the necessity of default values. To ensure that the administrator thinks things through, specifically the file extension, I think it is better to have it as a required field instead. A default value would just be wrong as we would have to presume which file extension is implied.

jokd:
I see but could a default name, like the date and time, be set and then the file type (as in extension) come from the service?

asemoller:
That would probably work and be a much better solution. However I just try to clean up the existing code without the intention to present a final solution.

PR #945, opened 2020-05-29
OL 6.3.1

jokd:
Just a patch


jokd:
Unnecessary

jokd:
The caret ^ in the version in package.json makes it unnecessary. It will grab 6.3.1 anyway if you make a fresh install.

daRasmussen:
@jokd great stuff!

PR #947, opened 2020-06-01
Fit to feature in share map

### LINKED ISSUE #946: Share map should fit to feature
When using the share map function the map is zoomed to a certain zoom level. When using polygons this often makes the map zoom in further than the extent of the feature.

The map is set to fit the shared feature instead of zooming to a specific zoom level.

jokd:
Fixes #946 
Can be tested with test config and the wfs layer

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -438,9 +438,10 @@ const Viewer = function Viewer(targetOption, options = {}) {
               const centerGeometry = getcenter(feature.getGeometry());
               const infowindowType = featureinfoOptions.showOverlay === false ? 'sidebar' : 'overlay';
               featureinfo.render([obj], infowindowType, centerGeometry);
-              map.getView().animate({
-                center: getcenter(feature.getGeometry()),
-                zoom: getResolutions().length - 2
+              map.getView().fit(feature.getGeometry(), {
+                maxZoom: getResolutions().length - 2,
+                padding: [15, 15, 40, 15],
+                duration: 1000
               });
             }
           });


src/viewer.js
```diff
@@ -438,9 +438,10 @@ const Viewer = function Viewer(targetOption, options = {}) {
               const centerGeometry = getcenter(feature.getGeometry());
               const infowindowType = featureinfoOptions.showOverlay === false ? 'sidebar' : 'overlay';
               featureinfo.render([obj], infowindowType, centerGeometry);
-              map.getView().animate({
-                center: getcenter(feature.getGeometry()),
-                zoom: getResolutions().length - 2
+              map.getView().fit(feature.getGeometry(), {
+                maxZoom: getResolutions().length - 2,
+                padding: [15, 15, 40, 15],
```
daRasmussen, line 443: I wonder if these values covers all possible use cases? 
If possble it would be useful this was expressed in code like: 
`const padding = [top, right, bottom, left];` or `const padding = Array(4).fill(myPadding);`

daRasmussen: Another cool way would be to base padding on geometry extent to get the optimal values for top, right, bottom and left. I could try to make this enhancement if someone thinks its a good idea.

jokd: Padding are pixels and just to get some space outside of the feature with a bit more in the bottom due to the footer. But sure, on mobile screen withou footer you can have another value but this was an easy way of doing it.
Or in the future if we use multiple layouts a global padding array would be super nice to use but perhaps this kick ass feature can be put on hold til then?

jokd: Anyway, can we merge this and enhance it later on?

daRasmussen: @jokd for sure

PR #958, opened 2020-06-18
Fixed position on coordinates

### LINKED ISSUE #914: Position controll performance issue
**Describe the bug**
Position controll in latest origo-master update coordinates when panning map, browser need to render more resulting in lag/stutter. The issue is also somehow related to the legend controll and how many layers/group/subgroup it contains so if you have a lot of them it the map it will adversely affects performance.

**To Reproduce**
1. Add more (>10 of each) layers/groups and subgroups to index.json example.
2. Click and pan/drag the map (see that coordinate update in footer when pan).
3. Test panning the map using legend controll on or off to see the difference.
4. Recording performance in chrome debug shows renderingtime for task ”layout” that is higher even if layers are turned off.

**Expected behavior**
Working in same way as in origo version 1 when coordinates from mousepointer locks and not update when you click/hold and drag/pan the map.
Alternative (or both solutions) have a option for position controll to only use positonmarker (sight icon) function and not the ”always on” mousepointer update coordinates.
Of course best solution should be to solve the core problem, why legend items affects performance in Origo version 2.x.x

 - OS: WIN10 build 18362
 - Browser chrome
 - Version 81

### LINKED ISSUE #955: Background maps abstract flutter
Background maps abstract flutter

`"attribution": "&copy Lantmäteriet Geodatasamverkan | Strängnäs kommun",
      "abstract": "<p>Information om när detta fotot är taget finner du <a href='https://geolex.etjanster.lantmateriet.se/?e=614815&n=6580926&z=7&profile=default_geolex_orto_latest_year_noauth' target='_blank'>här</a> på Lantmäteriets tjänst GeoLex.</p>"`

captured

### LINKED ISSUE #956: Background maps abstract does not handle a tag properly
Can not follow link with click in background map abstract. Maybe it is due to z-index?

mulfvik:
Fixes #955
Fixes #956
Fixes #914

### DIFF
--- a/scss/_o-scale.scss
+++ b/scss/_o-scale.scss
@@ -1,3 +1,8 @@
 .o-scale {
   padding-right: 5px;
+  padding-right: 10px;
+  font-family: Arial, "Helvetica Neue", sans-serif;
+  font-size: .625rem;
+  line-height: $line-height-larger;
+  ;
 }
--- a/scss/_o-scale.scss
+++ b/scss/_o-scale.scss
@@ -4,5 +4,4 @@
   font-family: Arial, "Helvetica Neue", sans-serif;
   font-size: .625rem;
   line-height: $line-height-larger;
-  ;
 }
--- a/scss/_position.scss
+++ b/scss/_position.scss
@@ -72,7 +72,6 @@
 }
 .o-position-coords,
 .o-position-find {
-  position: fixed;
   background-color: transparent;
   border: none;
   color: #fff;
--- a/scss/_position.scss
+++ b/scss/_position.scss
@@ -78,10 +78,11 @@
   color: #fff;
   font-family: Arial, "Helvetica Neue", sans-serif;
   font-size: 0.625rem;
+  line-height: $line-height-larger;
   margin: 0;
 }
 input.o-position-find {
   padding: 0;
   width:95px;
-  line-height: $line-height-large;
+  line-height: $line-height-larger;
 }
--- a/scss/_position.scss
+++ b/scss/_position.scss
@@ -77,11 +77,9 @@
   color: #fff;
   font-family: Arial, "Helvetica Neue", sans-serif;
   font-size: 0.625rem;
-  line-height: $line-height-larger;
   margin: 0;
 }
 input.o-position-find {
   padding: 0;
   width:95px;
-  line-height: $line-height-larger;
 }
--- a/scss/_position.scss
+++ b/scss/_position.scss
@@ -72,6 +72,7 @@
 }
 .o-position-coords,
 .o-position-find {
+  position: fixed;
   background-color: transparent;
   border: none;
   color: #fff;
@@ -82,4 +83,5 @@
 input.o-position-find {
   padding: 0;
   width:95px;
+  line-height: $line-height-large;
 }
--- a/scss/_viewer.scss
+++ b/scss/_viewer.scss
@@ -29,7 +29,7 @@
   color: white;
   font-size: 0.7em;
   height: 2.5em;
-  line-height: 2em;
+  line-height: $line-height-medium;
   z-index: 1;
   width: 100%;
 }
@@ -55,7 +55,7 @@
   padding-right: 5px;
 }
 .o-footer .o-footer-middle .o-footer-middle-content p {
-  line-height: 2em;
+  line-height: $line-height-medium;
   vertical-align: top;
   display: inline;
 }
--- a/scss/_viewer.scss
+++ b/scss/_viewer.scss
@@ -41,11 +41,6 @@
   color: white;
   text-decoration: underline;
 }
-.o-footer .o-footer-left {
-  float: left;
-  position: relative;
-  width: 30%;
-}
 .o-footer .o-footer-middle {
   float: left;
   position: relative;
--- a/scss/origo.scss
+++ b/scss/origo.scss
@@ -42,6 +42,7 @@
   @import 'featureinfo.scss';
   @import 'infowindow.scss';
   @import 'ui/validate.scss';
+  @import 'o-scale.scss';
   @import 'scalepicker.scss';
   @import 'embedded-overlay.scss';
 }
--- a/scss/ui/_input.scss
+++ b/scss/ui/_input.scss
@@ -86,7 +86,7 @@ input[type=range]:focus {
 input[type=range]::-webkit-slider-runnable-track {
   height: 4px;
   cursor: pointer;
-  animate: 0.2s;
+  animation: 0.2s;
   background: $primary-color;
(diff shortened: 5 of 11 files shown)


tonnyandersson:
Looks good at first glance. But! If you add the scalepicker control _after_ the position control, the scalepicker control will cover the mouse coordinates.

flutter

Also, the texts are not perfectly aligned. But you already know that.
image

jaha7:
The issue @tonnyandersson mention also appear using the "old" scalecontroll
origo_postition_probs

tonnyandersson:
@jaha7 Unable to reproduce. What browser are you using?

scalepicker

mulfvik:
Its dependent by in what order you call the controls so this shouldnt be dependent of config order i guess
captured

jaha7:
tested "old" [scale-control](origo-map.github.io/origo-documentation/latest/#scale-control) with Chrome, Edge (chromium) and Firefox, same results
in previous example we have temporary disabled "always on" because our performance issue with #914 

new example, position-control first
origo_postition_probs2

new example,  scale-control first
origo_postition_probs3

mulfvik:
I know this has been up for discussion before and we chose the alternative that scalepicker could render both before and after position control.
Now I think it looks a bit weird when scale controls is pushed aside when mouseposition coordinates is activated.
Any other thoughts on this?

tonnyandersson:
@jaha7 Sorry, my bad. My mind went straight to the newer scale _picker_ control. You are right, though. The fixed coordinates will cover the old scale control (or the other way around).

@mulfvik With the scale _picker_ control, I had the ambition to make it so it could be placed pretty much anywhere, in any toolbox. I totally agree with the weirdness of scale controls being pushed aside, though. But I also like the idea of having a high level of freedom...

However, we can't allow controls to be placed on top of each other. So, in my opinion, either we make _all_ footer controls fixed in some way, or none.

mulfvik:
Like the part about freedom, reminds me of Mel Gibson in Braveheart. Ill dig in to this and see what comes up.

jaha7:
@tonnyandersson Alright, no probs, we like the scalepicker and in my opinion the "old" scale-control is not needed because of scalepicker-control (if neccecary with option that disable droplist for those usercases)

maybe coordinates could show last in "footer-controls" row regardless of scalepicker (or other control being in between)?

mulfvik:
@jaha7 that sounds like the prettiest alt. so far.

Also, coordinates only show if screensize width >1158px

mulfvik:
Tested some more and suggestion to let position render last and override control init order sounds good for me.
What do you think @tonnyandersson ?
captured

jaha7:
Clarification of the suggestion i mention, i undertand this can be more tricky.

Screenshot example fixed in paint :)  (scalepicker after position, new placeholder after last control)
paint

Another suggestion, coords in same box, text for no coords needed.
paint2

tonnyandersson:
@mulfvik I'm kinda into the whole "freedom" thing... Also, the position control looks best when positioned first in the footer.

@jaha7 Two really nice options. The first option doesn't feel 100% intuitive, though, since the coordinates are a part of the position control, with the crosshair and projection changing functionality and all. I think it works best if they are kept together.

Which is why I like the second option more. Perhaps it will look weird if the coordinates are hidden when the mouse pointer is outside the map, but it will probably not look any weirder than the current sidestepping scale controls. We could probably let the coordinates show as well, even when the mouse pointer is outside the map, by "freezing" the last coordinates shown before mouseOut.

tonnyandersson:
@mulfvik After some diggin', I found a solution that seems to fix the fluttering legend issue, without touching the position control. Just change `overflow-y` to `scroll` in the `contentStyle` parameter on row 96 of src/controls/legend/overlays.js:

https://github.com/origo-map/origo/blob/6e738bbc51305ae4d5690922d0691f622c3e0747/src/controls/legend/overlays.js#L90-L101

This fixes the fluttering issue, but I'm not 100% certain it won't break anything else. Poked around for a bit but couldn't find anything leaking. It won't fix the sidestepping scale controls issue, though. but neither #955 nor #956 is about that anyway. So...

mulfvik:
Nice one @tonnyandersson . I think im on to fixing the position. Else lets go for your solution.

mulfvik:
Went with @tonnyandersson flutter fix. Then added option to define text or freeze coordinates at last position.
`{
      "name": "position",
      "options": {
        "noPositionText": "No position access"
      }
 },`
If `"noPositionText": ""` last coordinates show.

jaha7:
Nice work, tested and for me its working fine.

There's just one more thing... :)
When click/hold "pandrag" the map, coordinates jumping around. 
coords_origo
In Origo v1 the coordinates locks.

#914 can also be closed after this PR

tonnyandersson:
@mulfvik You know what? I don't like my flutter fix anymore. Sure, it fixes the fluttering, but it looks like crap in IE, for example, since the scrollbar is permanently visible when using `overflow-y: scroll`:

image

Looks better in Chrome, but you can see the "empty" space set aside for the scrollbar:

image

Sadly, I don't have any better ideas right now.

mulfvik:
@jaha7 Check out Jumping coordinates commit if it solves it.

mulfvik:
@tonnyandersson Ok, back to it then :)

jaha7:
@mulfvik Very nice, It solves the problem and the PR works really well for me overal because legend not affecting rendering time when zooming and panning.

Though there is one CSS bug left that i see.
css_p

mulfvik:
@tonnyandersson I can live with the scrollbar design in IE. The usage of IE is so low at this point.

mulfvik:
@jaha7 I think the footer should be closed if not all the controls fit. Or maybee in this case if position control and scalepicker or scale control fit. Otherwise its hard for the user to know why

@tonnyandersson Looking for media querys variables but cant find any, i.e. `$media-l`. Do you know where they are located? Part of a library that is being called from somewhere in the void?

jaha7:
@mulfvik I agree, sounds like a good solution. Test using width 40%,20%,40% (left, center, right footer content) also give some more space for footer-controls from the left.

mulfvik:
@tonnyandersson Can you have a look at latest commit please. Might be ready for merge...

tonnyandersson:
@mulfvik Nope, still got the flutters.

Here's my final suggestion: change class `flex` to `inline-block` on the main legend element. This will probably screw up something else, but I can't see any negative changes my self.

mulfvik:
@tonnyandersson  Weird, could have been the cache even though...
 Cant either see any negativ consequences, good job!

tonnyandersson:
Good work, @mulfvik! Just remember to add a few lines about the `noPositionText` option in the documentation.

PR #966, opened 2020-06-24
Added global variable filtering

### LINKED ISSUE #965: Filtering by global variables
It would be awesome if it was possible to filter layers based on global variable values, for example values stored in sessionStorage or localStorage.

tonnyandersson:
Fixes #965. This is only available for layer types where filtering is already implemented, i.e. WFS and AGS_FEATURE.

Configure like this:

`"filter": "someAttribute={{someGlobalVariable}}"`

or

`"filter": "someAttribute={{sessionStorage.someVariable}}"`

### DIFF
--- a/src/layer/agsfeature.js
+++ b/src/layer/agsfeature.js
@@ -2,6 +2,7 @@ import EsriJSON from 'ol/format/EsriJSON';
 import VectorSource from 'ol/source/Vector';
 import * as loadingstrategy from 'ol/loadingstrategy';
 import vector from './vector';
+import replacer from '../utils/replacer';
 
 function createSource({
   attribution,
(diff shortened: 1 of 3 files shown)


mulfvik:
When testing filter I noticed that only the = operator worked. 
`"filter": "column_name = '5'"`
This wont:
`"filter": "column_name LIKE 'a%'"`
`"filter": "column_name > '5'"`
Is this syntax wrong?

tonnyandersson:
@mulfvik No, your syntax is (somewhat) correct. My code, however, is wrong.

tonnyandersson:
@mulfvik Should be fixed now. Please have another look before I merge.

mulfvik:
@tonnyandersson Looking good to merge, thumbs up

PR #967, opened 2020-06-24
Possible for plugins to add params to mapstate

### LINKED ISSUE #959: Possibility for plugins to use mapstate functionality
In order to have changes in the map from a plugin to carry over to the shared map, it seems that each plugin would need its own functions in the core to add to and read from the mapstate object. Because of this, the core has to be adjusted for the plugin and not vice versa.

I suggest a more generalised option where the plugin can add a callback function to the core. The callback function assumes that it gets the state object as a parameter which gives the plugin the option to add an object with the properties the plugin requires.

The callback function can be added through exposing a new object from permalinkstore.js through the sharemap controller. The plugin can then fetch the sharemap controller through viewer and then add a key and its callback function to the object. When the map is saving the state, all the added callbacks can be called to add them to the mapstate.

When reading the mapstate, the plugin can get the saved state through viewer's getUrlParams()-function and use the properties as needed.

**Additional context**
As Eskilstuna is using the layermanager plugin, the described solution has been explored with layermanager to see if the added layers carries over when using sharemap with mapstate.

How the plugin can add its callback function to be used in getmapstate. (layermanager)
image

How a callback function can look like, state is assumed to be the mapstate. (layermanager)
image

How the plugin can read from the mapstate. (layermanager)
image

How the addition of the callback look in the core. (permalinkstore.js)
image

A loop that runs the callbacks when saving the mapstate and provides the mapstate object. (permalinkstore.js)
image

Bimane1900:
resolves #959 

Tested by having added layers (added with layermanager) in a map and then sharing the map to see if the added layers are still in the map. Sharemap used "storeMethod": "saveStateToServer".

The changes done in the Origo core can be seen in the commit, but the plugin needs to take advantage of them for it to work.

Example usage with specific code from Eskilstuna's layermanager:

Gets the sharemap controller and adds a key and a callback function.
image

Callback function assumes the state object as parameter.
image

Gets the mapstate params through viewer and uses them as needed.
image

### DIFF
--- a/src/controls/sharemap.js
+++ b/src/controls/sharemap.js
@@ -32,6 +32,9 @@ const ShareMap = function ShareMap(options = {}) {
 
   return Component({
     name: 'sharemap',
+    addParamsToGetMapState(key, callback) {
+      permalink.addParamsToGetMapState(key, callback);
+    },
     onInit() {
       if (storeMethod && serviceEndpoint) {
         permalink.setSaveOnServerServiceEndpoint(serviceEndpoint);
--- a/src/permalink/permalink.js
+++ b/src/permalink/permalink.js
@@ -56,10 +56,14 @@ export default (() => ({
         .then((data) => {
           const mapObj = {};
           Object.keys(data).forEach(key => {
-            mapObj[key] = permalinkParser[key](data[key]);
+            if (permalinkParser[key]) mapObj[key] = permalinkParser[key](data[key]);
+            else mapObj[key] = data[key];
           });
           return mapObj;
         });
     }
+  },
+  addParamsToGetMapState: function addParamsToGetMapState(key, callback) {
+    permalinkStore.AddExternalParams(key, callback);
   }
 }))();
--- a/src/permalink/permalinkstore.js
+++ b/src/permalink/permalinkstore.js
@@ -2,6 +2,7 @@ import urlparser from '../utils/urlparser';
 
 let getPin;
 const permalinkStore = {};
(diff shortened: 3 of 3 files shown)


Bimane1900:
Similar to #968 , this PR could be tested with a minimal plugin that tests the specific code related to the PR. Which makes the testing easier as plugins might otherwise require an environment to be set up. The tests of this PR would consist of sharing information related to a plugin using sharemap.

**This feature still needs a working Origo-server in the environment, since it is using mapstateId.**

For this PR, I created a small plugin that adds a textarea in the middle of the map. The text within it should still be there in the shared map. The text is shared by taking advantage of the changes in the code.

How to test with [MapStateParamsTester.zip](https://github.com/origo-map/origo/files/4897035/MapStateParamsTester.zip) : 
1. Download and unpack the zip.
2. Copy the js-file from the build folder to your origo folder.
3. Make sure you use the correct config in index.json:
```
{
      "name": "sharemap",
      "options" : {
        "serviceEndpoint": "Your Origo-server URL",
        "storeMethod": "saveStateToServer"
      }
}
```
4. Include it in your index.html `<script src="./mapparamstest.js"></script>`
5. Add it to map on the load event:
```
origo.on('load', (viewer) => {
	let mapStateParamTester = MapStateParamTester();
	viewer.addComponent(mapStateParamTester);
})
```
6. Run the map and you should see a textarea in it. Type something in the textarea and share the map.
7. Is the text still there?
8. Change the plugin and rebuild it if you want to test it in some other way.

PR #968, opened 2020-06-24
Add buttons to legend tools, available for plugins

### LINKED ISSUE #964: A way for plugins to add buttons to legend
There does not seem to be a good way for plugins to add UI elements to the legend. For example, a button that triggers the plugin's GUI.

Currently, the turnofflayer and layermanager control can be in a tools component. This component could be exposed to plugins for them to add their own buttons to it. Some function called "GetToolsCmp()" or similar could be called from the legend component, which could return needed component.

Plugins can create its own button and use "GetToolsCmp()" to get the tools component, and then add the button. 

The tools component is mentioned since it already exists for other buttons, but the idea is to have legend expose some component that the plugins can use. It does not necessarily have to be the mentioned tools component.

**Additional context**
The layermanager button is currently added in legend.js, which seems odd and it makes more sense for the plugin to add the button itself somehow.
Eskilstuna's layermanager plugin currently depends on local changes in Eskilstuna's map to have layermanager add its button to the legend (by exposing the legend groups components). I think a solution like the described one would benefit Eskilstuna's fork, but also others that might want to add buttons related to plugins to the legend. 

The tools components that has the layermanager and turnofflayers button.
image

Bimane1900:
resolves #964 

Instead of creating a getter for the component, as described in the issue, I went with an add function instead.
The addButtonToTools()-function handles logic depending on how many buttons has been added. If there is only one button it will be to the right in Legend, but additional buttons will fill from the left. The tools component will be added regardless if any buttons will be added, but it will be hidden until a button is added.

As can be seen in the commit, the button in legend.js can also use the function.

No buttons:
image

One button:
image

image

Two buttons:
image

Four buttons:
image

How a plugin can add a button:
image

### DIFF
--- a/src/controls/legend.js
+++ b/src/controls/legend.js
@@ -31,6 +31,7 @@ const Legend = function Legend(options = {}) {
   let layerButton;
   let layerButtonEl;
   let isExpanded;
+  let toolsCmp;
   const cls = `${clsSettings} control bottom-right box overflow-hidden flex row o-legend`.trim();
   const style = dom.createStyle(Object.assign({}, { width: 'auto' }, styleSettings));
 
@@ -157,6 +158,19 @@ const Legend = function Legend(options = {}) {
 
   return Component({
     name,
+    addButtonToTools(button) {
+      const toolsEl = document.getElementById(toolsCmp.getId());
+      toolsEl.classList.remove('hidden');
+      if (toolsCmp.getComponents().length > 0) {
+        toolsEl.style.justifyContent = 'space-between';
+        toolsEl.insertBefore(dom.html(divider.render()), toolsEl.firstChild);
+        toolsEl.insertBefore(dom.html(button.render()), toolsEl.firstChild);
+      } else {
+        toolsEl.appendChild(dom.html(button.render()));
+      }
+      toolsCmp.addComponent(button);
+      button.onRender();
+    },
     onInit() {
       this.on('render', this.onRender);
     },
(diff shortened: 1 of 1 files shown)


Bimane1900:
Since testing some features related to plugins might require a specific testing environment. I suggest that a PR can attach a minimal plugin that just tests the specific code or feature. At least when it comes to features related to how plugins communicate with Origo. Such as this PR, which is general to plugins and not specifically to layermanager.

As for this PR, I made this minimal testing plugin that just tests the function `addButtonToTools`. 

How to test with [AddButtonTesterPlugin.zip](https://github.com/origo-map/origo/files/4895595/AddButtonTesterPlugin.zip):
1. Download and unpack the zip.
2. Copy the js-file from the build folder to your origo folder.
3. Include it in your index.html `<script src="./addBtnTester.js"></script>`
4. Add it to map on load event: 
```
               origo.on('load', (viewer) => {
			let addButtonTester = AddButtonTester();
			viewer.addComponent(addButtonTester);
		})
```
5. Run the map and you should see a button in the legend which presents an alert when you click it.
6. Change the plugin and rebuild it if you want to test in some other way, (Maybe I did not think of the best way to test).

Testing this way is merely a suggestion, there might be cons that I have not thought about.

tonnyandersson:
@Bimane1900 Works fine and a great addition! Also, I like the idea with the zipped test plugin. It makes the testing procedure much smoother.

I think we can merge this after the conflicts have been resolved.

Bimane1900:
@tonnyandersson The conflicts has been resolved!

Issue #971, opened 2020-07-08
Need better solution for footer children


mulfvik:
@mulfvik Very nice, It solves the problem and the PR works really well for me overal because legend not affecting rendering time when zooming and panning.

Though there is one CSS bug left that i see.
![css_p](https://user-images.githubusercontent.com/47249045/86611533-145c5980-bfaf-11ea-80bf-c0b18cb7af24.gif)

_Originally posted by @jaha7 in https://github.com/origo-map/origo/pull/958#issuecomment-654315418_

PR #973, opened 2020-07-08
Minimizes middle content

mulfvik:
Improves #971


tonnyandersson:
@mulfvik Seems to do the job. But you need to resolve the conflicts.

PR #975, opened 2020-07-14
Adds visual indication on groups with active layers

### LINKED ISSUE #970: Indication when a group has visible layers within it
Whenever a group has active(visible) layers the group itself would indicate it. I think it would ease the user experience, especially if you have many groups. How the group would show that it has visible layers is up to discussion, i would suggest a slightly different background color or something similar.

Zakuto3:
fixes #970 
Implementation of visual suggestion nr 3 with option to turn it off via index.json.

```
   {
      "name": "legend", 
      "options": {
        "labelOpacitySlider": "Opacity",
        "useGroupIndication" : false
      }
    }
```

### DIFF
--- a/index.json
+++ b/index.json
@@ -39,7 +39,8 @@
     {
       "name": "legend",
       "options": {
-        "labelOpacitySlider": "Opacity"
+        "labelOpacitySlider": "Opacity",
+        "useGroupIndication" : true
       }
     },
     {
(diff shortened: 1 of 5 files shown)


jokd:
Works well but I suggest adding a css class instead of adding a style to the element, and in the css use a global color variable (origo blue that is). That way it will be easier to use different color schemes in the future and avoid styling in the js.

Bimane1900:
The css is now controlled with classes. Additionally, transitions was added to have it be easier on the eyes.

PR #976, opened 2020-07-15
Removes layermanager button in legend.js

### LINKED ISSUE #974: Remove layermanager button from legend.js
Since #968 has been merged now, plugins such as layermanager can add their own buttons. Therefore, the layermanager button that legend.js adds is no longer needed and should be removed to keep the code of plugins within the plugins themselves.

Bimane1900:
fixes #974 
This PR remove the layermanager button from main origo. The button should be added in the plugin itself, which is possible with legend.js function: `addButtonToTools(button)`.

### DIFF
--- a/src/controls/legend.js
+++ b/src/controls/legend.js
@@ -14,7 +14,6 @@ const Legend = function Legend(options = {}) {
     contentCls,
     contentStyle,
     turnOffLayersControl = false,
-    layerManagerControl = false,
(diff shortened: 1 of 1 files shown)


jokd:
Will you also fix the plugin so it won't break after this commit?

Bimane1900:
Yes. Will create an Issue and PR for it on origo-map/layermanager.

PR #996, opened 2020-09-15
Added jsPDF to package and removed download for jsPDF

### LINKED ISSUE #988: Not able to download jsPDF script
**Describe the bug**
The print function of creating PDFs aren't working since the jsPDF script can't download.

**To Reproduce**
1. Go to 'print'
2. Click on 'Skapa pdf'
3. No pdf is created

**Expected behavior**
A pdf shlould be created.

**Screenshots**
Error messages in Firefox
image

**Additional context**
In src/utils/download.js the following file is specified, which dosen't work
https://unpkg.com/jspdf@latest/dist/jspdf.min.js
If this file is used it works: https://cdnjs.cloudflare.com/ajax/libs/jspdf/2.1.1/jspdf.umd.min.js
It would suggest that it is something wrong with the unpkg.com server. Anyway we should perhaps do this in another way.

johnnyblasta:
Fixes #988


johnnyblasta:
Works when run local, but fails on build.

johnnyblasta:
It works building now, thanks @daRasmussen 
Also removed some more code that isn't used anymore.

daRasmussen (approved):

daRasmussen:
@johnnyblasta great stuff, keep up the good work.

Issue #987, opened 2020-09-08
Externalurl control error


mulfvik:
**Describe the bug**
When using the control "externalurl" I get this in the console:
![image](https://user-images.githubusercontent.com/24228538/92451471-09939000-f1bd-11ea-8af3-fc5b4c4a9d67.png)

jaha7:
Get similar in console when try to use "measure-elevation" (origoserver with HöjdDirekt)
![origo_probs](https://user-images.githubusercontent.com/47249045/93213791-9b267300-f764-11ea-88f4-c00bed8d4766.JPG)

johnnyblasta:
Closed by #998

PR #998, opened 2020-09-18
Avoid returning empty object

### LINKED ISSUE #997: External link and measure height not working
**Describe the bug**
When clicking on external url button or measue height nothing happens.

**To Reproduce**
If you have externalurl and/or measure controls activated, you won't be able to go to the external url or measure elevation.

johnnyblasta:
Fixes #997


johnnyblasta:
Also fixes #987 since it's the same error

daRasmussen (approved):

PR #1001, opened 2020-09-23
Adds abstract for groups

### LINKED ISSUE #904: Abstract for groups
Sometimes it would be useful to show information (abstract) for a group, just like with layers and background maps. This was possible in older version of Origo.

One solution could be to get "group info" when clicking on a group. But then the behaviour for expand and collapse groups must be changed. You would actually have to click on the arrow to expand and collapse the group. I'm not sure this is the best solution though. 

Another symbol next to the arrow? But it doesn't seem logical since layer info doesn't work that way. Any other ideas?

jokd:
Fixes #904

### DIFF
--- a/src/controls/legend/group.js
+++ b/src/controls/legend/group.js
@@ -15,6 +15,7 @@ const Group = function Group(options = {}, viewer) {
     title,
     name,
     parent,
+    abstract,
     position = 'top',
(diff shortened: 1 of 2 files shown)


daRasmussen (approved):
LGTM!

PR #1003, opened 2020-09-24
Fixed issue caused by moved util function

martinkero:


daRasmussen:
@martinkero what issue does PR resolve? If issue not submitted, please sumbit an issue and explain the problem.

PR #1008, opened 2020-10-12
Enhanced measuring tool

### LINKED ISSUE #391: Enhanced measuring tool
Measures length and area. Shows total length or area together with the length of each segment.

Based on an older Origo measure function version and further developed for the Open eMap version of Origo.

### LINKED ISSUE #982: Undo in measure control
I often misclick while measuring, especially on mobile, and the lack of an "undo" option causes me to either redo the entire measurement or give up.

A new button, either next to the existing measure control buttons or floating next to the last node, would be a simple way to visualize an undo option. I would also suggest binding the keyboard backspace key to the "undo" functionality.

An undo button would be somewhat redundant if the measure control allowed general editing of its shapes, but I expect that would be a much larger undertaking. In its place an undo option can go a long way.

Example images:
[Undo Button](https://karta.haninge.se/devmaps/origo-map-issue-982/origo-measure-undo.png)
["Subtract" last node when pointing with the cursor](https://karta.haninge.se/devmaps/origo-map-issue-982/origo-measure-undo-by-cursor.png)

tonnyandersson:
Fixes #391 
Fixes #982 

Adds option to show segment lengths. Also adds button to remove last node (only applies to unfinished measurements).

Config:

```
{
  "name": "measure",
  "options": {
    "showSegmentLengths": true
  }
}
```

Please note that the map click event was replaced because, in some extreme clicking situations, the map click event would not fire at all while the draw interaction would still add a node to the measure feature, resulting in non-labelled segments. And we can't have non-labelled segments. We just can't.


daRasmussen (commented):
Undo button works fine. 

But I see no line segments lengths.

tonnyandersson:
@daRasmussen Did you add the ```"showSegmentLengths": true``` option in the config?

daRasmussen:
@tonnyandersson yes

tonnyandersson:
@daRasmussen That's really weird. 

However, I just noticed a few bugs. It seems I didn't properly test the undo function with segment lengths turned off...

daRasmussen:
@tonnyandersson well I guess undo works like a reset it removes all previous points, when I tested it. I guess undo could remove the last created point. Not sure if this rest functionallity is desired in this PR.

tonnyandersson:
@daRasmussen Yeah, well, it is supposed to actually remove the _last created node_. Which it did, all the way until I made a last minute bug "fix", which in turn introduced this "kill 'em all" bug.

New fix is incoming.

daRasmussen:
@tonnyandersson great success!! works good both segements and undo (removes last point), tested both line and polygon.

jokd (approved):
Nice!

PR #1010, opened 2020-10-16
Compress PDF to reduce file size

### LINKED ISSUE #1009: Print creates very large files
The PDFs created by print control is very large, 54 MB for a A4 page, is there a way to reduce the size.

A compressed pdf should be produced which is much smaller in size.

johnnyblasta:
Fixes #1009


jokd (approved):
Works well!

PR #1018, opened 2020-11-11
Enhanced measure control on touch devices

### LINKED ISSUE #1016: Elevate the precision of certain touch interactions
When using touch devices it can be useful or even critical to be able to add nodes more accurately when editing or measuring.

Something similar to the crosshair used to pick coordinates can be used along with a button to elevate the precision of touch interactions.

Comments on this?

jokd:
Fixes #1016

### DIFF
--- a/scss/_position.scss
+++ b/scss/_position.scss
@@ -42,7 +42,7 @@
   }
 }
 .o-position-marker {
-  background-color: rgba(255,255,255,0.4);
+  background-color: rgba(255,255,255,0.2);
   border-radius: 50%;
   cursor: default;
   padding: .25rem;
--- a/src/controls/measure.js
+++ b/src/controls/measure.js
@@ -398,7 +398,10 @@ const Measure = function Measure({
     measure = new DrawInteraction({
       source,
       type,
-      style: style.createStyleRule(measureStyleOptions.interaction)
+      style: style.createStyleRule(measureStyleOptions.interaction),
+      condition(evt) {
+        return evt.originalEvent.pointerType === 'mouse';
+      }
     });
 
     map.addInteraction(measure);
@@ -477,7 +480,8 @@ const Measure = function Measure({
     const eventObject = {
       clientX: pixel[0],
       clientY: pixel[1],
-      bubbles: true
+      bubbles: true,
+      pointerType: 'mouse'
     };
     const down = new PointerEvent('pointerdown', eventObject);
     const up = new PointerEvent('pointerup', eventObject);
@@ -493,6 +497,9 @@ const Measure = function Measure({
     } else {
       if (showSegmentLengths) document.getElementsByClassName('o-tooltip-measure')[1].remove();
       measure.removeLastPoint();
+      if (touchMode) {
+        centerSketch();
+      }
     }
   }
 
--- a/src/controls/measure.js
+++ b/src/controls/measure.js
@@ -400,21 +400,24 @@ const Measure = function Measure({
       type,
       style: style.createStyleRule(measureStyleOptions.interaction),
       condition(evt) {
-        return evt.originalEvent.pointerType === 'mouse';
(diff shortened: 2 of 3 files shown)


tonnyandersson:
@jokd Looks nice! But I noticed a few bugs/unexpected features.

- When the measure tool is activated (but before any measuring has been initiated), the "Click to start measuring" tooltip pops up whenever the screen is touched
- Tapping on the screen while using the measure tool in _touch mode_ results in a new node. Perhaps _touch click_ could be disabled in _touch mode_? (Note that this is true for _tap-and-hold_ as well, please see the [_dragVertexDelay_](https://openlayers.org/en/latest/apidoc/module-ol_interaction_Draw-Draw.html) option)
- Removing all nodes of an active measure feature with the undo button will make the map go bananas

Tested in IOS Safari.

jokd:
@tonnyandersson Can it be useful to be able to both click in the map and the button in touch mode?
The tooltip issue is old I think and has nothing to do with this functionaligt. But now when we know the user is on touch device we can easily hide the tooltip.
Yeah removing nodes messes things up, has something to do with it not updating the sketch as supposed. Will look into that.

tonnyandersson:
@jokd To me, it's just confusing with touch and button at the same time, and it makes the tool feel a bit flimsy. _If_ anyone wants to keep the touch functionality, I say we go for a switch. Perhaps it's nice to use touch for low precision drawing, but for measuring, button's the way to go in my opinion.

jokd:
I removed the touch for touch devices....
And spent too much time wrestling the undo function just to realize it is a bug that is fixed in the dev branch.

tonnyandersson:
@jokd But not fixed in this branch, or...? Map still going bananas when using the undo button.

Steps to reproduce error in desktop Chrome:

- Open dev console and switch to mobile device mode
- Reload the map to activate "touch mode"
- Activate the measure tool and add one node or more, but don't finish the sketch
- Undo _all_ added nodes
- Pan the map
- Curse

jokd:
You are right, there was a similar issue that’s solved but I’ll fix this as well.

tonnyandersson:
@jokd Better! But fixing one bug will almost always reveal others...

When using the area measure tool with `"showSegmentLengths": true`, the "current" segment length is displayed as 0:

image

Also, clicking the _Undo_ button when using the area measure tool will remove the nodes in a weird order, starting with the second to last node. Actually, it seems to start with a non-existing node, then proceeding to the second to last node. Strange.

jokd:
@tonnyandersson Thats what happens when you are not using the dev branch or OL, Are you sure you have that version?

tonnyandersson:
@jokd No, I usually test stuff with an officially released version of OL... :)

Oh, an other thing. I'm a bit concerned about the placement of the _Add_ button. I find it far too easy to hit either the _Area_ button or the _Undo_ button by mistake. But maybe that's just me?

jokd:
Well I wrote earlier that this fix is depending on a bug fix in the dev branch of OL. Well, the length issue is however another issue Ill try to fix.

Regarding your big thumb I cannot do much about that.

tonnyandersson:
@jokd 

Regarding your first statement, you only mentioned "the dev branch", without specifying of what repository/library.

Regarding your second statement, you're absolutely right.

jokd:
Thinking about that total number... doesnt it make sense to include the last part of the polygon as well? The one from the last point back to the beginning that is. It is after all part of the sketch.

tonnyandersson:
@jokd I'd say it depends. But your suggestion is probably the most common case. I understand if someone would like to know the total length of segments _so far_ when measuring, but I can't come up with a resonable use case. Knowing the _total circumference_ at any given point is probably more useful. I think, however, that's a separate issue.

On that note, I think this one is good to go. Works quite well now. Good job!

tonnyandersson (approved):

PR #1024, opened 2020-11-13
Fix for IE 11 legend scroll issue

### LINKED ISSUE #1017: Legend max height problems in ie11
**Describe the bug**
When the legend reaches the maximum height, the scroll bar, background maps and the 'turn-off-layers' button disappear and the legend gets unscrollable. This only occurs in ie11.

**To Reproduce**
1. Create a map with latest origo
2. Click on 'legend'
3. Make the legend reach max height (either by changing the height of the browser or expand multiple groups)

 - OS: Windows 10
 - Browser: Internet explorer 11

**Screenshots**
image

**Additional context**
Does we still support ie11?

tonnyandersson:
Fixes #1017. 

Reverted the legend `display` property back from `inline-block` to the original `flex` setting, thus _undoing_ part of the fix that solved #955. Curiously enough, it did not bring back the abstract flutter issue, at least not what I can see.

Please, when testing this, also try to reproduce the background maps abstract flutter issue, just to make sure I didn't manage to reintroduce that issue. Please note that the fluttering was mainly a Chrome issue.

### DIFF
--- a/src/controls/legend.js
+++ b/src/controls/legend.js
@@ -32,7 +32,7 @@ const Legend = function Legend(options = {}) {
   let layerButtonEl;
   let isExpanded;
   let toolsCmp;
-  const cls = `${clsSettings} control bottom-right box overflow-hidden inline-block row o-legend`.trim();
+  const cls = `${clsSettings} control bottom-right box overflow-hidden flex row o-legend`.trim();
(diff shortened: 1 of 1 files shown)


jokd:
Tested in IE, Chrome and Edge and seems to work as supposed to.

jokd (approved):

daRasmussen (commented):
LGTM

Issue #961, opened 2020-06-22
Print - Get higher resolution


johnnyblasta:
**Is your feature request related to a problem? Please describe.**
When printing it should be possible to set the resolution to get a higher quality of the printed map.

**Describe the solution you'd like**
In the print settings there should be a option for chosing the print quality. It would probably suffice to have 3 buttons with low, mid and high as options.

**Describe alternatives you've considered**
To be able to set an arbitrarily resolution might be a option, but don't know if it's necessary.

**Additional context**
Not only the resolution of the map image should be changed, but also the map element on the map should be changed.

jokd:
Check this out, new in 6.4.0 release:
https://github.com/openlayers/openlayers/pull/10537
https://openlayers.org/en/latest/examples/print-to-scale.html

Issue #903, opened 2020-04-15
Print - See the scale


asemoller:
**Is your feature request related to a problem? Please describe.**
It would be really nice if one could see the scale when using the print control.

**Describe the solution you'd like**
As text, like "1:4000". Maybe above or below the scalebar? Or does anyone else have any idea where to put it?

PR #1032, opened 2020-11-20
Enable to set scale and resolution on print PDF

### LINKED ISSUE #960: Print - Be able to set the scale
When printing there should be a option to set the scale of the map.

In the settings window there should be a option for selecting scale with a dropdown list with the different scales. Use the resolutions of the map to populate the dropdown with the scales

Other ways of selecting the scale is probably no good since the list can be quite long, up to 10-15 scales.

**Additional context**
Once you've set the scale it's good to see on the map which scale it has, but that is a separate issue #903

johnnyblasta:
Makes it possible to set the scale and resolution for the PDF and to get the scale correctly shown on the map. Fixes #960 #961 #903

The scales can be configured by adding a list of scales otherwise the resolutions from config is used. There is also a option to don't show the scale on start, default is true.
` {
        "name": "print",
        "options": {
          "showScale": true,
          "scales": [
            "1:100000",
            "1:50000",
            "1:15000",
            "1:10000",
            "1:5000",
            "1:3000",
            "1:2000",
            "1:1000",
            "1:500"
          ]
        }
    },
`
The list of scales in configuring the print control must be within the resolutions of the map. If a scale is lower than the lowest resolution of the map it won't be able to set that scale.

image


jokd:
It would be nice if changing the resolution didn't affect the size of the header, margin, scale bar etc. but only the map itself.
Panning in other than 75dpi is a bit awkward but I suppose that has to do with the scaling.

Can zooming in the map also change the selected value for scale in the settings?

Issue #1030, opened 2020-11-19
Settings in print control to large for small screens


johnnyblasta:
**Describe the bug**
The settings window in print control can be to large for small screens, so that not all settings is visible.

**To Reproduce**
Open print control and settings on a small screen and not all settings is visible and only way to see them are to change the scaling on the webbrowser.

**Expected behavior**
There should be a scrollbar so that is possible to scroll down to view all settings.

johnnyblasta:
A alternative could be scaling the settings box as long as it's still legible.

PR #1034, opened 2020-11-27
Formatting of the title

### LINKED ISSUE #1033: Print - Be able to format the title and description
It would be nice to be able to change the alignment of the title/description and also change the size.

Next to the title there should be a button to open options for adjustments, see image.
image
When opened button for alignment and dropdown for size should be shown, see image.
image

johnnyblasta:
Fixes #1033 

Makes it possible to change alignment and size of the title for the printed map.
image

### DIFF
--- a/scss/ui/helpers/_sizes.scss
+++ b/scss/ui/helpers/_sizes.scss
@@ -26,7 +26,9 @@
 .width-100[style] {
   width: 100% !important;
 }
-
+.width-88[style] {
+  width: 88% !important;
+}
 .height-125 {
   height: 1.25rem;
 }
--- a/src/controls/print/description-control.js
+++ b/src/controls/print/description-control.js
@@ -123,10 +123,15 @@ export default function DescriptionControl(options = {}) {
         </div>
       </div>
       <div class="hidden" id="${formatId}">
+      <div class="padding-smaller">
         <h6>Justering beskrivning</h6>
         ${alignControl.render()}
+      </div>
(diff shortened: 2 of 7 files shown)


johnnyblasta:
Also adding the same formatting options for description.

jokd:
Works well. Maybe use the same checkboxes as the size rather than a dropdown, and name the sizes small-big rather than h1-h5?

johnnyblasta:
> Works well. Maybe use the same checkboxes as the size rather than a dropdown, and name the sizes small-big rather than h1-h5?

I opted out of using buttons since it was more than 3 sizes, so hasn't tested how it would look, but perhaps it could work. I used the classnames for the header, since it was the easiest. Don't know what to call all of the sizes :)

jokd:
Perhaps 3 sizes are enough?

johnnyblasta:
> Perhaps 3 sizes are enough?

Perhaps, since there are a very slight difference between some of the 5 headings.

johnnyblasta:
Made the sizes configurable, kept the dropdown.

jokd:
I guess I can live with it being a dropdown, but I find the settings page a bit messy now when there's a lot of stuff going on. Uniforming the spacing between headers and content and adding a delimiter between the expanded description settings and the page size setting might straighten things up:
image

This is related to #1030 as well.

jokd (approved):

johnnyblasta:
Fixed spacing and added divider

image

PR #1037, opened 2020-12-04
Handle overflow in print settings with scrollbar

johnnyblasta:
image


jokd:
Tested and works well in Edge and Chrome. Could have some spacing in the bottom but I'll approve it.
image

jokd (approved):

johnnyblasta:
Could someone test it on iOS and other mobile devices?

johnnyblasta:
> Could someone test it on iOS and other mobile devices?

Don't behave as expected on iPhone, either in Safari or Chrome.
Works well in Internet Explorer though.

johnnyblasta:
Now it should work with iPhone also.

PR #1039, opened 2020-12-14
Added geoposition tracking mode

### LINKED ISSUE #1036: Follow me option in geoposition control
It would be nice to have a "follow me" option in geoposition control.

When turning on the geoposition and "pan to" is on the map is centered at current position, but then the map don't continue following the further moving. There should be a way to enable a "follow me" function so that the map is always centered at current position.

There should also be configurable so that "follow me" is active at start.

**Additional context**
This is a function that mostly makes sense for mobile devices, so it should perhaps not be turned/visible on desktop computers if that can be determined.

tonnyandersson:
Fixes #1036.

Add the option `enableTracking: true` to, well, enable tracking. Just like this:
```
{
  "name": "geoposition",
  "options": {
    "enableTracking": true
  }
}
```

The geolocation control now has three modes - _Initial_, _Active_ and _Tracking_.

How to use it:

1. Press the Geolocation button once to enable geolocation (_Active Mode_)
2. Press the button again to activate tracking (note the subtle icon change) (_Tracking Mode_)
3. Press the button a third time to disable geolocation (_Initial Mode_)

Please note that if you drag the map while in _Tracking Mode_, the control will automatically switch back to _Active Mode_. This allows the user to pan around the map and then resume tracking by clicking the Geolocation button again.


johnnyblasta (approved):
Tested on Android phone and LGTM

PR #1041, opened 2020-12-14
Print IE and Edge fix

### LINKED ISSUE #1040: Print to PDF don't work for IE and Edge
**Describe the bug**
Print to PDF don't work for IE and Edge pre Edge Chromium

**To Reproduce**
Try to print to PDF and nothing will happen.

**Expected behavior**
A PDF should be created.

 - Browser Internet Explorer and Edge pre Edge Chromium

johnnyblasta:
Fixes #1040

### DIFF
--- a/src/controls/print/logo.js
+++ b/src/controls/print/logo.js
@@ -23,7 +23,7 @@ export default function Logo(options = {}) {
   }
 
   return El({
-    attributes: { src: `${baseUrl}${src}` },
+    attributes: { src: `${baseUrl}${src}`, alt: 'logo' },
     cls,
     tagName: 'img',
     style
--- a/src/controls/print/north-arrow.js
+++ b/src/controls/print/north-arrow.js
@@ -64,7 +64,7 @@ export default function NorthArrow(options = {}) {
       this.dispatch('render');
     },
     render() {
-      return `<img src="${baseUrl}${src}" class="${cls}" style="${styleString}" id="${this.getId()}">`;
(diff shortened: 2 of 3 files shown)


johnnyblasta:
It would be nice if someone could approve this since I have another PR on queue.

johnnyblasta:
> It would be nice if someone could approve this since I have another PR on queue.

Fixed it with a new PR, but nice if this can be tested anyway.

jokd:
Tested in IE and it works, though it takes like 40 seconds to generate A4 pdf with OpenStreetMap. The png is generated in ~3 seconds.

johnnyblasta:
> Tested in IE and it works, though it takes like 40 seconds to generate A4 pdf with OpenStreetMap. The png is generated in ~3 seconds.

Yes, noticed it took long time creating PDF, since it's basically the previous version it would have taken a long time before also?
Tested on my test server and on Edge it took 10 seconds and on IE it took 20 seconds.

jokd:
> > Tested in IE and it works, though it takes like 40 seconds to generate A4 pdf with OpenStreetMap. The png is generated in ~3 seconds.
> 
> Yes, noticed it took long time creating PDF, since it's basically the previous version it would have taken a long time before also?
> Tested on my test server and on Edge it took 10 seconds and on IE it took 20 seconds.

Probably, I am not a frequent IE user and I think there's not much to do about it. I say we merge this.

Issue #1047, opened 2020-12-21
Add sourcemap loader


daRasmussen:
**Describe the bug**
Could not load content webpack://Origo/node_modules/....

**To Reproduce**
1. npm start
2. open any browser
3. open console
4. See warnings

**Expected behavior**
No warnings, load js sources from ol. 

**Screenshots**
![image](https://user-images.githubusercontent.com/33061427/102764588-ab5d7800-437b-11eb-9172-846af68cf032.png)

**Desktop (please complete the following information):**
 - OS: Any
 - Browser chrome

PR #1048, opened 2020-12-21
Add source-map loader

daRasmussen:
Fixes issue #1047


johnnyblasta (approved):
Tested and LGTM, nice to be rid of those messages in console.

PR #1053, opened 2020-12-21
Print settings have separate collapse container class

### LINKED ISSUE #1042: First layer assigned to the root group isn't visible
**Describe the bug**
There's no obvious way to turn off the one active layer in the default map configuration

**To Reproduce**

1. Download and run the latest Origo main
2. Check the map and legend in your favourite compatible browser

**Expected behavior**
The active layer should be visible in the legend

**Screenshots**
The initial view
initial_view
With two layers in the root group
two_things_in_root
With no layer in the root group
nothing_in_root

**Additional context**
The issue seems introduced at the "Handle overflow in settings window" commit. 

The relevant bit therein appears to be
`  &.expanded {
    height: calc(100% - 2rem);
  }` of scss/_collapse,scss

johnnyblasta:
Fixes #1042

### DIFF
--- a/scss/ui/_collapse.scss
+++ b/scss/ui/_collapse.scss
@@ -32,6 +32,10 @@
       }
     }
   }
+}
+
+.collapse-scroll {
+  @extend .collapse;
   &.expanded {
     height: calc(100% - 2rem);
   }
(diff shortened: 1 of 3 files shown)


daRasmussen (approved):
LGTM

PR #1065, opened 2020-12-28
Replaces jq carousel

daRasmussen:
replaces owl.carousel with glidejs, removes jquery dependency.


jokd:
Will this eventually replace the owl carousel? How much is left to do (thinking about the upcoming release)?

daRasmussen:
@jokd will commit more work later today.

daRasmussen:
@jokd Check out the working draft.

jokd:
Tested and works quite well. I noticed a couple of things though:
Popup/sidebar overflows when content is too big. Should be a scrollbar
When lots items are hit, the dots are on a long line instead of wrap in the popup.
Arrows and dots are sometimes rendered outside the popup.
image

Perhaps are all of these issues related and easily solved

jokd:
Solves the issue where tha map stopped working in IE. However the carousel looks a bit funny and wont load more than one item.

jokd:
One feature that is still lacking is the panning of the popup when rendered outside the map window. Besides that I think it is good to go.

PR #1086, opened 2021-01-12
Adds the exclusive option for groups

### LINKED ISSUE #796: Make layers in group exclusive
It would be nice to set layers in a group as exclusive, making it possible to view only one layer at a time. Similar to the way background maps work. This would for instance prevent multiple coverages on top of each other making the map (and server) faster.

In other words, radio buttons working as radio buttons.

jokd:
Fixes #796 
Setting "exclusive": true for a group will only allow one layer to be visible at a time. Except when loading the map, that is.


jokd:
OK to merge this?

tonnyandersson:
@jokd Seems to work as intended. LGTM.

tonnyandersson (approved):

PR #1091, opened 2021-01-14
Fixed editor bug after Jquery removal

### LINKED ISSUE #1090: Editor control not working as expected
**Describe the bug**
Editor control not working as expected

**To Reproduce**
1. Test using examples/editor.html 
2. Deactivate editor using button should close toolbar.
3. Change layer in dropdown and try to draw using + button results in error in chromedevtools "Uncaught TypeError: Cannot read property 'getSource' of undefined"

tonnyandersson:
Fixes #1090.

Instead of using `bubbles: true` to fix the non-responding editor button issue, another solution would be to dispatch the `enableInteraction` event directly on `document`. Not sure if that would cause other problems or not, though.

### DIFF
--- a/src/controls/editor/editform.js
+++ b/src/controls/editor/editform.js
@@ -1,5 +1,5 @@
 const createForm = function createForm(obj) {
-  const id = obj.elId.slice(1);
+  const id = obj.elId;
   let cls = obj.cls || '';
   cls += id;
   cls += obj.isVisible ? '' : ' o-hidden';
--- a/src/controls/editor/edithandler.js
+++ b/src/controls/editor/edithandler.js
@@ -344,26 +344,26 @@ function onAttributesSave(feature, attrs) {
     // Read values from form
     attrs.forEach((attribute) => {
       // Get the input container class
-      const containerClass = `.${attribute.elId.slice(1)}`;
+      const containerClass = `.${attribute.elId}`;
       // Get the input attributes
-      const inputType = document.getElementById(attribute.elId).attr('type');
-      const inputValue = document.getElementById(attribute.elId).val();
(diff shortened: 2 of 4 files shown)


jaha7:
Just tested the fix, changing layers and edit now working.
But attribute-form is not, when trying to save attribute-data and press Ok button in form i get this.

probs

tonnyandersson:
Thanks for the input, @jaha7! I'll look into that as well.

daRasmussen (approved):

tonnyandersson:
Fixed a few remaining bugs. Please test this thoroughly before approving, preferably with a more advanced editor config than the default config.

jokd:
Seems to work. Tested in IE without success but that is maybe related and solved in another PR. I suggest we merge this along with the other jQuery related PRs and test it again

jaha7:
Massive work, applaud!

Tested with Edge (Chromium) and different options for autoForm, autoSave, active and the combinations I tested working. Snaplayers also seems to work. But found some things that not working as expected.

- Button to activate/deactivate editor not working for me when drawplugin is enabled.
image
Maybe not related to this PR.

- Select lines when using editor working strange, problem occur only for lines, not sure if its only our usercase (geoserver wfs).
 select_lines

- Map zooms in one step when doubleclick to close line or polygon, not so big deal imo.

tonnyandersson:
Thanks for noticing, @jaha7!

1. _May_ be caused by changes in this PR. But since the error seems to occur in the draw plugin, this issue should be registered in that repo.
2. Unable to reproduce in the editor test map. Please check if the issue is present in other line type layers as well, or in older versions of Origo.
3. Known "feature". :)

PR #1094, opened 2021-01-14
Fix the title/tooltip in Legend control "Turn Off Layers" button

### LINKED ISSUE #1093: Missing tooltip for the Legend's "Turn Off Layers" button
There is a syntax error [where the title of the Legend's "Turn Off Layers" button is defined](https://github.com/origo-map/origo/blob/7081ca3c5a156a1ba6f424e010252f31bff451c0/src/controls/legend.js#L103). Additionally, because the button's icon is an SVG sprite, the title is not added even when defined properly.

1. Mouse hover over the "Turn Off Layers" button
2. Nothing happens.

Preferably, all buttons in the Legend should have tooltips that are consistent with the other controls' and buttons' tooltips (bootstrap tooltips?), but they should at least be consistent within the Legend control.

**Desktop:**
 - Any

**Smartphone:**
 - Any

MattiasSp:
Fixes #1093.

Correcting its title definition and generally letting SVG sprite Icons have their title elements set gives the "Turn Off Layers" button the same type of tooltip as the background buttons.

image

### DIFF
--- a/src/controls/legend.js
+++ b/src/controls/legend.js
@@ -100,7 +100,8 @@ const Legend = function Legend(options = {}) {
   });
 
   const turnOffLayersButton = Button({
-    cls: 'round compact icon-small margin-x-smaller\'title=\'Släck alla lager',
+    cls: 'round compact icon-small margin-x-smaller',
+    title: 'Släck alla lager',
     click() {
       viewer.dispatch('active:turnofflayers');
     },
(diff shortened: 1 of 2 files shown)


MattiasSp:
An example build is published here: https://karta.haninge.se/devmaps/PR1094/

jokd (approved):

PR #1095, opened 2021-01-14
New option hideWhenEmbedded for all controls, by including control options in Viewer's addControl

### LINKED ISSUE #1092: Option to not show Splash when embedded
When embedding a map with the Splash control, the splash modal may cover the entire view if small enough.

It would be very nice if there was an option in the control config to control whether the Splash screen shows when the map is embedded.

Not sure what the default behaviour should be, but I'm leaning towards not showing it unless an option, e.g. "showWhenEmbedded: true" is set.

image

MattiasSp:
Fixes #1092.

Adds a new option to the Splash control; `hideWhenEmbedded` which defaults to false.

If set to true, the modal is not displayed when the map is embedded.

Example config:

```
{
      "name": "splash",
      "options": {
        "title": "hideWhenEmbedded",
        "url": "splash/splash-content.html",
        "hideButton": {
          "visible": true
        },
        "hideWhenEmbedded": true
      }
},
```

### DIFF
--- a/origo.js
+++ b/origo.js
@@ -66,6 +66,7 @@ const Origo = function Origo(configPath, options = {}) {
         const controlOptions = def.options || {};
         if (controlName in origoControls) {
           const control = origoControls[controlName](controlOptions);
+          control.options = control.options || controlOptions || {};
           controls.push(control);
         }
       }
--- a/origo.js
+++ b/origo.js
@@ -66,7 +66,7 @@ const Origo = function Origo(configPath, options = {}) {
         const controlOptions = def.options || {};
         if (controlName in origoControls) {
           const control = origoControls[controlName](controlOptions);
-          control.options = control.options || controlOptions || {};
+          control.options = Object.assign(control.options || {}, controlOptions);
           controls.push(control);
         }
       }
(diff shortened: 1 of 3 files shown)


MattiasSp:
Example build available at: https://karta.haninge.se/devmaps/PR1095/

jokd:
Works well!
Another way of doing it is to include the control config in the viewerOptions parameter from origo.js to viewer and then do the check for all controls without the need to add hideWhenEmbedded to each component. That config can perhaps be useful for other stuff as well.

MattiasSp:
That's an even better suggestion, I'll get back to it 👍

MattiasSp:
Although, that PR should probably be named something else...

MattiasSp:
So with this new implementation, any control (except default controls such as `zoom`) can use the config option `hideWhenEmbedded: true` to stop the control from being added when the map is in the embedded mode.

So in addition to the splash config example above, the following now also works:
```
{
    "name": "home",
    "options": {
      "zoomOnStart": true,
      "hideWhenEmbedded": true
    }
},
```

jokd (approved):

PR #1098, opened 2021-01-15
Fix for baseUrl not effecting footer image and image style.

### LINKED ISSUE #954: baseUrl option doesn´t affect all resources
**Describe the bug**
When loading origo with baseUrl option not all resources are affected.
File path to footer image, and file path to image for image style ignores baseUrl option.

Footer image and image style seems to be relative to location of html file where Origo is loaded instead of baseUrl.

Icon styles, print logo and GeoJSON source works as expected. 

**Expected behavior**
Loading Origo in index.html
`var origo = Origo('index.json', { baseUrl: './origo/' });`

In index.json
Print logo src 
`"src": "img/png/bild.png",`
Footer image src
`"img": "origo/img/png/bild.png",`

As print logo and footer image is using the same resource, both are expected to use the same path.
As baseUrl is set to './origo/', Footer image should not need the 'origo/' part of the path in this example.

Flodkvist:
Fixes #954

Added support for use of baseurl option to also include path for
- footer image
- image layer style.

### DIFF
--- a/src/components/footer.js
+++ b/src/components/footer.js
@@ -2,6 +2,7 @@ import { Component } from '../ui';
 
 export default function Footer(options = {}) {
   const {
+    baseUrl,
     cls = '',
     data = {}
   } = options;
@@ -9,7 +10,7 @@ export default function Footer(options = {}) {
   return Component({
     render: function render() {
       let middleContent = '';
-      middleContent += data.img ? `<img src="${data.img}" />` : '';
+      middleContent += data.img ? `<img src="${baseUrl}${data.img}" />` : '';
       middleContent += data.text ? `<p>${data.text}</p>` : '';
       middleContent = data.url ? `<a href="${data.url}" target="_blank">${middleContent}</a>` : middleContent;
       return `<div id=${this.getId()} class="o-footer relative flex row ${cls}">
--- a/src/controls/legend.js
+++ b/src/controls/legend.js
@@ -37,7 +37,7 @@ const Legend = function Legend(options = {}) {
 
   const addBackgroundButton = function addBackgroundButton(layer) {
     const styleName = layer.get('styleName') || 'default';
-    const icon = viewer.getStyle(styleName) ? viewer.getBaseUrl() + imageSource(viewer.getStyle(styleName)) : `${viewer.getBaseUrl()}img/png/farg.png`;
+    const icon = viewer.getStyle(styleName) ? `${viewer.getBaseUrl()}${imageSource(viewer.getStyle(styleName))}` : `${viewer.getBaseUrl()}img/png/farg.png`;
     backgroundLayerButtons.push(Button({
       icon,
       cls: 'round smallest border icon-small',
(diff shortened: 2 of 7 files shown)


Flodkvist:
My gut tells me there should be a more elegent solution to this problem but I had trouble finding it.

Also, my testing i based almost entirley on our configuration for gi.karlstad.se. So I´m not sure if these changes could have a negative effect on layertypes we´re not using.
Test map.
https://gi.karlstad.se/jnft/karlstadUppd/
Config.
http://gi.karlstad.se/jnft/karlstadUppd/origo/index.json

tonnyandersson:
For consistency, please use template literals for concatenation instead of `+` in _src/controls/legend.js_.

tonnyandersson:
@Flodkvist A [base tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base) might be a more elegant solution.

Works if you add the lines below to the `onInit` function of origo.js. However, I had to use the full `https://gi.karlstad.se..../origo/` baseUrl since I don't have the same file structure as you. Perhaps it won't work with just `./origo`. Not saying you should update this PR just yet. I'm just putting it up for discussion.

```
const base = document.createElement('base');
base.href = defaultConfig.baseUrl;
document.getElementsByTagName('head')[0].appendChild(base);
```

Not sure if it's the _best_ solution, though. There might be consequences...

Flodkvist:
That does look elegent.
I can give it a go and see what I run into.

Flodkvist:
@tonnyandersson 
I tested using your example and commented out all use of baseUrl ( https://github.com/Flodkvist/origo/tree/html-base-element )
The build is in use on https://gi.karlstad.se/jnft/karlstadBaseElement/

Seems to work great, and would leave a cleaner core.

I´m not sure if there are any drawbacks in needing to use a "complete" URL.
Or if it would have any impact on situations where one uses custom javascript styles.

tonnyandersson:
@Flodkvist You shouldn't have to use a "complete" URL. Relative URLs work just fine. But you would have to remove all existing mentions of `baseUrl` throughout the code to avoid double base URLs.

Not sure what's the better approach here. Personally, I like the idea of a "global" base URL that is automatically added to all URLs. That way you should never have to think about adding a base URL string when implementing new stuff. On the other hand, some may find it much easier to read and understand the code if base URLs are explicitly stated in the code.

Thoughts on this, anyone? @jokd, @daRasmussen, @johnnyblasta, @Grammostola, @hallstaelin

Flodkvist:
@tonnyandersson  Yes, quite right. No problem using relative URL. 
https://github.com/Flodkvist/origo/tree/html-base-element is cleaned up if the collective finds the base tag solution preferable.

I prefer the base tag solution, using baseUrl as before led to a bug that might have been there a long time without being noticed. Using the base tag seems safer for the feature.
I dont know if its positive in the big picture, but less lines of code and less files feels good.

jokd:
I vote for the base-tag

tonnyandersson:
Issue fixed with PR #1162.

PR #1101, opened 2021-01-18
Last stuff to drop jQuery

### LINKED ISSUE #489: Replace jQuery
Since Origo now is written in es6, we can now move on and replace jQuery as a dependency. To ahieve this, some replacements are required:
* DOM handling with a component library
* Handling of ajax requests
* Replace owl carousel, as it's a jQuery plugin

jokd:
Fixes #489


johnnyblasta:
Tested and LGTM. 
Had an issue with lmsearch-plugin not working, but after updating some packages it works again.

PR #1107, opened 2021-01-19
Autopan and search

### LINKED ISSUE #1106: autopan on small screens
Autopan along with the carousel works not so well on small screens because of wrong width setting in css.

jokd:
Fixes #1106


tonnyandersson:
@jokd Seems to do the trick.

PR #1115, opened 2021-01-20
Merge (assign) config options for default controls

### LINKED ISSUE #1109: Enable control config options for default controls
After opening up the possibility to let some config options be common to all controls (see `hideWhenEmbedded`, PR #1095) there is a use case for letting such options be available also for the default controls set in origo.js:

https://github.com/origo-map/origo/blob/003f4db9cd9ace5a74376174c46f4f3cff9fc39d/origo.js#L45
```
defaultControls: [
    { name: 'scaleline' },
    { name: 'zoom' },
    { name: 'rotate' },
    { name: 'attribution' },
    { name: 'fullscreen' }
]
```

Currently, any config options supplied for the default controls are ignored since the default and non-default control options are just concatenated in `loadresources.js`:
https://github.com/origo-map/origo/blob/003f4db9cd9ace5a74376174c46f4f3cff9fc39d/src/loadresources.js#L66
and
https://github.com/origo-map/origo/blob/003f4db9cd9ace5a74376174c46f4f3cff9fc39d/src/loadresources.js#L111

MattiasSp:
Fixes #1109.

Loops through the controls from the config and adds them to the array of default controls, unless the control is already among the defaults. If a control in the config is also among the default controls, they are merged.

This is to produce the following behavior:

- If, in the future, a default controls is defined with default options, then those options should be included
- If a default control is also defined in the config, the options in the config should be combined with the default options
- If the default definition and the config define the same option, the value in the config will overwrite the default value.

### DIFF
--- a/src/loadresources.js
+++ b/src/loadresources.js
@@ -62,10 +62,18 @@ const loadResources = async function loadResources(mapOptions, config) {
       }
       baseUrl = config.baseUrl || '';
       map.options = Object.assign(config, mapOptions);
+      map.options.controls = config.defaultControls || [];
       if (mapOptions.controls) {
-        map.options.controls = config.defaultControls.concat(mapOptions.controls);
-      } else {
-        map.options.controls = config.defaultControls;
+        mapOptions.controls.forEach((control) => {
+          const matchingControlIndex = map.options.controls.findIndex(
+            (defaultControl) => (defaultControl.name === control.name)
+          );
+          if (matchingControlIndex !== -1) {
+            Object.assign(map.options.controls[matchingControlIndex], control);
+          } else {
+            map.options.controls.push(control);
+          }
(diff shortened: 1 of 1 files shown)


MattiasSp:
An example build is published here: https://karta.haninge.se/devmaps/PR1115/

MattiasSp:
Do you think the identical "merging control arrays" section should be put into its own function, e.g. mergeControlArrays taking the default controls and the config controls as parameters?

jokd:
LGTM

tonnyandersson:
LGTM2.

@MattiasSp According to the _Really Important Rules of Coding_, those identical section should probably be put in a separate function. But I don't think it would have a noticeable impact on build size or performance, so in my opinion, you can leave it as it is.

PR #1116, opened 2021-01-21
Fixed broken legend animation

### LINKED ISSUE #1114: Legend group animation broken
**Describe the bug**
The legend group animation does not ~~work~~ animate when closing/contracting the legend group.

**To Reproduce**
1. Load a map with a legend group, preferably a group with several layers in it
2. Expand and contract the group and see the difference in the animations

**Expected behavior**
A smooth animation when expanding AND when contracting a legend group

**Screenshots**
legendanimation

 - OS: Windows
 - Browser: Chrome

tonnyandersson:
Fixes #1114.

Animation broke due to changes for the print control (#1037), which also affected the legend and any component using the `collapse-container` class.

### DIFF
--- a/scss/ui/_collapse.scss
+++ b/scss/ui/_collapse.scss
@@ -1,11 +1,4 @@
 .collapse-container {
-  height: 100%;
-  overflow: hidden;
-  position: relative;
-  transition: all 0.3s;
-}
-
-.collapse-container-no-height {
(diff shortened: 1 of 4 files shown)


johnnyblasta (approved):
Tested and both legend and print-control works as expected, so LGTM

Issue #1035, opened 2020-12-01
Labels of individual segments in measure makes it difficult to see


johnnyblasta:
**Is your feature request related to a problem? Please describe.**
When measuring a small object the labels of individual segments makes it difficult to see the measured object
![image](https://user-images.githubusercontent.com/48793206/100732969-ccaaf400-33cd-11eb-9297-1b86079f3ea4.png)

**Describe the solution you'd like**
A button to turn on/off labels of individual segments.

**Describe alternatives you've considered**
When measuring the labels hides the polygon if it is small and also can make it impossible to end measuring. Would it be possible to delay showing the labels until the measuring is ended?

tonnyandersson:
A few additional alternatives:

- Only show segment length when hovering the mouse over a specific segment
- Add a right-click context menu with options to turn labels on/off for individual segments
- Add a label collision avoidance system to move or hide colliding labels
- Use smaller labels
- Avoid measuring small polygons

PR #1127, opened 2021-01-22
Smaller labels .measure

johnnyblasta:
Less disturbing labels on the sections of the measuring tool.
Doesn't solve #1035 but makes it less annoying


johnnyblasta:
Example of new design of segments labels
image

jokd (approved):

PR #1128, opened 2021-01-22
Proposed workaround for missed scales when constrainResolution:true

### LINKED ISSUE #984: Shared map zoom level doesn't always match the shared map's
**Describe the bug**
When sharing a map the zoom level isn't always respected. Specifically with constrainResolution: false the opened shared map seems to interpret zoom 6.23932323 as 6.0. In terms of the OSM default background map that makes sure labels are always "crisp" even if the view isn't precisely what was shared. If constrainResolution: true then a shared zoom 18 is often interpreted as 17. This is true also if the start zoom level of the map (not shared) is set to 18.

**To Reproduce**
1. Latest Origo in debug
2. Set the map property constrainResolution to true
3. Zoom in via a scroll wheel or touch to the largest scale and share the map: 
4. Open it in another tab: it will likely show one zoom level out
2.1 Set the map property constrainResolution to false
3.1 Zoom in  or out a bit with the scroll wheel or - and + buttons, so that labels look slightly dim
4.1 Open in in another tab; labels will likely look sharp. The URL will include zoom=6.23231351341 and if that's changed to 6 and the page reloaded the map view doesn't change.

**Expected behavior**
For certain the constrainResolution: true shared map zoom level should be respected when opening
for constrainResolution: false then at least for WMS layers the current functioning can arguably be called a feature perhaps but I would personally feel better if 6.82323 was 6.82323

**Additional context**
An additional possibly related issue is that if constrainResolution is true: then zooming in and out with the plus and minus buttons enough will not respect the zoom levels given for the map, unlike when the map is zoomed via a mouse scroll wheel. In addition there seems to always be an extra zoom level out when zooming with the minus button as opposed to a scroll wheel. This level does not appear to adhere to the map resolutions. 

I have been unable to reproduce any of these issues in a release 1.x version of Origo (karta.sigtuna.se)

Grammostola:
Further partially fixes #984 (see the last comment there).

If constrainResolution:true for a map and the plus or minus zoom buttons are clicked repeatedly in quick succession the viewer will end up at a scale not defined in the list of resolutions given for the map. If the scale control is employed this will result in 1:529 for example instead of the preset 1:500 and the WMS background map will be a bit blurry. The map view including the scale will correct themselves if the map is panned but a user won't always do that before zooming in and out further (which doesn't generally help).

The issue seems to be because of the animate() method for the zoom buttons and its duration of 250msecs. There's a check and a cancellation of running animations that triggers if you zoom in repeatedly faster than the animation duration but it isn't helping in regards to this issue.  AdjustZoom(), even if always given a delta of +1 or -1, has issues when the map is max or min zoomed.

So this proposed fix is a workaround. The animation is still present (and works well) with scrollwheel zooming and it is never skipped if constrainResolution:false (which is default).

Accessing view.options_ is a linter error, a fix is on the way (a getConstrainResolution() https://github.com/openlayers/openlayers/blob/main/src/ol/View.js )

Before:
prefix

After:
postfix

### DIFF
--- a/src/controls/zoom.js
+++ b/src/controls/zoom.js
@@ -14,13 +14,17 @@ const Zoom = function Zoom(options = {}) {
   const zoomByDelta = function zoomByDelta(deltaValue) {
     const map = viewer.getMap();
     const view = map.getView();
-    if (view.getAnimating()) {
-      view.cancelAnimations();
+    if (map.getView().options_.constrainResolution === true) {
+      view.setZoom(view.getZoom() + deltaValue);
+    } else {
+      if (view.getAnimating()) {
+        view.cancelAnimations();
+      }
+      view.animate({
+        zoom: view.getZoom() + deltaValue,
+        duration
+      });
     }
-    view.animate({
-      zoom: view.getZoom() + deltaValue,
-      duration
-    });
   };
 
   return Component({
--- a/src/controls/zoom.js
+++ b/src/controls/zoom.js
@@ -14,7 +14,7 @@ const Zoom = function Zoom(options = {}) {
   const zoomByDelta = function zoomByDelta(deltaValue) {
     const map = viewer.getMap();
     const view = map.getView();
-    if (view.options_.constrainResolution === true) {
+    if (view.getConstrainResolution() === true) {
       view.setZoom(view.getZoom() + deltaValue);
     } else {
       if (view.getAnimating()) {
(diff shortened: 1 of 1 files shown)


jokd:
getConstrainResolution should be in the 6.5.0 and that will also solve the Travis fail.

jokd:
@Grammostola Will you push a fix using getConstrainResolution() instead?

PR #1132, opened 2021-01-25
Adjusts InputRange and styling to browser default

### LINKED ISSUE #1129: Make use of  InputRange component and new styling of range sliders

mulfvik:
Fixes #1129


jokd:
I liked the old look!

mulfvik:
Ok, does that include the value that renders next to the slider instead of the static values below the slider?

jokd:
That feature is cool but I like the thin blue bar more:
image
image

mulfvik:
Hmm, PR should give you this design. I´m not sure how it looks in mobile env.

Chrome
image
image

Edge
image
image

Firefox
image
image

jokd:
Chrome looks good but newer Edge is grayish

mulfvik:
What version is that? I run Microsoft Edge 44.17763.831.0

jokd:
Latest I think. Version 88, chromium

tonnyandersson:
I agree with @jokd. Not a fan of this look. Looks blunt in Chrome/Edge and absolutely horrendous in IE 11 (I know, we don't care about IE anymore, but still). The old one looked pretty much the same across all browsers; sleek, elegant, sexy.

Also, I don't like the shading on hover. Makes the control flicker in a cheap way.

Oh, and the texts are too large. The `text-smaller` class gives a cleaner look, I think.

And yes, I feel old and grumpy today.

mulfvik:
Ok, so if I read them expressions correctly from above the default designs are not in question. 
When updating to Edge Chromium, IE changed behavior aswell but if there is no more support for IE I dont bother with the new crazyness in IE

jokd:
Much better. If we use text-smaller class for the text in the slider and sliderlabel and removes the abstract-div unless the layer has an abstract I'd say it's a homerun.
image

mulfvik:
Nice, I just noticed, and missed earlier, the custom-size-control and the syntax for that one so I put `in progress` on this while I dig a litte bit deeper.

src/ui/input-range.js
```diff
@@ -31,7 +33,11 @@ export default function InputRange(options = {}) {
     },
     render() {
       return `
-      <input id="${this.getId()}" type="range" min="${minValue}" max="${maxValue}" value="${initialValue}" step="${step}" class="${cls}" style="${style}">
+      <div class="flex no-wrap text-small align-middle">
+        <input id="${this.getId()}" type="range" min="${minValue}" max="${maxValue}" value="${initialValue}" step="${step}" class="${cls}" style="${style}" oninput="this.nextElementSibling.value = this.value">
+        <output class="padding-left-small">${initialValue}</output><div>${unit}</div>
+      </div>
+      <div class="text-small text-align-center padding-smallpadding-top-smallest width-full">${label}</div>
```
tonnyandersson, line 40: Here be a typo, a missing space. Might wanna fix that.

mulfvik: Still grumpy I see 😉

tonnyandersson: Always.

jokd:
LGTM

jokd (approved):

PR #1135, opened 2021-01-26
Be able to put a style on Draganddrop control

### LINKED ISSUE #1134: Style DragAndDrop features
It would be nice to be able to configure a style for the features in DragAndDrop control.

A style object could be added in config with style rules for the features.

**Additional context**
Default style almost not visible feature:
image

Same feature with a config style:
image

johnnyblasta:
Closes #1134

### DIFF
--- a/src/controls/draganddrop.js
+++ b/src/controls/draganddrop.js
@@ -20,29 +20,51 @@ const DragAndDrop = function DragAndDrop(options = {}) {
       const groupName = options.groupName || 'egna-lager';
       const groupTitle = options.groupTitle || 'Egna lager';
       const featureStyles = options.featureStyles || {
-        stroke: {
-          color: [100, 149, 237, 1],
-          width: 4,
-          lineDash: null
-        },
-        fill: {
-          color: [100, 149, 237, 0.2]
-        },
-        circle: {
-          radius: 7,
-          stroke: {
-            color: [100, 149, 237, 1],
-            width: 2
-          },
-          fill: {
-            color: [255, 255, 255, 1]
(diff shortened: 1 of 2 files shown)


daRasmussen:
@johnnyblasta Very nice!

tonnyandersson:
@johnnyblasta Could this be made to work like [selectionStyles](https://github.com/origo-map/origo-documentation/blob/master/content/settings.md#example-featureinfooptions-with-overlay-as-infowindow), with individual styles for each geometry type? For example, if I want to use different stroke styles for lines and polygons. Now it seems like there's only one stroke style to rule them all.

johnnyblasta:
> @johnnyblasta Could this be made to work like [selectionStyles](https://github.com/origo-map/origo-documentation/blob/master/content/settings.md#example-featureinfooptions-with-overlay-as-infowindow), with individual styles for each geometry type? For example, if I want to use different stroke styles for lines and polygons. Now it seems like there's only one stroke style to rule them all.

Thought this was a simple and neat solution. Points gets a different style anyway. 
Don't have any good use cases, but some might want to be able to style their features on the fly also. Would be cool to also drop a style at the same time.

jokd:
I understand Tonnys opinion. It is also a bit confusing that we style stuff different ways in the config, perhaps we should try to unify the methods/syntax.

johnnyblasta:
Updated the style syntax to same as selectionStyles in featureinfo.
Also added new style type GeometryCollection which can be used by GeoJSON.

jokd (approved):

PR #1131, opened 2021-01-25
Clear tooltip on deactivate

johnnyblasta:
Fixes #1130

Issue #1130, opened 2021-01-25
Help tooltip on measure not erased on disable


johnnyblasta:
**Describe the bug**
The help tooltip on measure is not erased when disabling the tool.

**To Reproduce**
1. Activate tool and measure something
2. Deactivate the tool
3. The help text is left on the map at last position of cursor.

**Expected behavior**

**Screenshots**
![image](https://user-images.githubusercontent.com/48793206/105679653-0cceb480-5eef-11eb-9f71-4526c034b4bd.png)

PR #1140, opened 2021-01-27
Fix for when other tool is activated

johnnyblasta:
Updated fix for PR #1131


tonnyandersson:
@johnnyblasta What does this PR actually fix?

johnnyblasta:
> @johnnyblasta What does this PR actually fix?

This fixes an error that was introduced by the PR #1131 which should solve #1130. The problem was that an error occured in measure control when another tool was activated/deactivated. Should have made a new issue.

jokd:
In @johnnyblasta we trust. Approved.

jokd (approved):

PR #1146, opened 2021-01-28
port update

### LINKED ISSUE #1145: Ability to run start script and lint-dev script
It would be nice if it was possible to run both the start script and the lint-dev script together.

daRasmussen:
resolves #1145


tonnyandersson:
Not a fan of this autocorrect type lint-dev script, but if anyone wants to run it while running the start script, a different port seems to be the way to go. So... LGTM.

tonnyandersson (approved):

Issue #1072, opened 2021-01-05
Style, CSS & Sass-loader


daRasmussen:
add the following loaders to lint dev script.

* style-loader
* css-loader
* sass-loader

PR #1147, opened 2021-01-28
adds style loaders

daRasmussen:
solves #1072


tonnyandersson:
What is the purpose of these loaders? Why do we need them? Not that I am against it, I just don't know what the benefits are.

Besides that, the start script fails. Perhaps that's what's causing the Travis CI build to fail as well. Please check.

image

jokd:
Closed due to inactivity.

PR #1155, opened 2021-02-01
Enable filename option in print control

### LINKED ISSUE #1154: Print - Be able to set in config the name of the file that is created
As it is now the default name of the pdf or png that is created when printed is origo-map, it would be nice if that could be changed in the config for print control.

In the config a alternative for name should be added.

johnnyblasta:
Fixes #1154

### DIFF
--- a/src/controls/print/index.js
+++ b/src/controls/print/index.js
@@ -47,7 +47,7 @@ const Print = function Print(options = {}) {
         showScale,
         classes,
         defaultClass,
-        name: filename
+        filename
       });
       mapMenu = viewer.getControlByName('mapmenu');
       menuItem = mapMenu.MenuItem({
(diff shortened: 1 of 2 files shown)


MattiasSp:
Sweet! That was quick!

tonnyandersson:
Works fine! However, I think it would make more sense if the variable was named `filename` throughout, instead of changing name to `name`. Or is there a reason for the name change?

johnnyblasta:
> @johnnyblasta Works fine! However, I think it would make more sense if the variable was named `filename` throughout, instead of changing name to `name`. Or is there a reason for the name change?

The reason for having filename in the option was to avoid disambiguity about what name could refer to. Pure laziness that I didn't change the code that was already there for the rest.

johnnyblasta:
Updated PR so that it consistently using filename for variable.

jokd (approved):

PR #1159, opened 2021-02-04
Removes requirement for the position control

### LINKED ISSUE #1144: Measure control requires position control to work in touch mode
**Describe the bug**
When measuring on touch devices, the position control is used. However if this control is not activated, and added before the measure control in the config, measuring on touch devices is a no-go.

**To Reproduce**
1. Remove the position control or place it after the measure in the config
2. Try measuring on a touch device
3. The crosshair and the plus-button will not appear

**Expected behavior**
Either add the position control as a dependency or fall back to the old behaviour (opinions on that?)

 - OS: iOS
 - Browser Safari

jokd:
Fixes #1144 
Removes requirement for the position control

### DIFF
--- a/src/controls/measure.js
+++ b/src/controls/measure.js
@@ -8,7 +8,7 @@ import LineString from 'ol/geom/LineString';
 import Point from 'ol/geom/Point';
 import Projection from 'ol/proj/Projection';
 import * as Extent from 'ol/extent';
-import { Component, Element as El, Button, dom } from '../ui';
+import { Component, Icon, Element as El, Button, dom } from '../ui';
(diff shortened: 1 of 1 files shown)


tonnyandersson:
@jokd LGTM.

The only thing that bugs me is that you get two center position markers on top of each other if you activate the measure and position controls at the same time while in touch mode (only on large screen touch devices like Ipads where both controls are visible). It's not a big deal, though, and I realize that it's not worth the effort to fix it.

tonnyandersson (approved):

jokd:
Fully agree with you.

PR #1162, opened 2021-02-08
Alternative fix for baseUrl not effecting all resoures

### LINKED ISSUE #954: baseUrl option doesn´t affect all resources

Flodkvist:
Fixes #954

This is an alternative solution to PR #1098

Sets baseUrl to [base tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base). As suggested by @tonnyandersson in original PR.
Instead of baseUrl being applied to resources indivudually through out the code base.


tonnyandersson:
LGTM.

Issue #1119, opened 2021-01-21
Print - Set print settings' default values and visibility


MattiasSp:
We would like to be able to choose deafult values for all the settings in the print control, as well as controlling whether or not they are visible in the print settings menu.

It could be set as control options in the config.

johnnyblasta:
Some is able to set a default value, but you are right that it should be the same for all.

MattiasSp:
If all print settings could have their defaults set, it wouldn't be a long leap to be able to support multiple print settings templates.

Sets of default values could be named and supplied either as JSON objects in the config or as separate template files. Then a user could choose between supplied templates with a drop down list, or something similar.

What I'm after is to have the possibility to control the allowed print settings, to make it as convenient and easy as possible for users to print standardised map products.

MattiasSp:
Updated the issue three separate development steps that build on each other:

- setting default values
- controlling visibility/editability
- using named sets as print templates

MattiasSp:
PR incoming for part one of this issue.

asemoller:
Maybe we could pin this issue as well?

PR #1172, opened 2021-02-25
Print - Expose all settings as config options

MattiasSp:
The first PR relating to #1119. Also related to #785.

All current print settings are exposed as config options, letting their default values (and in some cases which values should be allowed) be controlled in the config.

Two maps for testing:

- [Custom values for all new print settings](https://karta.haninge.se/devmaps/PR-print-expose-all-settings-as-config-options/all.html)
- [All defaults (no config options)](https://karta.haninge.se/devmaps/PR-print-expose-all-settings-as-config-options/none.html)

I made the choice to move the default values for most variables out of their respective print controls, the [print settings](https://github.com/origo-map/origo/blob/master/src/controls/print/print-settings.js) and the [print component](https://github.com/origo-map/origo/blob/master/src/controls/print/print-component.js) and put them in the [index](https://github.com/origo-map/origo/blob/master/src/controls/print/index.js) file. That way all the defaults can be seen and changed in one place instead.

Additionally, the rotation of the map is now reset after closing the print control, so that rotating a map for printing doesn't affect the regular map.


MattiasSp:
An example config with all the new options:

```
{
  "name": "print",
  "options": {
    "headerText": "print-expose-all-settings-as-config-options",
    "headerPlaceholderText": "Här kan du skriva din rubrik",
    "headerAlignment": "right",
    "headerSizes": ["h1", "h2", "h3", "h4", "h5"],
    "headerSize": "h5",
    "headerFormatIsVisible": true,
    "descriptionText": "Karta med alla utskriftsinställningar ändrade.",
    "descriptionPlaceholderText": "Här kan du skriva din beskrivning",
    "descriptionAlignment": "right",
    "descriptionSizes": ["h3", "h4", "h5", "h6"],
    "descriptionSize": "h6",
    "descriptionFormatIsVisible": true,
    "sizes": {
      "a2": [594, 420],
      "a3": [420, 297],
      "a4": [297, 210],
      "a5": [210, 148],
      "custom": [100, 100]
    },
    "sizeInitial": "a5",
    "sizeCustomMinHeight": 20,
    "sizeCustomMaxHeight": 1189,
    "sizeCustomMinWidth": 20,
    "sizeCustomMaxWidth": 841,
    "orientation": "landscape",
    "resolutions": [ {"label": "Låg", "value": 72}, {"label": "Standard", "value": 90.7}, {"label": "Mellan", "value": 150}, {"label": "Hög", "value": 300} ],
    "resolution": 90.7,
    "scales": ["1:250 000", "1:100 000", "1:50 000", "1:25 000", "1:10 000", "1:5000", "1:2500", "1:1 000", "1:500", "1:400"],
    "scaleInitial": "1:100 000",
    "showMargins": false,
    "showCreated": false,
    "createdPrefix": "Skapad ",
    "showScale": false,
    "showNorthArrow": false,
    "rotation": 180,
    "rotationStep": 30
  }
},
```

jokd:
Works well except resolution integers cast to strings and therefore the initial state is not active:
image

Besides that, LGTM

MattiasSp:
Thanks @jokd, good catch! :)

jokd (approved):

PR #1173, opened 2021-02-26
Fix when rotation is disabled

### LINKED ISSUE #1168: Print - Hide rotation setting when rotation is disabled
**Describe the bug**
The rotation of the print can be set (with no effect) even when the map has disabled rotation.

**To Reproduce**
1. set enableRotation to false
2. Print setting still has the rotation option

jokd:
Fixes #1168
Hides the rotation slider when enableRotation: false in config (An Openlayers setting)


tonnyandersson:
Seriously weird way to check if rotation is enabled, but I sure can't find a better solution. LGTM.

tonnyandersson (approved):

jokd:
I can't disagree.

PR #1175, opened 2021-02-26
Fix for iPhone

### LINKED ISSUE #1169: Print - Blank prints in iOS
**Describe the bug**
When printing in iOS the map isn't always shown. Escepcially the png export. pdf often works on the second attempt.

**To Reproduce**
1. Open print
2. Save as png
3. Nothing
4. Save as pdf
5. Nothing
6. Save as pdf again
7. Maybe correct layout

**Expected behavior**

 - Device: iPhone 8+
 - OS: iOS 14
 - Browser: Safari and Chrome

jokd:
Fixes #1169 
Quick and dirty.

### DIFF
--- a/src/utils/download.js
+++ b/src/utils/download.js
@@ -149,7 +149,8 @@ export const printToScalePDF = async function printToScalePDF({
   const pdf = new jsPDF({ orientation, format, unit: 'mm', compress: true });
   const styleAttributes = el.getAttribute('style');
   el.setAttribute('style', styleAttributes.split('transform: scale')[0]); // Remove scaling to get correct print size of image
-  const image = await dom2image(el, exportOptions);
+  let image = await dom2image(el, exportOptions);
+  image = await dom2image(el, exportOptions); // Fix for iPhone
   pdf.addImage(image, 'JPEG', 0, 0, pdf.internal.pageSize.getWidth(), pdf.internal.pageSize.getHeight());
   el.setAttribute('style', styleAttributes); // Restore scaling
   pdf.save(`${filename}.pdf`);


tonnyandersson:
Now that may be one of the weirdest fixes I've ever seen. _"A variable so nice, they set it twice"_. Is that kinda like hitting Ctrl-C twice (or more) just to make sure it sticks?

All jokes aside, it is not a pretty solution. It seems to do the trick, but I feel there's something else at play here. Perhaps something async/await related?

jokd:
Actually I think it's related to rendering the pdf from the memory, in which iPhone doesn't store it correctly until after the first try.
It's hard to debug.

tonnyandersson:
Apparently, there's an open issue regarding this in the dom-to-image repo, issue 343. Perhaps you've read it? Not linking because of Github "mentions" shenanigans.

jokd:
Haven't seen but it's obviously a known bug. The solution seems to be switching to html2canvas if this supersweet fix leaves a bad taste in your mouth.
Another reason would be that dom2image hasn't been updated in years.

tonnyandersson:
Well, I can live with this fix since it seems the _actual_ issue is out of our hands. But perhaps switching to html2canvas would be a better solution. On the other hand, we're currently using dom-to-image-_more_, a more active fork of the original dom2image. This issue still remains, though.

jokd:
html2canvas is already a dependecy and available in download.js so I wonder why we are not using it, any downsides?
Otherwise I think we shall use that instead if we get it kicking.

tonnyandersson:
Not particularly well-versed in either library, so I don't know what the main differences are. But it seems html2canvas might be a bit more unpredictable and crappy:

https://stackoverflow.com/questions/43755750/div-or-html-to-image-alternative-to-html2canvas

jokd:
@johnnyblasta is probably da man to answer that question^^

johnnyblasta:
> @johnnyblasta is probably da man to answer that question^^

As far as I recall the dom2image had to be used to get it to work with setting resolutions on the printed map.
I see that there is a another fork called dom-to-image-improved perhaps this is better?

jokd:
There are dozens of forks and Im testing them all. So far no success.

jokd:
Tested the other forks, I had especially high hopes on the dom-to-image-even-more-with-world-domination, but none of them have solved this issue. So I suggest we go with this beautiful fix and are on the lookout for alternatives in the future.
The html2canvas dependency must be possible to replace but thats another issue.

tonnyandersson:
Well, if dom-to-image-even-more-with-world-domination can't fix it, it can't be fixed. I'll allow this abomination. For now.

tonnyandersson (approved):

PR #1180, opened 2021-03-01
Added Bookmarks control

### LINKED ISSUE #1171: Bookmarks control
A bookmarks control could be useful. Essentially a list of pre-configured named "views", i.e. a list of preset center coordinates.

A dropdown list with presets is probably what first comes to mind. This is, however, not really an optimal solution since the UI has become quite crowded, especially on smartphones.

Instead, I suggest a different approach, with a detached, movable list element. Neatly tucked away when not in use, but only a click away when needed.

bookmarks

tonnyandersson:
Fixes #1171.

Config example:
```
{
  "name": "bookmarks",
    "options": {
      "items": [
        {
          "name": "Hudiksvall",
          "coordinates": [1904063, 8794995]
        },
        {
          "name": "Karlstad",
          "coordinates": [1503136, 8262205]
        },
        {
          "name": "Sigtuna",
          "coordinates": [1986473, 8315061],
          "zoomLevel": 10
        }
      ]
    }
  }
```

Other useful options:

| Property | Data type | Description |
|---|---|---|
| `isActive` | boolean | Sets whether the bookmarks control should be active from start. Default is `false`. |
| `maxZoom` | integer | Sets the general zoom level. This zoom level is used if no zoom level is set on the bookmark. Default is `15` (or highest available zoom level). |
| `duration` | integer | Sets the duration of the view animation to the bookmark. A value of `0` gives no animation. Default is `300`. |
| `title` | string | Sets the title of the control. Also sets the tooltip text. Default is `Bokmärken`. |


jokd:
Works well on both desktop and mobile, really nice. 
How about using the same layout as the modal with header, size of close button etc? That will be more consistent but perhaps it gets a bit big?
image

tonnyandersson:
How about using the same layout as the Bookmarks control (and featureinfo popup) instead? It looks about 63% better, according to recent studies.

jokd:
I can go with that one as well and you seems to have done your homework.
And while we are at it, the border and its radius seems to be a bit inconsistent:
image
Not saying which one is the better...

tonnyandersson:
The bookmarks control uses the same border radius as mapmenu and legend and probably a bunch of other stuff as well, since the radius is set by the `.control` css class. The popup is running its own `.popup` game.

Personally, I like the borderless style of the bookmarks/mapmenu/legend controls better. Regarding the radius, it's a toss-up.

jokd:
You know what? I don't think anyone will bother whatever way we go so I'll rest my face and approve this and you do what you do and it'll be awesome.

jokd (approved):

tonnyandersson:
Haha, right on!

But you are right, though. There are inconsistencies in the UI that probably should be addressed at some point. We just need to pick a standard (since there are several) and use it across the board. Or perhaps a few standards. You wouldn't want the mapmenu and the legend to look _exactly_ the same, for example.

PR #1181, opened 2021-03-02
Print - toggle between map and print layout interactions

### LINKED ISSUE #1157: Print - Use new button in print control to toggle map interaction
Having map interaction always on in the print control creates a number of issues from a UX perspective:

- It's possible to pan to the middle of the print layout, where there is nothing to grab to pan back out to the sides
- Pinch zooming in and out of the layout can be a bit tricky when only the margins and logos are "grabbable" (and pinch zooming in the layout is not currently possible either)
- After changing map settings and/or panning the map, when a user tries to preview the print, the user could easily pan or zoom in the map by mistake

Locking the user interactions to either the map view or the layout/page, one at a time, and toggling between the two states with a new button.

We suggest locking the print preview to layout/page interactions by default, and using a button similar to that used in e g QGIS Desktop for activating map interactions: 
image

Using pan- and zoom buttons for map interaction and pinch (or mouse scroll) and grab for layout is another alternative, but using pinch and grab interactions for both seems like a better solution.

Using e g 3 point interactions for the map and 1-2 point interactions for the layout could also work, but is less obvious and may not be supported by all devices.

MattiasSp:
Fixes #1157. Also part of #785.

- Adds a new `o-print-tools-left` div section for print tools/buttons in the upper left.
- Puts the current settings (Collapse) control in the o-print-tools-left section
- Adds a second button (Button) control to the o-print-tools-left section that toggles between "no map interactions" and default map interactions
  - Including a new config option `mapInteractionsActive` to set whether map interactions should be enabled by default (true) or not. Default is false.

I'd happily take suggestions for the button icon (now using `#fa-pencil-square`).

I made some changes to css classes to accommodate the new `o-print-tools-left` div and the extra button. I also removed the class `width-20` from the print-settings.template.js which fixes an issue where the print settings collapsible was too wide for some mobile screens. It would be great of I could get some help with testing on different screens and devices though :)


MattiasSp:
Here's a map for testing the behaviour: https://karta.haninge.se/devmaps/pr-print-map-vs-layout-interaction-toggle/

mulfvik:
Nice work! The icon for the button is used in the editor for editing data, maybe it can be replaced with an map icon? Map on/off

jokd:
Agree with mulfvik, well done and using a map icon would probably be better. 
Works well on both desktop and my iPhone 8+!

MattiasSp:
FYI the icon `fa-pencil` was removed because it was a duplicate, so it's not removed completely.

tonnyandersson:
I suggest using the Material Icons map icon instead (`ic_map_24px`), since most of the icons in the UI are from MI. But it seems it's not in the `material-icons.svg` sprite yet, so you will have to copy it from `css\svg\vendors\material-sprites\svg-sprite-maps-symbol.svg`.

MattiasSp:
Based on a user suggestion, perhaps I should change the tooltip text to something less technical, e g "Ändra kartans läge" instead of "Aktivera kartvyn". What do you think?

jokd:
Sounds good!

jokd (approved):

PR #1186, opened 2021-03-04
Update set-scale-control.js

### LINKED ISSUE #1182: Print - Set default scale
**Describe the bug**
You are not able to set the resolution of the print unless the scale is set (which it isn't initially). Therefor it makes sense to default the scale to something when activating the print control, eg the one closest to the present map.

**To Reproduce**
1. Open print
2. Try to change resolution

**Expected behavior**
Change resolution without picking scale first.

 - OS: Windows
 - Browser: Edge
 - Version 88

jokd:
Sets default scale to ~same resolution as the map
Fixes #1182

### DIFF
--- a/src/controls/print/print-component.js
+++ b/src/controls/print/print-component.js
@@ -356,9 +356,6 @@ const PrintComponent = function PrintComponent(options = {}) {
       pageContainerElement.style.height = orientation === 'portrait' ? `${sizes[size][0]}mm` : `${sizes[size][1]}mm`;
       pageContainerElement.style.width = orientation === 'portrait' ? `${sizes[size][1]}mm` : `${sizes[size][0]}mm`;
       this.updateMapSize();
-      if (printScale > 0) {
-        this.changeScale({ scale: printScale });
-      }
     },
     removeViewerControls() {
       const controls = map.getControls().getArray();
--- a/src/controls/print/print-component.js
+++ b/src/controls/print/print-component.js
@@ -339,6 +339,7 @@ const PrintComponent = function PrintComponent(options = {}) {
       });
     },
     onRender() {
+      printScale = 0;
       today = new Date(Date.now());
       viewerMapTarget = map.getTarget();
       pageContainerElement = document.getElementById(pageContainerId);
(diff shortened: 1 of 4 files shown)


tonnyandersson:
Setting the default scale seems to work fine. But I noticed that the print scale picker doesn't change when zooming in the print preview map. Is this a bug or a feature?

Please note that this issue is present in the master branch as well.

print_zoom_scale

johnnyblasta:
> Setting the default scale seems to work fine. But I noticed that the print scale picker doesn't change when zooming in the print preview map. Is this a bug or a feature?
> 
> Please note that this issue is present in the master branch as well.
> 
> print_zoom_scale

Something to consider when implementing that is the case when the zoom levels in the map and those defined in the print control doesn't match. What should be shown in scale picker then?

jokd:
That’s another issue I’d say. This is just to get the resolution picker working without having to select scale first. 
Like Johnny says, there are things to consider when changing behavior of the scale picker.

jokd:
So yeah it’s a feature^^

tonnyandersson:
Fair enough.

But there's a big boo-boo right around line 62:

image

jokd:
I shouldn’t have edited in word.

tonnyandersson:
LOL! Word... You should use Notepad instead. Superior tool, so much better.

jokd:
Just kidding, stole it from SO.
Notepad is great though.

tonnyandersson:
Found a tiny, tiny issue. To replicate:

1. Load the map and zoom to a scale of your liking, let's call it Scale A
2. Open the print control and open the settings tab, the scale picker now shows Scale A (or closest available)
3. Now close the print control and zoom to a completely different scale, henceforth called Scale B
4. Open the print control once again and gasp at the horrific fact that the scale picker still shows Scale A

Bug or feature?

jokd:
Clearly a bug, will look into it next week

jokd:
There are still a bit funny behaviour when using the high and low dpi settings because when changing the resolution in the print settings you are actually zooming in or out in the map while scaling the print. When the print window is closed, the map has zoomed in (when using 300dpi) or out (when using 75dpi). If then activating print mode again, the map has stepped 1 zoom level in or out. 
This is however another issue.

tonnyandersson:
...aaand now the map scale doesn't match the scale picker scale:

image

jokd:
Yes it does unless you zoom ( or pan in a web mercator map). The scale is not updating as mentioned earlier in this thread

jokd:
You are far up north. around Berlins latitude it works great.
Jokes aside, constrainresolution will have this effect, especially when using web mercator.

tonnyandersson (approved):

jokd:
Agree but at least it works well when constrainResolution is false and using sweref.
Havent tested when setting scale in the print settings config though....

PR #1189, opened 2021-03-08
Update editortoolbar.js

jokd:
Necessary for https://github.com/origo-map/draw-plugin/pull/29


jokd:
I fix this in the draw plugin instead so there is less chance other plugins might break.

PR #1191, opened 2021-03-08
Image carousel for popup

### LINKED ISSUE #1190: Image carousel for feature popup
From a feature that has a attribute containing links to images set that they should be displayed in a image carousel in the feature info popup.

ImageCarousel

Configurable if the image carousel should be started automatically.

johnnyblasta:
Closes #1190 

There is a little bug for when both image carousel and feature carousel is enabled. The images are doubled, see example where there are only 3 images, there is 6 buttons so it is repeated.
image


johnnyblasta:
Added a frame around the images, to make it a bit nicer and avoid having the images bleeding over between frames.

ImageCarouselNew

jokd:
Do you have an example config/map to test with?

johnnyblasta:
> Do you have an example config/map to test with?

Here is a map you can test:
[https://kartatest.sundsvall.se/test/carousel/]

jokd:
It is this class that messes it up below the 768px-breakpoint:
image

width should probably not be set to 100%

johnnyblasta:
> It is this class that messes it up below the 768px-breakpoint:
> image
> 
> width should probably not be set to 100%

Great that did the trick.

There is also a problem with wandering slides when the screen is really small about below 350px.
Also a scrollbar sometimes appear unnecessarily.
ImageCarouselSmall

jokd:
That is probably also because of the width is set, to 88% in that case.
I also noted that the popup is higher than it seems, which prevents map interactions above the popup:
image

johnnyblasta:
I removed the width and then it works for really small screens and large screens, but not for the segment between 320px to 768px.

jokd:
Have you tested it in IE, the "overflow-y: auto;" was recently added to make IE work with glide.

johnnyblasta:
> Have you tested it in IE, the "overflow-y: auto;" was recently added to make IE work with glide.

Ofcourse IE couldn't handle it, so made some corrections

jokd:
I still have issues in Edge, the scrollbar is present and the #o-popup is much higher than the content div as pasted above.

johnnyblasta:
Fixed that the #o-popup is much higher than the content div.

Don't know what to do with the scrollbar, because it's needed to have `overflow-y:scroll;` for the scrolling of content to work, but when the image-carousel is present the scrollbar is shown even though it's not needed.

jokd:
Might be just some padding or similar to get it a pixel off. overflow-y:auto will that do it?

johnnyblasta:
> Might be just some padding or similar to get it a pixel off. overflow-y:auto will that do it?

No go on overflow-y:auto because then the slides starts wandering.

johnnyblasta:
No scrollbar if the image carousel is followed by other attributes, maybe a clue there?

ImageCarouselTop

jokd:
Tested and when emulating an iPhone, using the arrows will skip a step and create a warning in the console. It can also stop between two images when touchsliding, perhaps its related?:
image

The popup still gets the wrong height (sometimes):
image

And there could be a little bit more margin below the dots if I am to be picky.

johnnyblasta:
> Tested and when emulating an iPhone, using the arrows will skip a step and create a warning in the console. It can also stop between two images when touchsliding, perhaps its related?:
> 
> The popup still gets the wrong height (sometimes):
> 
> And there could be a little bit more margin below the dots if I am to be picky.

The warning when skiping has been present previousley when the glide has been used for the infowindow so that is nothing new.

I can't provocate a case when the popup height is wrong, so very hard to solve.

Perhaphs it's because of the dots being slightly cut the scrollbar wants to appear. I'll look into it.

jokd:
The warning may be the same but it jumps two steps when using the arrow buttons in touch mode and at least that is new. Perhaps has to do with the autoplay function?

johnnyblasta:
Changed so that image height is fixed and lets the image fit inside. Should sort out the problem with popup height is to high sometimes, but hard to be definitive since it happen sparingly.
ImageCarouselFit

jokd:
@johnnyblasta Is this ready for testing?

johnnyblasta:
> @johnnyblasta Is this ready for testing?

I made some changes realising that if you get two features with a image carousel they should have unique id to not interfer with each other. It didn't resolve the problem with double bullets and also there isn't a carousel for the second feature.

I would be greatful if somone could take a look at it and see what I'm missing.

jokd:
@johnnyblasta I think I know what is going on and will have a look

jokd:
@johnnyblasta I pushed a fix, please have a look

johnnyblasta:
> @johnnyblasta I pushed a fix, please have a look

It seems to be working great, thanks!

jokd:
Perfect then there is just the styling left. Ping @tonnyandersson the ux 🥷

PR #1193, opened 2021-03-09
Splitter fix

### LINKED ISSUE #1187: Be able to split attributes in featureinfo
If a attribute contains a delimited string of one or many of the same attribute it would be nice to be able to split them. It makes most sense to use this when dealing with url or img attributes.

In the attributes for a layer there should be able to specify that the attribute should be split by some character
`"splitter": ","`

johnnyblasta:
Does this fixes #1187 and not causing a bug in other cases? @jokd 
The bug was caused by me by creating a function for 2 almost identical pieces of code snippets and missing the subtle difference.

### DIFF
--- a/scss/_featureinfo.scss
+++ b/scss/_featureinfo.scss
@@ -7,3 +7,12 @@
   max-height: calc(70vh - 46px);
   overflow-y: auto;
 }
+
+.o-image-content {
+  width: 240px;
+}
+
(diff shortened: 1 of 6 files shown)


jokd:
That was one sneaky way of slipping in new functionality!
The popup looks well and the carousel seems to work as expected.

johnnyblasta:
New try with PR for this #1195

Issue #969, opened 2020-06-26
Sharing map with features from certain layer types selected not working


tonnyandersson:
**Describe the bug**
Sharing a map with group layer features selected triggers an error and the map crashes when loading the shared map.

**To Reproduce**
1. Create a map with a group layer (layer type "GROUP")
2. Load the map and select a feature in the group layer
3. Share the map
4. Open the shared map in a new tab/window

**Expected behavior**
The map should load without issues, zooming to the selected feature.

tonnyandersson:
Sharing a map with multiple layers with the same layer name also produces a similar result. I.e. multiple layers from the same source layer, specified with double underscore suffixes in the config file. Example:

```
{
  "name": "mylayer__1",
  "title": "My Layer 1",
  "source": "myDataSource"
  ...
},
{
  "name": "mylayer__2",
  "title": "My Layer 2",
  "source": "myDataSource"
  ...
}
```

Grammostola:
I've had a look and think I've a prelim solution for group layers (I haven't tested all kinds of trees but a few basic ones with grouplayers in grouplayers seem ok). 

 "__"-wfs-layers are a different species of fish, as you mentioned. (I wasn't able to repro that issue with a local geojson layer).  Possibly there's a narrow path that needs to be found when handling layer names and feature ids and viewer and probably permalink seem involved. Keep them fishes as one issue?

I'll continue digging anyways.

tonnyandersson:
Yeah, I believe it's all connected. How layer names are handled, how they are split and transformed into id:s, settin' and gettin' permalinks and selecting items for feature info stuff... It's all connected.

I guess layer groups are not much of an issue if you stick to unique group and layer names. In such cases, all you have to do is to check for group layers and then dive into them to find the layers within, or something like that. However, problems will arise if you use the same naming scheme as in the [release test config](https://www.malardalskartan.se/origo/v2.2.0/test.json), e.g.:

```
{
  "name": "layername",
  "type": "GROUP",
  "layers": [
    {
      "name": "layername__1",
      "type": "WFS",
      ...
    },
    {
      "name": "layername__2",
      "type": "WFS",
      ...
    }
  ]
}
```

I know, this is essentially an _underscoreunderscore_ issue rather than a group layer issue. But I think this is a not so uncommon scenario, grouping several filtered layers from the same source layer in the same group.

Perhaps we should treat them as separate issues, just to get this group layer issue out of the way. But that feels like half-assing it.

[...]

PR #1200, opened 2021-03-10
Make getLayers aware of layers in groups

### LINKED ISSUE #1197: WMS GetFeatureInfo not working for layers in a group
**Describe the bug**
WMS GetFeatureInfo not working for layers in a group.

**To Reproduce**
Make a group and add a queryable WMS layer in it. (Geoserver layer used for test)

Editor example can be used for fast testing, change editor_polygon layer type from WFS to WMS and add it in a group.
Close editor and try infocklicking results in error.

**Expected behavior**
Should work the same as a standalone WMS layer featureinfo click.

**Screenshot devtools**
probs_wms_grouplayer

jokd:
Fixes #1197 
The getLayers function in viewer will instead of returning a layerGroup return the individual layers in the group. I have not thoroughly tested how this affect other tools but I am not aware of any that uses the getLayers in that way.

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -161,19 +161,7 @@ const Viewer = function Viewer(targetOption, options = {}) {
 
   const getMap = () => map;
 
-  const getLayers = function getLayers() {
-    const arr = [];
-    const layerArray = map.getLayers().getArray();
-    layerArray.forEach((element) => {
-      arr.push(element);
-      if (element.get('type') === 'GROUP') {
-        element.getLayers().getArray().forEach((layerInGroup) => {
-          arr.push(layerInGroup);
-        });
-      }
-    });
-    return arr;
-  };
+  const getLayers = () => map.getLayers().getArray();
 
   const getLayersByProperty = function getLayersByProperty(key, val, byName) {
     const layers = map.getLayers().getArray().filter(layer => layer.get(key) && layer.get(key) === val);
@@ -184,7 +172,17 @@ const Viewer = function Viewer(targetOption, options = {}) {
     return layers;
   };
 
-  const getLayer = layerName => getLayers().filter(layer => layer.get('name') === layerName)[0];
+  const getLayer = function getLayer(layerName) {
+    const layerArray = getLayers();
+    if (layerArray.some(layer => layer.get('name') === layerName)) {
+      return layerArray.find(layer => layer.get('name') === layerName);
+    } else if (layerArray.some(layer => layer.get('type') === 'GROUP')) {
+      const groupLayerArray = layerArray.filter(layer => layer.get('type') === 'GROUP');
+      const layersFromGroupLayersArray = groupLayerArray.map(groupLayer => groupLayer.getLayers().getArray());
+      return layersFromGroupLayersArray.flat().find(layer => layer.get('name') === layerName);
+    }
+    return undefined;
+  };
 
   const getQueryableLayers = function getQueryableLayers() {
     const queryableLayers = getLayers().filter(layer => layer.get('queryable') && layer.getVisible());
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -161,7 +161,20 @@ const Viewer = function Viewer(targetOption, options = {}) {
 
   const getMap = () => map;
 
-  const getLayers = () => map.getLayers().getArray();
+  const getLayers = function getLayers() {
+    const arr = [];
+    const layerArray = map.getLayers().getArray();
+    layerArray.forEach((element) => {
+      if (element.get('layerType') === 'group') {
+        element.getLayers().getArray().forEach((layerInGroup) => {
+          arr.push(layerInGroup);
+        });
+      } else {
+        arr.push(element);
+      }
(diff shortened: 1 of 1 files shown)


Grammostola:
Hi, I'll check how this affects https://github.com/origo-map/origo/tree/share-selection-in-grouplayer-fix (a changed getLayer() ) before initiating a PR for it. It's one of two planned for #969

tonnyandersson:
I think you need to make two different getLayer functions, or at least pass an argument that states if groups should be split open on not. Different contexts need different results, as I'm sure you've noticed.

jokd:
Tough one. We have the getGroupLayers function so I think it makes sense not to include grouplayers in getLayers but thats not how its been used and will probably take some effort to get things straight.

tonnyandersson:
Actually, it does make sense to include group layers in getLayers when the layers in the group need to be treated as one, e.g. for the legend, some permalink stuff (visible/not visible) etc. But when feature access is crucial, the grouped layers must be treated as separate layers.

I think an argument would takes us quite far, like:

```javascript
const getLayers = function getLayers(explodeGroup) {
  const arr = [];
  const layerArray = map.getLayers().getArray();
  layerArray.forEach((element) => {
    if (element.get('type') === 'GROUP' && explodeGroup) {
      element.getLayers().getArray().forEach((layerInGroup) => {
        arr.push(layerInGroup);
      });
    } else {
      arr.push(element);
    }
  });
  return arr;
}
```

jokd:
Well that might work but some functions will need to be rewritten. But on the other hand we will gain some functionality. Make layers in group searchable for instance.

jokd:
However! the getLayer function will probably want both the group and the individual layers since it is filtering on layer name.

tonnyandersson:
Depends on the situation, I'd say.

Grammostola:
I'm merely a little lost so will venture : is getLayer ever expected to return a group layer object? One of two problems described in 969 was that getLayer() filtered the top layer of the array, with potential layer groups when groups inside of layer groups were wanted. If Tonny's suggestion above is implemented then my tweak of getLayer isn't necessary and getLayer can merely summon getLayers with a truthy value for expose group contents.  Unless getLayer is somewhere expected to return a group layer object.

(I suppose there's no point to layers in group layers in group layers)

jokd:
getLayer is used for both group layers and the individual layers so your tweak is not unnecessary, unless we make getLayers return both group layers its contained layers

tonnyandersson:
I think the main changes should be made to `getLayers`, since that's the "base" function for getting layers. `getLayer` uses it, other functions goes straight for `getLayers`, and so on.

But then there's a few bad guys going straight for the `map.getLayers()` function...

jokd:
Yeah but getLayer must be able to find both group layers and the layers it consists of so somehow must be rewritten depending on how we handle getLayers

tonnyandersson:
Yeah, perhaps by passing the explode/expose argument when necessary?

```javascript
const getLayer = (layerName, explodeGroups) => getLayers(explodeGroups).filter(layer => layer.get('name') === layerName)[0];
```

jokd:
You are not always aware whether it is a group layer or not you are looking for so it needs 'em all

tonnyandersson:
Perhaps you need to make yourself aware. Since group layers and regular layers can't be treated the same way in any given situation (not the same methods or properties etc.), you can't just get them all at once and expect to be able to process them the same way. You would have to implement all kinds of checks and safeguards every time you call `getLayers()`.

Take a look at your latest commit, for example. There you're getting all layers - regular layers, group layers, grouped layers - in a mixed bag, and it messes the legend up gloriously.

jokd:
I am talking about getLayer

tonnyandersson:
Oh, of course. That would work. As long as you don't have group layers and grouped layers with the same name, but that's a minor issue. But I guess you'd still have to check for group layers in `getLayers` and possibly handle them differently depending on the situation.

jokd:
Maybe @Grammostola idea is the way to go after all, leave getLayers as is and let getLayer find layers contained in a group as well as ordinary layers.

Grammostola:
If getFeatureInfo uses getLayer (for this issue) and if everything calling getLayer wants to find a layer obj my suggestion or I guess Tonny's explode above; if there are pieces of Origo that are calling getLayer and expecting a groupLayer obj or groupLayer obj / layer obj then it can't as it is, If I've understood things correctly.

jokd:
@Grammostola I would say most (all?) cases are covered in your solution; first if there's a layer with the name it is returned, if not, check the group layers if any of them contain the requested layer and in that case return that layer, not the group layer.
Not touching the getLayers function is probably wise after all.
I pushed your suggestion which partly fixes #969 as well (haven't tested the __-layers)

tonnyandersson:
Works in most cases. Sharing a map with group layer objects selected still doesn't work with the release test config, but that's because of layer names and __-layers.

Grammostola:
Alright : ) No, __ layers are special and there'll be some cheap trickery in the branch I made for that shortly.

PR #1202, opened 2021-03-11
Share wfs layer selection amelioration

### LINKED ISSUE #969: Sharing map with features from certain layer types selected not working

Grammostola:
Fixes #969  (well it's something to talk about and limited in scope, see below)

 It targets wfs layers with the "__" naming scheme as shown in the origo layer documentation.

WFS layers because I have been unable to reproduce the issue with the default geojson layer, as mentioned in the related issue. WMS layers don't get to have selections with info popups in shared maps and other layers would need to be tested, but the suggested changes should only affect WFS layers because it seemed they primarily are treated special and are mentioned in the docs  (("__") I mean. Well, besides WMS layers but as mentioned they don't count)

It seems to work when I define the same wfs layer twice in a group layer, suffixing the names with "__rust" and "__11"


jokd:
Works well as far as I can see!

jokd (approved):

PR #1205, opened 2021-03-15
Round buttons in print control

### LINKED ISSUE #1184: Round buttons for print control
The print control should have round buttons like the rest of Origo. (settings that is, and the upcoming map-pan-toggle-button)

jokd:
Fixes #1184


johnnyblasta (approved):
Test and LGTM

Issue #1223, opened 2021-03-30
Need a copy tool


steff-o:
**Is your feature request related to a problem? Please describe.**
It should be easier to copy a geometry from another layer when editing. Today you have to draw it manually and use snapping. This function would be useful for drag-and-drop layers to enable users to add features to a layer from external sources. It can also be useful for creating new features in one layer from an existing feature in another layer, e.g. create a contract that covers a land parcel.

**Describe the solution you'd like**
There should be a copy tool in the editor menu that can copy a geometry from another layer, including drag-and-drop layers.

**Describe alternatives you've considered**
Create an external function that posts an event to the editstore to make a new feature, but that can't utilize a lot of the sweet attribute functions in edithandler and would not be consistent with other edit operations for the user.

Create a server side function that makes a copy in the database, but that is too application specific.

_I have written a copy tool that integrates with the edit toolbar. Soon to appear in a PR if approved_

jokd:
This will be a great addition to the editor so you are welcome to submit a PR.

PR #1225, opened 2021-04-06
Copy geometry

steff-o:
Implementation for #1223. Possibility to copy a geometry from one layer to another.

Instead of polluting edithandler with if:s everywhere, as the copy tool does not work like openlayers draw interaction, I have introduced the concept of _custom draw tools_, external to open layers. What it basically means is that it is possible to send an event containing a new feature to edithandler and it will be added just as if a new feature had been created inside the edithandler. With this framework in place it is possible to add more tools without having to change edithandler.

The actual copying is performed in a new tool: copyTool, that is added to the drawTools menu using configuration. The implementation is rather rudimentary and can only select one feature from one of the configured source layers. It uses openlayers select interaction for selecting in the map and the eventhandler performs the coping and sends a feature as an event that is handled by edithandler.

In order to make _custom draw tools_ work with the drawTools menu, the drawTools has been changed to handle more complex configuration than the previous openlayers shape tools, but for backwards compability reasons, the old shape name still works. It also have to call the actual tool when selected.  

Example configuration of a layer using copyTool:
```
"drawTools": [
        "Polygon",
        {
          "toolName": "Copy",
          "layers": [ "To2018_Swe99TM__Y", "SketchPoly" ],
          "groups": [ "drag-and-drop-layers" ]
        }
      ],
```
Where _toolName_ is required and at least a _layers_ or _groups_ array should to be specified. The _layers_ array specifies which layers are available for copying from. If the _groups_ array is specified all layers within that legend group will be available as source. Groups are not recursive. If neither _layers_ nor _groups_ are specified, all feature layers will be used as source.

### DIFF
--- a/src/controls/editor/copyTool.js
+++ b/src/controls/editor/copyTool.js
@@ -1,12 +1,15 @@
 import Select from 'ol/interaction/Select';
 import Feature from 'ol/Feature';
+import { MultiPolygon, MultiLineString, MultiPoint } from 'ol/geom';
 import dispatcher from './editdispatcher';
 
 // Point of entry. Create on of these each time the tool is selected
 //  viewer: the viewer
+//  editLayer: the destination layer for edits
 //  options: the current drawTools configuration for this layer
-const copyTool = function copyTool(viewer, options) {
+const copyTool = function copyTool(viewer, editLayer, options) {
   const map = viewer.getMap();
+  const destinationLayer = editLayer;
   let selectLayers = [];
   let hasGoups = false;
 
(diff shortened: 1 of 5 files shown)


johnnyblasta:
Tested and got it to work!
Will do some more testing, before I approve, since only tested one simple case.

johnnyblasta:
Done some more test cases now:

- Specified "layers" to copy from and it works fine.
- Specified "groups" and added a "draganddrop"-control group. Didn't work unless I also specfied the drag and drop group in the groups hierarchy.
- Also tested on clicking copy where I would hit on multiple features. Got one of the features from one of the layers, don't know how it selected which feature it would use.

steff-o:
The group thing seems strange. I have tested without defining the group in groups hierarchy, only setting it in the drag and drop tool config. The copyTool just searches all layers with property group=_groupname_. The drag and drop tool also just adds the group name unless already present before setting group name on layer. So I think it should work either way as long as the layer actually has the group name set.

Regarding which object is selected, it is up to Openlayers. It uses the default behaviour of the select interaction in single select mode. Idea was to start small and improve functionality when requested. Main object with this version is performing those changes needed to edithandler without messing things up. The tool itself can be improved later as it is new and opt-in. If requested it is possible to add a multiselect interaction which triggers a selection popup (or whatever) instead of the javascript _confirm_ modal.

johnnyblasta (approved):
Strange I got it working with just the "draganddrop"-control now.
Good thinking on starting small and improve if requested.
Nice addition!
LGTM

PR #1234, opened 2021-04-16
Fixes geojson feature ids

### LINKED ISSUE #1232: Geojson feature IDs
**Describe the bug**
Geojson features gets an incremented ID even when there's an ID field in the geojson file. This can cause IDs to differ from time to time, especially when features are added/removed from the source, messing up the share map function for instance. The ID generation function should only kick in when necessary.

jokd:
Fixes #1232
Geojson features gets ID in this order:
1. The id property of the feature object (this is the correct geojson way of doing it)
2. The id property of the properties object (can be set to a different property on layer level, defaults to "idField": "id")
3. Incremented value starting at 1000000

This feature will get id 1:
{ "type": "Feature", "id": 1, "properties": { "id": 12, "name": "Karlstad" }, "geometry": { "type": "Point", "coordinates": [ 1503136, 8262205 ] } }

### DIFF
--- a/src/layer/geojson.js
+++ b/src/layer/geojson.js
@@ -12,7 +12,13 @@ function createSource(options) {
         const numFeatures = vectorSource.getFeatures().length;
         for (let i = 0; i < numFeatures; i += 1) {
           vectorSource.forEachFeature((feature) => {
-            feature.setId(i);
+            if (!feature.getId()) {
+              if (feature.get(options.idField)) {
+                feature.setId(feature.get(options.idField));
+              } else {
+                feature.setId(1000000 + i);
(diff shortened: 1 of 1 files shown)


tonnyandersson:
What's the point of having the incremented value (in #3 above) starting at 1000000? I can't see that it makes any difference.

jokd:
Just to minimize the probability for IDs to conflict in case you have id for some features in your geojson-file. And to easier distinguish the real values from the Origo generated.

tonnyandersson:
Fair enough. LGTM.

tonnyandersson (approved):

PR #1235, opened 2021-04-16
Adds the option to send headers when requesting url

johnnyblasta:
Add the headers that is to be sent as a object in the layers config

`"headers": { "Accept": "application/geo+json" }`


jokd:
LGTM

jokd (approved):

PR #1237, opened 2021-04-19
Fix that handles multiple nested attributes in a html attribute string

### LINKED ISSUE #1236: Multiple nested values in attributes html causes error
**Describe the bug**
When specifing multiple nested values in attributes html it causes an error.

**To Reproduce**
In the layer config setup this and it will cause an error
`"attributes": [
        {
          "html": "<p><b>Namn:</b> {{name.value}}</p><p><b>Beskrivning:</b> {{description.value}}</p><p><b>Temperatur:</b> {{waterTemperature.value}} &#176;C</p><p><b>Typ:</b> {{type}}</p>"
        }
      ]`

It works if only one instance of nested values are used, but when multiple are used it crashes an show an error 
`Uncaught TypeError: Cannot convert undefined or null to object`

**Expected behavior**
There should be possible to have multiple nested values.

johnnyblasta:
Fixes #1236

### DIFF
--- a/src/utils/replacer.js
+++ b/src/utils/replacer.js
@@ -29,8 +29,20 @@ const replacer = function replacer() {
         }
         if (matches[1].indexOf('.') > 0) {
           const splitMatch = matches[1].split('.');
-          object = obj[splitMatch.shift()];
-          val = `${start}${splitMatch.join('.')}${end}`;
+          let objectTemp = obj;
+          // Drill down to the end of the nested attribute and return the value
+          do {
+            const key = splitMatch.shift();
+            if (key in obj) {
+              objectTemp = objectTemp[key];
+            }
+          } while (splitMatch.length < 1);
+          if (typeof objectTemp[splitMatch[0]] !== 'undefined') {
+            val = objectTemp[splitMatch[0]];
+          } else {
+            val = '';
+          }
+          object = obj;
         } else {
           object = obj;
         }
--- a/src/utils/replacer.js
+++ b/src/utils/replacer.js
@@ -33,10 +33,10 @@ const replacer = function replacer() {
           // Drill down to the end of the nested attribute and return the value
           do {
             const key = splitMatch.shift();
-            if (key in obj) {
+            if (key in objectTemp) {
               objectTemp = objectTemp[key];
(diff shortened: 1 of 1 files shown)


jokd:
Haven't tested but this looks a bit odd:
 
          do {
            const key = splitMatch.shift();
            if (key in obj) {
              objectTemp = objectTemp[key];
            }
          } while (splitMatch.length < 1);

Should it be splitMatch.length > 0 ?

johnnyblasta:
> Haven't tested but this looks a bit odd:
> 
> ```
>       do {
>         const key = splitMatch.shift();
>         if (key in obj) {
>           objectTemp = objectTemp[key];
>         }
>       } while (splitMatch.length < 1);
> ```
> 
> Should it be splitMatch.length > 0 ?

Nah, that dosen't works but this do:

`          do {
            const key = splitMatch.shift();
            if (key in obj) {
              objectTemp = objectTemp[key];
            }
          } while (splitMatch.length > 1);`

If it feels better.

johnnyblasta:
If you want to test, you can use this layer:

```
    {
      "name": "beachtemperature",
      "title": "Badplatser",
      "group": "root",
      "source": "https://kartatest.sundsvall.se/test/beach.geojson",
      "style": "simple",
      "queryable": true,
      "visible": true,
      "projection": "EPSG:4326",
      "type": "GEOJSON",
      "headers": { "Accept": "application/geo+json" },
      "attributes": [
        {
          "html": "<p><b>Namn:</b> {{name.value}}</p><p><b>Beskrivning:</b> {{description.value}}</p><p><b>Temperatur:</b> {{waterTemperature.value}} &#176;C</p><p><b>Typ:</b> {{type}}</p>"
        }
      ]
    }
```

jokd:
Checking the code again I see why splitMatch.length > 1 makes more sense so if you change to that it will probably work with more levels as well. The way it is written now it will just run the do while once.
With that change it is good to merge.

johnnyblasta:
Done the change from less than to greater than sign and also changed the check for if the key exists in the temp object, so now it works for multiple levels. Have tested for both 3 and 4 levels.

PR #1241, opened 2021-04-20
Changes target for links

### LINKED ISSUE #1240: Ability to specify target on attribute links
When there is a map that suppose to always be embedded the default target of attribute links _blank will create a lot of new windows when clicked.

I want it possible to set the target to _top, _self and _parent for the url attribute.

To specify a name of the target is also a possibility, but don't know if we should so permissive.

johnnyblasta:
Closes #1240


jokd:
Looks ok,suppose it can be done with 
`else { aTarget = attribute.target ? attribute.target : '' }`
as well to support other obscure frame targets but either way works

jokd (approved):

johnnyblasta:
I did it like that so only official target attribute values would be supported.

PR #1243, opened 2021-04-22
Expose the embedded status of the viewer

### LINKED ISSUE #1239: hideWhenEmbedded for plugins(?)
Controls have a hideWhenEmbedded param that can be specified in index.json, I've not seen that plugins have the same possibility

Would be good if plugins (that introduce a UI element) could have a similar property

There is a possibility that I've missed something

Grammostola:
Fixes #1239 . (Negation is because of eslint(no-shadow))

Plugins can be added if isNotEmbedded is true or not depending on whether if one wishes them when the map is embedded or not
`
      origo.on('load', function (viewer) {
	  if (viewer.isNotEmbedded()) {
        var draw = Draw({..`

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -254,6 +254,10 @@ const Viewer = function Viewer(targetOption, options = {}) {
 
   const getMain = () => main;
 
+  const isNotEmbedded = function isNotEmbedded() {
+    return !isEmbedded(this.getTarget());
+  };
+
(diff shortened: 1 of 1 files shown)


tonnyandersson:
Why not...

```
const getEmbedded = function getEmbedded() {
  return isEmbedded(this.getTarget());
};
```
...?

jokd (approved):

PR #1245, opened 2021-04-28
Adds padlock for layers

### LINKED ISSUE #1207: Padlock for layers that is not included in getCapabilities
If a layer is not included in getCapabilites, it should be indicated in the legend, for example by adding a padlock on the layer.

asemoller:
Fixes #1207.

Config:

```
"source": {
    "sigtuna": {
      "url": "https://karta.sigtuna.se/geoserver/ows",
      "capabilitiesURL": "https://karta.sigtuna.se/geoserver/ows?service=WMS&version=1.1.1&request=GetCapabilities"
    }
```

If a layer isn't included in getCapabilities it gets a padlock. Only tested with Geoserver.
image


johnnyblasta:
Tested and got undesired behavior when setting the layer to `"visible": false` the padlock is added and it's not possible to turn on the layer although it should be readable.

asemoller:
@johnnyblasta I don't get the same results as you, for me it works as expected. Maybe you need the following setting in geoserver to make it work:
 
image

You find it under Security > Data in Geoserver.

johnnyblasta:
Actually tested with a static GetCapabilities XML file, which reasonbly should produce the same result since the xml is converted to a array

It also throws a deprecation warning which should be looked:
image

Jonas074:
Tested and works as expected in our environment with Geoserver

jokd:
Seems to work when configured in certain way, but we have configured layers in our map as "name": "workspace:layername" in our config but that syntax seems to be invalid.

PR #1253, opened 2021-05-04
Overscroll behavior on body

### LINKED ISSUE #1252: Page refresh by accident
When a modal is open and you accidentally drag outside the modal vertically (top to bottom), the page reloads. If youre almost finished with a form and this happens cause you wants to scroll down a bit it can be a bit frustrating.

So, page refresh should not interfere with UI

mulfvik:
Fixes #1252

### DIFF
--- a/scss/origo.scss
+++ b/scss/origo.scss
@@ -51,6 +51,7 @@ html,
 body {
   height: 100%;
   margin: 0;
+  overscroll-behavior-y: contain;
 }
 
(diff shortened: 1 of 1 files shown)


johnnyblasta (approved):
Have now tested on Android with Chrome and Samsung browser and through Browserstack on an Iphone 11 with Safari and the fix works on those, so LGTM.

Issue #1227, opened 2021-04-14
Editing can produce invalid geometries


steff-o:
**Describe the bug**
It is possible to create self-intersecting geometries when drawing polygons and linstrings in the editor control. Either by creating a new geometry or edit an existing. These are almost always (OGC) considered invalid in the database and may cause problems when trying to work with them later.

**To Reproduce**
1. Start editing a polygon layer
2. Create a new feature and make it like a figure 8 (self intersect)
3. Save
4. Go to your database and check STIsValid() (or equivalent) on created geometry. It will return 0.

**Expected behavior**
It should not be possible to make the outer boundary self-intersect when drawing. If it would help to allow self-intersecting  while drawing it should at least not be possible to finnish the editing or add the finished feature to the transaction.

**System set up**
 - Origo 2.3.1
 - Geoserver 2.16.2
 - Sql Server Express (yeah I know. It's a test setup)

**Additional context**
While it could be considered a business rule for each application to require valid geometries it is diffucult to handle the errors that can arise as a result of invalid geometries. As most usage only includes editing in origo and saving to database using WFS  there are not many places to enforce valid geometries. 

The first attemtp to work around this would be adding a trigger in the database that either corrects the geometry or rejects the transaction. Both of which comes way too late for a decent user experience. If rejected, the entire transaction will be rejected. If automatically fixed the users really doesn't know what they just saved.

jokd:
Real time validation and error handling would be nice. JSTS and turf can do this but will add some unwanted overhead.
Another neat feature would be drawing features similar to the measure control on mobile devices with a crosshair and a button rather than tap on screen but that is another issue.

steff-o:
I also saw JSTS and Turf had that, but implementing just a self-intersect is not that hard. Especially as at the moment it is only possible to create single ring polygons.  It's slightly more complex to validate multipolygons with holes as all rings has to be checked. As a side note, that feature request (creating multipolygons and holes) will probably come soon...

I can have a look at implementing a geometry validation function when editing. Not sure how to make it user friendly to in real-time avoid selfintersect when drawing polygons though. Maybe it's possible to delete the last added vertex if if will create self-intersect using the geometryFunction callback to draw interaction or get a reference to the feature being drawn at drawstart-event and add a change handler to that geometry. But at least rejecting a selfintersecting geometry after it has been created should be easy.

jokd:
[...]

PR #1255, opened 2021-05-05
Prevent self intersect when drawing

steff-o:
Fix for origo-map/origo#1227.

Adds a couple of validation functions that is called when drawing geometries in edithandler.

It prevents adding vertices that would create an invalid geometry. At the moment it only considers self intersecting LineString and self intersecting outer ring of Polygon. For MultiLineString and Multipolygon it only considers the first object as it is currently not possible to draw multi-versions or holes.

When drawing freehand, no check is performed until feature is completed. If invalid, the entire geometry is rejected.

When modifying features an invalid move is reversed on completion. It only checks first poly outer ring. But if modifying features created elsewhere, features could have multiple polygons and inner rings that should be checked as well.

Feature is opt-in and is activated using the option `"validateOnDraw" :  true` set on the editor. 

Some features considered but not implemented:
1. Flash where an invalid vertex insertion is attempted. Right now it only refuses to create vertex
2. Separate configuration for "live" validation and on completion.


johnnyblasta:
Tested with both linestring and polygon and it LGTM, resolve the conflict and then it is good to merge.

steff-o:
Fixed conflict.

johnnyblasta (approved):
LGTM

PR #1271, opened 2021-05-25
Fixes issue with subgroups

### LINKED ISSUE #1121: Legend with nested groups
**Describe the bug**
When nesting groups more than two levels (sub-subgroup that is), the initial symbol for the group layer may be wrong. Toggling groups on and off makes it work expectingly.

**To Reproduce**
1. Have at least three levels of groups with som layer not visible at start
2. The symbols for the group layer (empty, grey checkmark, black checkmark) may not be accurate.

**Screenshots**
image

jokd:
Fixes #1121 
This solves the bug but still remains how to handle empty groups. Remove? Checked? Unchecked?

### DIFF
--- a/src/controls/legend/overlays.js
+++ b/src/controls/legend/overlays.js
@@ -140,6 +140,15 @@ const Overlays = function Overlays(options) {
     }
   };
 
+  const updateLegend = function updateLegend(comp) {
+    const parentName = comp.parent;
+    const parent = groupCmps.find((cmp) => cmp.name === parentName);
+    parent.dispatch('add:overlay');
+    if (parent.parent) {
+      updateLegend(parent);
+    }
+  };
+
   const addLayer = function addLayer(layer, {
     position
   } = {}) {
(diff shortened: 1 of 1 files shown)


jokd:
After discussion we aim to treat empty groups as if they didn't existed. That will hopefully be handled in another PR.

tonnyandersson:
@jokd I think it's a good idea to fix the issue with empty groups in this PR as well. Small issue, sure, but why not?

I think there might be quite a simple solution to it.

Empty groups are never checked properly for visible layers. I _think_ that's done only when layers (overlays) are added, or when layers or subgroups are toggled. Since empty groups never gets checked for visible layers, an empty group will retain the default icon, which is the "checked" icon (`#ic_check_circle_24px`). I think it makes more sense to use the "unchecked" icon (`#ic_radio_button_unchecked_24px`) as the default icon.

This is set on the `tickButton` const in **src\controls\legend\group.js**. See image below.

image

I guess the same result can be achieved by `icon: getCheckIcon(visibleState)`. I think.

Tested and working with a handful of scenarios. Not sure if it's absolutely foolproof, though, but I believe it's quite solid.

jokd:
Works ok but not flawless. Made an awesome gif to show it. The empty subsubgroup messes things up. Depending on the order of clicks the parent group can be checked/semi-checked when layer is visible and semi-checked/empty when layer is not visible.
chrome-capture

tonnyandersson:
True. Oh well, at least the original issue is solved. And empty subgroups should be a rare sight so I guess it won't be a big deal.

tonnyandersson (approved):

Issue #1268, opened 2021-05-18
Add more access control options on editor


steff-o:
**Is your feature request related to a problem? Please describe.**
If a layer is editable the user can perform all edit actions. This could potentially cause damage if the user can remove features or modify geometries when the task is to update attributes on existing objects.

**Describe the solution you'd like**
It should be possible to define in configuration which edit actions are allowed on a layer. It should be possible to set a combination of:

- Create
- Delete
- Update

Also the geometry should have the option of being read only like any other attribute. I know that this might seem odd in a map application, but in certain administrative applications it can be useful to use a map to edit attributes.

**Describe alternatives you've considered**
Limit access in database, but that would result in an confusing error on save.

Update attributes outside Origo when the geometry should be read only, but that seems like double work and may be a bad user experience,

johnnyblasta:
Not odd that the geometry also could be readonly if only the other attributes should be edited when the geometry is a reference to a external source that maintains the geometry.

steff-o:
Closed by #1281

PR #1281, opened 2021-05-31
Set allowed tools for each editable layer

steff-o:
Fix for #1268. Makes it possible to set which editing tools are available for each editable layer.

By providing a list of allowed edit operations it can be controlled what a user can do with an editable layer. The feature requries that the layer is configured as editable, otherwise configuration has no effect. For backwards compability if a layer is configured as editable but no `allowedEditOperations` array is specified all operations are allowed. If  the `allowedEditOperations` array is specified but empty no tools are available.

```
"allowedEditOperations": [
        "updateAttributes",
        "updateGeometry",
        "create",
        "delete"
      ],
```

Possible values are:
- updateAttributes: it is possible to update attributes on existing features. Newly created features are treated as existing. So if create is allowed, but not updateAttributes it is not possible to set any attributes unless `autoForm` is set to true.
- updateGeometry: it is possible to modify an existing feature. Newly created features are treated as existing. So if create is allowed, but not modify it is not possible to change a feature once it is created even if it is not saved yet. By specifying all options except `updateGeometry` the requested feature _read only geometry_ in #1268 is accomplished.
- create: Allows creation of new features. 
- delete: allows deletion of features.

The configuration can only be set at layer level, not editor level.

The editor toolbar only shows the tools that are available for currently edited layer. Also the edit handler enforces the options to some extent. However, as it is client side it is still possible to inject changes by circumventing the editorhandler, so if security is a concern the editing rules must be enforced at database level.


johnnyblasta:
Tested and it works fine accept for one thing: 
When the default layer has some tools disabled they are not removed from the tools menu. When changing layers the tools are removed, but not initial.

steff-o:
Well spotted. I forgot to set the allowed tools when first displaying the toolbar. If editor is enabled at startup there is no change layer event and all tools were visible.

Fixed now by always setting the allowed tools when the toolbar is set to visible.

johnnyblasta (approved):
LGTM and great addition.

PR #1283, opened 2021-05-31
Exclude print footer from filter (so it prints)

### LINKED ISSUE #1282: Print - The new print footer element is not included when printing to PDF
**Describe the bug**
Since PR #1097, the left footer component and the created time on the print page are not included when printing to PDF.

**To Reproduce**
1. Open the print control
2. Open the print settings
3. Choose to include the created time ("Visa skapad tid")
4. Print to PDF

**Expected behavior**
Everything visible in the print layout should be included in the PDF

 - OS: Windows
 - Browser: Chrome
 - Version: 90

MattiasSp:
Fixes #1282.

The print control uses the [printToScalePDF()](https://github.com/origo-map/origo/blob/d279d9c1f3d230c97a7af77e08b2bd7a2cb4c744/src/utils/download.js#L116) function in [utils/download.js](https://github.com/origo-map/origo/blob/master/src/utils/download.js) to generate downloadable PDFs.

This adds the elements `o-print-footer` and `o-print-footer-left` to the list of `o-print`-related elements that _should_ be included in the print (i e not excluded by the filter).

### DIFF
--- a/src/utils/download.js
+++ b/src/utils/download.js
@@ -136,6 +136,8 @@ export const printToScalePDF = async function printToScalePDF({
         || className.indexOf('padding-right-small') > -1
         || className.indexOf('padding-bottom-small') > -1
         || className.indexOf('o-print-description') > -1
+        || className.indexOf('o-print-footer') > -1
+        || className.indexOf('o-print-footer-left') > -1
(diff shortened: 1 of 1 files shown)


johnnyblasta (approved):
LGTM

PR #1286, opened 2021-06-04
Fix for SVG errors when printing to PDF with featureinfo overlay

### LINKED ISSUE #1285: Unable to print to PDF after opening featureinfo overlay
**Describe the bug**
Trying to print to PDF after first displaying the featureinfo overlay popup results in the following error, and then nothing:
```
Uncaught (in promise) TypeError: className.indexOf is not a function
    at filter (download.js:229)
    at cloneNode (dom-to-image-more.js:200)
    at eval (dom-to-image-more.js:230)
```

**To Reproduce**
1. Open the feature info popup in default (overlay) mode for any feature, either by clicking on it or by selecting a search result from the search control.
2. Close the popup, or don't, it's enough that it has been rendered.
3. Open the print control.
4. Click on "Skapa pdf".

**Expected behavior**
A short wait followed by an offer to download a pdf file.

 - OS: Windows 10
 - Browser: Chrome / Firefox
 - Version: 91 / 89

 - Device: Samsung Galaxy A20e
 - OS: Android 10
 - Browser: Chrome Mobile
 - Version: 90

**Additional context**
The underlying issue is that there is an SVG element (the Close button of the feature info overlay) that is not handled correctly [in the print filter of utils/download.js](https://github.com/origo-map/origo/blob/2a11f75bc5df0c24748713dabc6c40c97eb6a6f7/src/utils/download.js#L131). Because the featureinfo overlay popup is hidden rather than removed when clicking the close button, the SVG element is still present as long as it has been rendered in the first place.

The culprits, an `<svg>` and its `<use>` element, have an `SVGAnimatedString` object as their `className` property instead of a string of CSS classes. These objects cannot be searched with `indexOf()` in the print filter.

MattiasSp:
Fixes #1285.

The print filter checks the `className` property of each DOM element supplied using `indexOf()` which assumes that `className` is always a string. In the case of the close button of the featureinfo overlay, its `<svg>` and `<use>` elements have objects as their `className` property.

In these cases, classes set on those elements can instead be accessed through their `classList` property.

### DIFF
--- a/src/utils/download.js
+++ b/src/utils/download.js
@@ -128,7 +128,10 @@ export const printToScalePDF = async function printToScalePDF({
   // See: https://github.com/bubkoo/html-to-image#options
   const exportOptions = {
     filter(element) {
-      const className = element.className || '';
+      let className = element.className || '';
+      if (typeof className === 'object' && element.classList) {
+        className = `${element.classList}`;
+      }
       return (
         className.indexOf('o-print') === -1
         || className.indexOf('o-print-header') > -1


johnnyblasta (approved):
Tested and LGTM

Issue #1158, opened 2021-02-03
Configurable control button placement


MattiasSp:
**Is your feature request related to a problem? Please describe.**
At the moment, every control has its placement fixed, defined in the control itself. With many controls enabled, the "northwest" screen section can be quite cluttered. Conversely, some controls (such as the Print control) may in certain cases be better placed on screen rather than in the menu.

**Describe the solution you'd like**
A general control placement option for each control in the config which lists the containers its button(s) should be displayed in:
```
{
  "name": "print",
  "options": {
    "placement": ["screen", "menu"]
  }
},
```

**Describe alternatives you've considered**
Placement options could include all screen corners; `northwest | northeast | southwest | southeast | menu`, but perhaps it would make it unnecessarily complex to write controls which display perfectly in either of the five containers. Choosing between `screen | menu` is easier and can be a perfectly acceptable first step.

**Additional context**
The initial idea comes from the desire to show an on-screen button for the Print control.

MattiasSp:
This relates to #1208.

MattiasSp:
Perhaps control button placement options should be built into a sort of OrigoControl parent class which controls could inherit from? Then the handling of control button placement would be in a single place, and all controls could be passed button placement options on creation...

tonnyandersson:
@MattiasSp Sorry for the delayed response.

Personally, I agree with your last comment. The button placement handling should be done in one place. Seems more logical than adding the same thing on each control separately. This will probably be up for discussion on the next dev meeting.

jokd:
I merged this PR and this code can be used when rebuilding how buttons are placed.

sweco-senari:
```  
"options": {
    "placement": ["screen", "menu"]
  }
```

Only a few tools that support placement configuration (about,  draw, link, print).

Is it ok to implement this in more controls or did the dev team discuss alternate ways the placement configuration should be designed?

I'm looking to solve an issue where sharemap has a hard dependecy to the "mapmenu"-control. If mapmenu does not exists then sharemap is not possible to use.

MattiasSp:
> Is it ok to implement this in more controls or did the dev team discuss alternate ways the placement configuration should be designed?
> 
> I'm looking to solve an issue where sharemap has a hard dependecy to the "mapmenu"-control. If mapmenu does not exists then sharemap is not possible to use.

Go ahead and post a new issue for adding the placement parameter for the sharemap control.

[...]

PR #1287, opened 2021-06-08
Configurable print button placement

MattiasSp:
Relates to #1158.

Adds a new config option for the print control:
 `"placement": ["menu", "screen"]`

The new option controls whether to put buttons for opening the print control on the map screen, in the menu or both.

Default is  `"placement": ["menu"]`.


jokd (approved):
Works well and LGTM

PR #1289, opened 2021-06-10
fix: correct the two instances of ~getCenter() to handle multiPoint g…

### LINKED ISSUE #1288: Multipoint geometry does not seem to work with search.js
**Describe the bug**
Crash when search hit of multipoint layer is clicked

**To Reproduce**
Have a PostGIS/other layer with geom Geometry(MultiPoint, epsg)
Configure an Origo client and Origo server to search and serve results from that layer

**Expected behavior**
When search hit is clicked, map is zoomed and panned and infowindow shown

**Additional context**
The issue seems to be getcenter.js case for MultiPoint geom:
```
   case 'MultiPoint':
      center = geometry[0].getCoordinates();
      break;
```
it fails because the geometry object never seems to be an array. The same method is also employed in maputils.js

Grammostola:
…eometry presumably anew

Fixes #1288 , Well, only getcenter.js is used in the search-multipoint-geom-and-click-search-hit scenario described in the issue, maputils.js has the same function though more or less and seems to make the same assumption about the incoming geom so my suggestion is to update that too, if somewhat blindly.

### DIFF
--- a/src/geometry/getcenter.js
+++ b/src/geometry/getcenter.js
@@ -18,7 +18,7 @@ export default function getCenter(geometryIn, destination, axisOrientation, map)
       center = geometry.getCoordinates();
       break;
     case 'MultiPoint':
-      center = geometry[0].getCoordinates();
+      center = geometry.getPoint(0).getCoordinates();
       break;
     case 'LineString':
       center = geometry.getCoordinateAt(0.5);
--- a/src/maputils.js
+++ b/src/maputils.js
@@ -90,7 +90,7 @@ const maputils = {
         center = geometry.getCoordinates();
         break;
       case 'MultiPoint':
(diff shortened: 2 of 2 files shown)


tonnyandersson:
LGTM.

However, one could argue that it's weird to return the first point only. In some cases you would probably want to return the center point of the extent of a MultiPoint geometry, or something like that. Handling of _true_ MultiPoint geometries, that is. But that's a different issue.

tonnyandersson (approved):

Grammostola:
Fair comment, there are probably cases where getCenter() should return more of a center / centroid and less of an actual point. As you say it can be deemed a separate issue; this PR merely makes getNoCenter() work :)

Issue #1259, opened 2021-05-10
Change pointer when hovering over feature


steff-o:
**Is your feature request related to a problem? Please describe.**
It is not obvious for the user that a feature is clickable to get featureinfo.

**Describe the solution you'd like**
The mouse pointer should change to 'pointer' when hovering over an area in the map where the user can click to get featureinfo.

**Describe alternatives you've considered**
The following code can be added whithout changes to Origo, but it would be nice if it was a feature of origo.

The code snippet only works on WMS-layers if the layer has the setting `className` set to a unique value, otherwise it produces false positives on other layers around WMS-features (don't know why, it's documented briefly in OL documentation for foEachLayerAtPixel).

Also it only works on visible symbols where there actually is a pixel, so it won't work on hollow area symbols and hatched areas becomes "nervous".

```
let pointerActive = true;
document.addEventListener('enableInteraction', evt => pointerActive = evt.detail.interaction !== 'editor');

 viewer.getMap().on('pointermove', e => {
   if (!pointerActive || e.dragging) return;
   e.map.getViewport().style.cursor = viewer.getMap().forEachLayerAtPixel(e.pixel, () => true, { layerFilter: (l) => l.get('queryable') }) ? 'pointer' : '';
            });
```

**Additional context**

PR #1292, opened 2021-06-11
Link pointer when hovering over clickable feature in map.

steff-o:
Implementation for #1259

Changes the mouse pointer to link pointer when hovering over a feature that can be queried for feature info.

Feature is opt in to be backwards compatible and is activated by adding `"linkPointer": true` to _featureInfoOptions_.

Make sure to turn off _queryable_ on background maps, as _queryable_ is default _true_ and would quite spoil the usefulness of this feature.

The solution is based on [forEachLayerAtPixel](https://openlayers.org/en/latest/apidoc/module-ol_Map-Map.html#forEachLayerAtPixel) to detect features from WMS as well. As _forEachLayerAtPixel_ requries that each layer has a unique css class, a class is added to each queryable layer if the layer does not have one and the feature is activated.

### DIFF
--- a/src/featureinfo.js
+++ b/src/featureinfo.js
@@ -423,6 +423,26 @@ const Featureinfo = function Featureinfo(options = {}) {
           setActive(false);
         }
       });
+
+      // Change mouse pointer when hovering over a clickable feature
+      if (viewer.getViewerOptions().featureinfoOptions.linkPointer) {
+        let pointerActive = true;
+        document.addEventListener('enableInteraction', evt => {
+          pointerActive = evt.detail.interaction !== 'editor';
+          // Avoid getting stuck in pointer mode if user manages to enable editing while having pointer over a clickable feature.
+          // If the user manages to disable editing while standing on a clickable feature, it will remain arrow until moved. (Sorry)
+          map.getViewport().style.cursor = '';
+        });
+
+        // Check if there is a clickable feature when mouse is moved.
+        map.on('pointermove', evt => {
+          if (!pointerActive || evt.dragging) return;
+          // Just check if there is a pixel here. Pretty annoying on hatched symbols or hollow areas.
+          // Note that forEachLayerAtPixel requires that layers have unique class names to work correct.
+          // Hit tolerence seems to be ignored. It would probably look funny anyway.
+          map.getViewport().style.cursor = map.forEachLayerAtPixel(evt.pixel, () => true, { layerFilter: (l) => l.get('queryable') }) ? 'pointer' : '';
+        });
+      }
     },
     render
   });
--- a/src/layer.js
+++ b/src/layer.js
@@ -54,7 +54,8 @@ const Layer = function Layer(optOptions, viewer) {
 
   layerOptions.name = name.split(':').pop();
 
-  if (layerOptions.css) {
+  // Some features require that each layer has a unique class name. Set one if not provided.
+  if (layerOptions.css || (viewer.getViewerOptions().featureinfoOptions.linkPointer && layerOptions.queryable)) {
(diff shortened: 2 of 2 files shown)


tonnyandersson:
Only tested on WMS and GEOJSON type layers, but it seems to do the trick.

A few questions, though.

As I understand it, giving layers a class name _may_ have a negative impact on performance since layers with class names will be rendered as separate HTML elements. With that in mind, I wonder if it's wise to set a class name on _all_ layers like this. Perhaps class namnes should be set on non-vector type layers only? But then there's the issue with possible false positives. I didn't notice anything like that when testing, but I only tested on a map with very few layers/features. I guess it comes down to a tradeoff between performance and possible false positives. Not sure which one will have the most negative effect on the user experience at this moment.

Also, I don't think the property name `linkPointer` feels very intuitive since we're not dealing with links here. I suggest renaming it to `pointerOnHover` or something like that. Just a suggestion, though.

steff-o:
@tonnyandersson: I have experienced false positives using only two layers, one WMS and one WFS on separate canvases but same class name.

The idea is to only set class name on layers that are configured as "queryable" to get a pointer change on all layers, regarding of type, as it also works on vector layers. Maybe it is possible to use `forEachFeatureAtPixel` for feature layers if that produces better performance.

I have also had concerns about performance, that's one reason the feature is opt-in. But when I have tested, my map is always drawn with with a separate div and canvas for each layer, regardless if I use a class name or not. There might be something wrong with my setup, as I also think that all layers not using separate class name should use the same canvas. At least I have seen other OL-applications using only one canvas-element. I have to look into it and see if it is OL, origo or my configuration that creates all those canvases. Not sure how much it affects performance though. But in another application we have serious performance issues when when we have a lot of WMS:es using separate canvases for each layer. The perfomance issue in that case is drawing the map, the mouse pointer works fine. Never managed to make origo merge them in to one canvas there either. In that application I traced the performance issue to have something to do with drawing the actual pixels (according to chrome dev tools), as it was slow even if all tiles where fetched from local cache. But changing layer type to WFS made it lightning fast, even with separate classes (and canvases) for each layer.

I also think the option should be renamed, I never liked it myself as it could be interpreted as that the pointer should be linked (whatever that would mean). Maybe `changePointerOnHover`.

To sum it up: I will do some more performance analysis.

steff-o:
After investingating how forEachLayerAtPixel actually behaves, I have changed the code so it will not add a unique className to each queryable layer, but instead adds a `o-layer-queryable` class to all layers that do not already have a className.  In this way queryable and non-queryable layers are not mixed in the same canvas (as forEachLayerAtPixel only checks if canvas has a pixel, not which actual layer). If consecutive layers have the same class, they will be drawn on the same canvas. For optimal performance the configuration should not intersperse queryable and non-queryable layers in the configuration.

That said, after testing actual performance there is no improvement in drawing speed between using one canvas or one canvas for each layer, but there may be some other advantages of keeping the number of canvases low. (Tested with 100 layers wich is slooow in both cases).

Also renamed the option to `changePointerOnHover`

jokd:
@tonnyandersson Thoughts on this?

tonnyandersson:
@steff-o It seems layers are automatically rendered as separate elements if you add any kind of visual trickery on one or more layers. E.g. in the default config, the opacity is changed on the `mask` layer. If you remove that property, you'll see that all layers are rendered on the same canvas.

I think this feels pretty solid and I guess a (possible) slight decrease in performance is preferable to false positives.

However, there are some linter errors in layer.js, mainly regarding spaces after `//` in your comments. Please clean them up before I approve.

steff-o:
New lint friendly comments. Seems like lint autocorrects those errors, so it passed lint on my machine but left repo changed as I had already commited before running lint the last time.

PR #1294, opened 2021-06-17
Fix for map not autopanning vertically

### LINKED ISSUE #1279: The animation for the infowindow doesn't work
**Describe the bug**
The animation for the info window "overlay" doesn't work if it's close to the top of the browser window. It's works fine when clicking on an object close to the left or right. It has worked in previous versions.

**To Reproduce**
1. Click on an object close to the top of the window.
2. The info window hides

**Expected behavior**
The info window should be adjusted so it can be readable.

 Only tested on desktop/chrome.

animation_infowindow

jokd:
Fixes #1279 - Rearranges stuff for the carousel

### DIFF
--- a/src/featureinfo.js
+++ b/src/featureinfo.js
@@ -256,17 +256,6 @@ const Featureinfo = function Featureinfo(options = {}) {
         });
         popup.setVisibility(true);
         initCarousel('#o-identify-carousel');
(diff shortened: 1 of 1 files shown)


johnnyblasta (approved):
Tested and LGTM!

PR #1301, opened 2021-06-29
New config option to close the bookmarks popup when clicking on a link on small screens

### LINKED ISSUE #1300: Close Bookmarks popup on click
On small screens, a long list of bookmarks tends to hide a large portion of the map as well as the zoom animation when clicking on a bookmarked location.

A config option to hide the bookmarks popup automatically when the user clicks on a bookmark link would be easy and elegant. If the auto-close behaviour could be dynamic so that it is only active on smaller screens would be even better.

**Alternatives considered**
- Always auto-close
  - Changes current behaviour
  - Less flexible
  - More tedious for users on big screens to step between bookmarks

MattiasSp:
Fixes #1300.

Adds a new config option to the bookmarks control: `autoClose`. When set to `true` and using screen size `m`, `s` or `xs`, the popup will close after clicking on a bookmarks link. Default is `false`.

Example config:

```
{
  "name": "bookmarks",
  "options": {
    "title": "Bookmarks",
    "isActive": true,
    "duration": 1000,
    "autoClose": true,
    "items": [
      {
        "name": "Hudiksvall",
        "coordinates": [1904063, 8794995]
      },
      {
        "name": "Karlstad",
        "coordinates": [1503136, 8262205]
      },
      {
        "name": "Sigtuna",
        "coordinates": [1986473, 8315061],
        "zoomLevel": 10
      }
    ]
  }
}
```

### DIFF
--- a/src/controls/bookmarks.js
+++ b/src/controls/bookmarks.js
@@ -8,7 +8,8 @@ const Bookmarks = function Bookmarks(options = {}) {
     duration = 300,
     closeIcon = '#ic_close_24px',
     bookmarksIcon = '#ic_bookmark_24px',
-    title = 'Bokmärken'
+    title = 'Bokmärken',
+    autoClose = false
   } = options;
   let {
     isActive = false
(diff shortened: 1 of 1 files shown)


MattiasSp:
An example build is published here: https://karta.haninge.se/devmaps/PR1301/

jokd (approved):
Useful!

PR #1302, opened 2021-07-19
Added option to disable the subgroup toggle-all function

### LINKED ISSUE #502: Disable group visibility toggle
Today, the user can toggle all layers within a subgroup on or off simultaneously by clicking the subgroup checkbox. This non-configurable default behavior can cause serious performance issues if a user tries to switch on a subgroup filled with, for example, large raster layers.

I suggest making this behavior configurable, preferably on application level in order to maintain consistency thoughout the map menu.

tonnyandersson:
Fixes #502.

Adds the option `toggleAll` to (sub)groups. Default is `true`.

Setting `toggleAll` to false on a subgroup will remove the group toggle ability. Unlike the `exclusive` option, this will not restrict the ability to toggle individual layers on/off at will.

toggle_all_false

### DIFF
--- a/src/controls/legend/group.js
+++ b/src/controls/legend/group.js
@@ -19,7 +19,8 @@ const Group = function Group(options = {}, viewer) {
     position = 'top',
     type = 'group',
     autoExpand = true,
-    exclusive = false
+    exclusive = false,
+    toggleAll = true
   } = options;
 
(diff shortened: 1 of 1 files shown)


jokd (approved):
Works well, nice!

Issue #858, opened 2020-03-10
More style filter options


asemoller:
It would be very useful to be able to use filters with logical operators such as “AND” and “OR” when styling layers. Either with text (AND, OR etc.) or with regex.

Or is this already possible?

tonnyandersson:
No, this is not already possible, as far as I know. The filter function can only deal with one argument.

asemoller:
I have seen solutions where you add support for regex in style.js, would that be an approved solution for Origo?

I have not tested it myself so I don't know how well it works. Regex is common and it is easy to find good examples and help by googling. Although a better solution is preferable, I think it's better with regex than no support at all.

hallstaelin:
We have regex in our style.js, https://github.com/Hallstahammarskommun/origo/blob/master/src/style.js

asemoller:
Nice @hallstaelin! Does it work well for you?

Are you ready to submit a PR if the solution is approved? :)

hallstaelin:
It works great!

Yes I can submit a PR if it gets approved :)

asemoller:
It works well with OR, AND and regex so I'm closing this issue. Thanks!

PR #1305, opened 2021-07-21
Adds AND or OR filtering on style

johnnyblasta:
More filtering on style as mentioned in #858

In thematic style add ' && ' to do AND filtering or ' || ' to do OR filtering.
Can be used multiple times, but not combine AND and OR filtering on the same filter

```
"filter": "[zon] == '5' && [status] == 'Avstängd'"
"filter": "[zon] == '5' || [status] == 'Avstängd'"
```

### DIFF
--- a/src/style.js
+++ b/src/style.js
@@ -168,33 +168,79 @@ function checkOptions(feature, scale, styleSettings, styleList, size) {
       });
       if (Object.prototype.hasOwnProperty.call(s[j][0], 'filter')) {
         let expr;
+        const exprArr = [];
         // find attribute vale between [] defined in styles
         let regexExpr;
         let regexFilter;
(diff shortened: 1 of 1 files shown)


jokd:
It doesn't work for me, tried this on the origo-logo-style:
"filter": "[id] == '3' || [name] == 'Sundsvall'"

Only the latter shows up in the map

johnnyblasta:
Fixed a problem with OR logic.
Changed the syntax to use AND/OR instead of &&/||

```
"filter": "[zon] == '5' AND [status] == 'Avstängd'"
"filter": "[zon] == '5' OR [status] == 'Avstängd'"
```

jokd (approved):

Issue #1310, opened 2021-08-13
toggleFeatureinfo event is no longer triggered correctly


tablackmore:
**Describe the bug**
When selecting a feature the toggleFeatureinfo event is not triggered unless you are on the second page of a carousel and it is then triggered after every click on the carousel div.

**To Reproduce**
1. Activate the dialog plugin example with voting
2a. Click on an object, no voting button will be displayed.

2b. Click on a stacked object, and then use the carousell button to move to the next object.
3b. The voting button will be displayed but you will be unable to trigger a click event on it.

**Expected behavior**
The voting button should be displayed with all object selections when active.
The voting button should be clickable.

**Additional context**
Fix on it's way. Based on "move" event in glidejs is the wrong event to wire upto "run.after" event is more appropriate.

steff-o:
Also mentioned in #1256. This would probably fix part of that.

PR #1311, opened 2021-08-13
fix: incorrect triggering toggleFeatureinfo event

tablackmore:
[Fix for issue 1310](https://github.com/origo-map/origo/issues/1310)

Two problems:
1) The toggle feature event was not triggered when the first feature is loaded
2) It was triggered too early and excessively. For example on carousell click irrespective of whether the click triggers the next feature in the carrousell or not.

This is most noteable when using the voting section of the dialogue-plugin.

These problems are solved by 
1) Always triggering the toggleFeatureinfo event when the first or a single feature is loaded.
2) Changing the glidejs event from 'move' to 'run.after', which is triggered only after a new feature is loaded in the carousell.
[glidejs events](https://glidejs.com/docs/events/)

### DIFF
--- a/src/featureinfo.js
+++ b/src/featureinfo.js
@@ -102,6 +102,16 @@ const Featureinfo = function Featureinfo(options = {}) {
     return title;
   };
 
+  const dispatchToggleFeatureEvent = function dispatchToggleFeatureEvent(currentItem) {
+    const toggleFeatureinfo = new CustomEvent('toggleFeatureinfo', {
+      detail: {
+        type: 'toggleFeatureinfo',
+        currentItem
+      }
+    });
+    document.dispatchEvent(toggleFeatureinfo);
+  };
+
   // TODO: direct access to feature and layer should be converted to getFeature and getLayer methods on currentItem
   const callback = function callback(evt) {
     const currentItemIndex = evt.item.index;
@@ -122,13 +132,7 @@ const Featureinfo = function Featureinfo(options = {}) {
         sidebar.setTitle(title);
       }
 
-      const toggleFeatureinfo = new CustomEvent('toggleFeatureinfo', {
-        detail: {
-          type: 'toggleFeatureinfo',
-          currentItem
-        }
-      });
-      document.dispatchEvent(toggleFeatureinfo);
+      dispatchToggleFeatureEvent(currentItem);
     }
   };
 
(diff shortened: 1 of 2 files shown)


steff-o:
Not sure if it is entirely correct implemented for _infowindow_ when multiple objects are selected.  The _ toggleFeatureinfo_ event is triggered for the first feature in the selection, but no feature is highlighted when _addItems_ is called. Maybe the first feature should also be highlighted. But it may depend on what the event really means when it comes to selection manager.

To be complete the _ toggleFeatureinfo_ probably should also  be fired when the selectionmanager highlights another feature. (depending on what the event really is for)

tablackmore:
> Not sure if it is entirely correct implemented for _infowindow_ when multiple objects are selected. The _ toggleFeatureinfo_ event is triggered for the first feature in the selection, but no feature is highlighted when _addItems_ is called. Maybe the first feature should also be highlighted. But it may depend on what the event really means when it comes to selection manager.
> 
> To be complete the _ toggleFeatureinfo_ probably should also be fired when the selectionmanager highlights another feature. (depending on what the event really is for)

Are we discussing two different events here, my understanding of toggleFeatureInfo event was it should be triggered when a features info is displayed in the carousell and is not related to an active selection (but I might be misunderstanding what an active selectionis). 

This event is currently used by the dialogue plugin so it can add a vote button to the feature information dialogue and that was where this bug started as the dialogue plugin was no longer working with the latest origo build since the event was not triggered in the same way it was before. If we use the event for other usecases such as selection at present it would be good with a code example for me to test so I can ensure it works for all current scenarios. 

If however it is a new feature that we wish to have which relates to a feature selection event I think it might be wise to do it as a seperate pr as this pr is a bug fix and todo with maintaining compatability.

I am very green with the origo repo so excuse any misunderstanding here.

tablackmore:
@steff-o I think I understand a bit more. I see that this PR is not working for  "infowindow": "infowindow", it only works with sidebar. I will have a look into that in this PR.

steff-o:
Well, I'm not sure what the event is used for, but I understand the reason for this PR. It annoyed me too that the event disappeared. If I look in the history it seems like the event has never been sent for infowindow. Probably because infowindow is a newer addition.  

The event could still be valid for infowindow when it used instead of overlay for popups. The highlighted feature in infowindow would then correspond to the visible feature in the carousel. This PR will emit the event for infowindow, but only when it is first shown, not when the highlight changes. But since it has never been fired for infowindow att all, that is like you say, another PR. (and I'm working on it)

tablackmore:
@steff-o Ok if toggleFeatureInfo for infowindow: infowindow has never worked and you are working on that I will leave it to you and leave this PR as is for now.

steff-o:
That's OK with me.

jokd (approved):

Issue #1316, opened 2021-08-19
Batch edit attributes


steff-o:
**Is your feature request related to a problem? Please describe.**
Updating a set of features with the same value must be done for each feature individually. This is annoying if the task is to set for instance a status on a selection.

**Describe the solution you'd like**
The user should be able to select a set of features and then update one or more fields in the selection with the same value. Fields not updated should keep their previous value.

**Describe alternatives you've considered**
None really.

**Additional context**
Special care must be taken when designing the GUI  to make it possible to distinguish between clearing out all values and keep the old value.

PR #1326, opened 2021-08-23
Batch edit attributes

steff-o:
Solves #1316. Makes it possible to batch edit attributes using the normal attribute editor form.

To batch edit attributes, the attribute must be configured with `"allowBatchEdit": true` in the _attributes_ section of the layer configuration.

If more than one feature is selected when the attribute edit form is opened, there will be a checkbox for each attribute that can be changed. Upon completion, all selected features will receive the new value for the attributes that have been checked.

To select multiple features use shift-click in the map.

### DIFF
--- a/src/controls/editor/editform.js
+++ b/src/controls/editor/editform.js
@@ -32,7 +32,7 @@ const createForm = function createForm(obj) {
       break;
     case 'checkbox':
       checked = val ? ' checked' : '';
-      el = `<div class="o-form-checkbox ${cls}"><label>${label}</label><input type="checkbox" id="${id}" value="${val}"${checked}${disabled}></div>`;
+      el = `<div class="o-form-checkbox ${cls}"><label for="${id}">${label}</label><input type="checkbox" id="${id}" value="${val}"${checked}${disabled}></div>`;
       break;
     case 'dropdown':
       if (val) {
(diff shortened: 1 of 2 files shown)


johnnyblasta:
Tested and it works as expected editing in batch.
However there are som errors introduced.

1. When only one feature is edited this error is thrown:
 ```
Uncaught TypeError: Cannot read properties of null (reading 'addEventListener')
    at Object.fn [as addListener] (edithandler.js:907)
```
2. When no attribute has `"allowBatchEdit": true` and multiple features are selected this error is thrown:
```
Uncaught TypeError: Cannot read properties of undefined (reading 'reduce')
    at editAttributes (edithandler.js:1076)
```

Also there is some lints in edithandler.js with comments and let instead of const.

steff-o:
Great testing. Apparently I forgot to check some of the basic cases. I'll have a look.

steff-o:
I have fixed the problem where no attribute is configured for batch edit, but I can not reproduce the first error. It seems to be related to when constraints are in place as that function is only called when using a constraint. But the only error I can get related to constraints is when an attribute with a constraint has `allowBatchEdit` but the attribute that it depends on does not and you select at least two objects. But that seems to me to be missconfigured.

While digging in the constraints I changed so that if a batch edited attribute is changed, its dependent attributes are also always changed without a checkbox to keep consistent with non-batch edits.

Also fixed some lints.

Issue #1167, opened 2021-02-17
Print - Highest scales cannot be printed in high resolution


MattiasSp:
When selecting to print in high resolution (300 dpi) and using the highest (most zoomed in) scale, the map does not change to the correct scale.

**To Reproduce**
1. Open the print control
2. Open the print settings
3. Choose high resolution (300 dpi)
4. Select the most zoomed in scale
5. The map scale is not what was just selected

Another path could be to provide a print scale in the print control's `scales` option that is below the minimum resolution in the maps `resolutions` config option.

**Expected behavior**
Any selectable scale should be possible to use for printing.

**Screenshots**
![image](https://user-images.githubusercontent.com/2445979/108229403-150dbe80-7140-11eb-810b-d9d2047ae927.png)

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Chrome
 - Version 87

**Smartphone (please complete the following information):**
 - Device: Samsung Galaxy S6
 - OS: Android 7.0
 - Browser: Chrome
 - Version: 88

**Additional context**

After going down the rabbit hole, I have concluded that when an OpenLayers map view object is initialized, it creates a [`resolutionConstraint`](https://github.com/openlayers/openlayers/blob/65df12068b47d6dcf1c43813f9ceb219bd5c5710/src/ol/View.js#L1862) which restricts the map's resolution (pixels per map units) to the minimum and maximum resolution that is was initalized with.

These resolutions are [converted to map scales in origo](https://github.com/haninge-geodata/origo/blob/994e72bb3ccca71728de686916afd2bd62ff1b94/src/maputils.js#L109), assuming a DPI of approx. 90.7.

When the DPI is increased, the effective _resolution_ will decrease proportionately (as each pixel will now represent a smaller map distance).

This means that increasing the DPI from e.g. 150 to 300 DPI cuts the requested map resolution value in half, which for the most zoomed in scale(s) means that the requested resolution will be lower than what is allowed by the above mentioned resolution constraints.

I believe there might be ways to deactivate/ignore the map view's resolution constraints, but I'm not sure how yet. That's as far as I've gotten for now.

Grammostola:
Well investigated. I'm not sure how to disable a View's resolution constraint either.

A tangent: if constrainResolution: true is set for the map (to zoom-snap to its defined resolutions) it becomes very hard to use any selectable scale for printing.

MattiasSp:
[...]

PR #1327, opened 2021-08-30
Recalculate resolutions when entering print view

steff-o:
Fix for #1167. Recalculates the resolutions array when entering print preview and restores them when print component is closed. By doing this the scale selector and the print preview gets the same scale when _constrainResolution_ is set regardless of DPI setting. 

It also recalculates max zoom and min zoom to allow the print preview to zoom beyond the original limits to provide for other DPI settings. As an implementation note, temporarily setting _constrainResolution_ on the map is not sufficient to solve the min and max zoom problem, as that parameter only controls if zoom values between the defined levels are allowed.

The feature is opt-out as it is most likely how it should work, but if it just happens to destroy something else, it can be turned off by setting `"supressResolutionsRecalculation" : true` on the print component.

It also changes the print component to zoom to the closest defined _print scale_ instead of _map resolution_ as it probably more intuitive for the user (and easier to implement when calculating scale from resolution and vice versa) 

If the _print scales_ option is used on the print component and _constrainResolution_ is set to true, the resolution corresponding to the scale must exist in the _resolutions_ array, otherwise you still get the wrong scale.

If the map uses tiled services you probably need to set up tile grids that matches all corresponding print resolutions as well for each dpi. 

Maybe there should have been a default (90-ish) dpi option instead of 75 for those who really does not want to mess with the resolution/scale can of worms, but that is another issue...

### DIFF
--- a/src/controls/print/index.js
+++ b/src/controls/print/index.js
@@ -47,7 +47,7 @@ const Print = function Print(options = {}) {
     leftFooterText = '',
     filename,
     mapInteractionsActive = false,
-    recalculateResolutions = false
+    supressResolutionsRecalculation = false
   } = options;
   let {
     showNorthArrow = true
@@ -105,7 +105,7 @@ const Print = function Print(options = {}) {
         rotationStep,
         leftFooterText,
         mapInteractionsActive,
-        recalculateResolutions
+        supressResolutionsRecalculation
       });
       mapMenu = viewer.getControlByName('mapmenu');
       menuItem = mapMenu.MenuItem({
--- a/src/controls/print/index.js
+++ b/src/controls/print/index.js
@@ -47,7 +47,7 @@ const Print = function Print(options = {}) {
     leftFooterText = '',
     filename,
     mapInteractionsActive = false,
-    supressResultionsRecalculation = false
+    supressResolutionsRecalculation = false
   } = options;
   let {
     showNorthArrow = true
@@ -105,7 +105,7 @@ const Print = function Print(options = {}) {
         rotationStep,
         leftFooterText,
         mapInteractionsActive,
-        supressResultionsRecalculation
+        supressResolutionsRecalculation
       });
       mapMenu = viewer.getControlByName('mapmenu');
       menuItem = mapMenu.MenuItem({
--- a/src/controls/print/index.js
+++ b/src/controls/print/index.js
@@ -46,7 +46,8 @@ const Print = function Print(options = {}) {
     rotationStep = 1,
     leftFooterText = '',
     filename,
-    mapInteractionsActive = false
+    mapInteractionsActive = false,
+    supressResultionsRecalculation = false
   } = options;
   let {
     showNorthArrow = true
@@ -103,7 +104,8 @@ const Print = function Print(options = {}) {
         rotation,
         rotationStep,
         leftFooterText,
-        mapInteractionsActive
+        mapInteractionsActive,
+        supressResultionsRecalculation
       });
       mapMenu = viewer.getControlByName('mapmenu');
       menuItem = mapMenu.MenuItem({
--- a/src/controls/print/index.js
+++ b/src/controls/print/index.js
@@ -47,7 +47,7 @@ const Print = function Print(options = {}) {
     leftFooterText = '',
     filename,
     mapInteractionsActive = false,
-    supressResolutionsRecalculation = false
+    recalculateResolutions = false
   } = options;
   let {
     showNorthArrow = true
@@ -105,7 +105,7 @@ const Print = function Print(options = {}) {
         rotationStep,
         leftFooterText,
         mapInteractionsActive,
-        supressResolutionsRecalculation
+        recalculateResolutions
       });
       mapMenu = viewer.getControlByName('mapmenu');
       menuItem = mapMenu.MenuItem({
--- a/src/controls/print/print-component.js
+++ b/src/controls/print/print-component.js
@@ -1,6 +1,9 @@
 import olAttribution from 'ol/control/Attribution';
 import olScaleLine from 'ol/control/ScaleLine';
 import { getPointResolution } from 'ol/proj';
+import TileImage from 'ol/source/TileImage';
+import TileWMSSource from 'ol/source/TileWMS';
+import TileGrid from 'ol/tilegrid/TileGrid';
 import {
   Button, Component, cuid, dom
 } from '../../ui';
@@ -41,7 +44,7 @@ const PrintComponent = function PrintComponent(options = {}) {
     rotationStep,
     leftFooterText,
     mapInteractionsActive,
-    supressResolutionsRecalculation
+    recalculateResolutions
   } = options;
 
   let {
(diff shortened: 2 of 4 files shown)


Grammostola:
Hi, I tested a few scenarios with the default map (OSM and its gridset) and observed a couple of things:

> the scale selector and the print preview gets the same scale when constrainResolution is set regardless of DPI setting

This seems to be the case when constrainResolution for the map is false, which addresses the linked issue i.e Print - choose the largest scale in the scale selector, see that scale reflected in the scale bar in the print preview then choose 300DPI. The scale remains when supressResolutionsRecalculation is false in my somewhat limited tests.

The same happens when constrainResolution is true for the map ie the print preview / scalebar scale does not change at the largest scale when choosing 300DPI, though the selectable scales do not quite match the print preview's scale bar. They are closer than they are with supressResolutionsRecalculation : true however.

With constrainResolution: true for the map:

"supressResolutionsRecalculation" : true
dropdown: 1:8540 selected, print preview 1:6697
"supressResolutionsRecalculation" : false
dropdown: 1:8540 selected, print preview 1:8531

> If the print scales option is used on the print component and constrainResolution is set to true, the resolution corresponding to the scale must exist in the resolutions array, otherwise you still get the wrong scale.

If the print control is initiated with 
```
        "scales": [
          "1:69890000",
          "1:34940000",
          "1:17470000",
          "1:8736000",
          "1:4368000",
          "1:2184000",
          "1:1092000",
          "1:546000",
          "1:273000",
          "1:136500",
          "1:68250",
          "1:34120",
          "1:17060",
          "1:8531",
          "1:4265",
          "1:2133"
        ]
```
and constrainResolution is true then some chosen scales in the print scale selector match the print preview scale indication precisely like 1:4265  whereas 1:2184000 becomes 1:2183924.

Your proposition does appear to fix the primary issue described in the issue so that largest scale selected  - 300 DPI -  largest scale remains in the print preview and does not change to something else. Whether constrainResolution is true or not and making the print preview scale number always match the print scale selector when the map has prop constrainResolution:true might be a different issue or PR. Maybe the default gridset is problematic? One point of setting constrainResolution for a map would be to display the gridset's even scales. If there's no interjection I will try this with our gridset. 

Do tell if I've misunderstood something : )

Lints ok.

steff-o:
I agree, it is not a perfect solution. The result will probably depend on the gridset, the coordinate system and where in the world you are. The scale bar label seems to calculate the scale using the center of the view. By panning in the print preview the scale value will change even if the zoom has not changed. Also there are some rounding errors involved, some intentional, some as a result of projection and probably some just the result of the maths. The intentional rounding is a function that always round to closest 10 in the scale denominator at some point. Didn't mess with that as I didn't write that part myself.

It probably takes some time and thought to actually configure the gridsets to produce the desired result. But during my tests I managed to get same scale in selector and print preview in all cases. The only obvious error was when closing the print preview it returned to the closest resolution to the recalculated resolution, which would probably not be the scale set in the print preview or the scale that the map had before entering the preview.

My gridset was however small and only even 10:s. (1:10000, 1:5000 and 1:3000) for a Sweref grid. Didn't bother to actually set up the gridset in geoserver though, as we don't use tiles more than for the dynamic cache. When printing it is OK to create dynamic requests as they are few and the user will probably have more patience when waiting for a refresh in print preview.

One use case for setting constrainResolution is to get crisper images when using tiles as they are not interpolated from the server's images, but that would of course setting upp tiles for all DPIs as well.

Maybe it's enough for now, but I think a better solution would be creating a new map object and play with that as print preview. Then maybe it is possible to utilize the PixelRatio setting. (But that __is__ another issue)

steff-o:
I'll have someone at Övik try it out.

jokd:
@steff-o Have they tested it yet? We are about to make a release and it would be nice if this got merged before.

steff-o:
We have to investigate a problem with the extent that seems related to this fix. When switching to higher resolutions it seems like the extent is set smaller so not all of the map is displayed.

Grammostola:
Ok, hmm, I'm not seeing a difference in the extent change when switching between resolutions in the print panel with this fix enabled and not enabled (it shrinks a bit when choosing a lower resolution) with the default gridset, it might differ between gridsets.

steff-o:
I have conformed that there is indeed a problem with the extent when using tiled services (e.g. WMS using `renderMode='tile'`. The problem is that when the layer source is created it calculates max and min extents for the layer in terms of tiles. When the resolution is increased the grid extents is not recalculated to provide for the increased number of tiles that is required to fill the extent. The result is that tiles are not fetched if the tile index is outside the original set of tiles.

It is possible to update all tiled layers with a recalculated _TileGrid_, but it requires accessing properties in OL considered private.

A workaround would be switching to image-type in printmode as the source will probably not have a gridset that matches the print grid anyway, but that would probably only work or WMS.

Right now I'm not sure if it is worth the effort as high dpi prints as they are implemented now still has some big issues like symbols getting small and unreadable. I can do a best effort fix and make the feature otp-in instead.

steff-o:
It was a bit more complicated that just updating the extent of the tiles. The main issue was that the resolutions array for the map is shared by reference to all tiled layers that does not specify their own grid, which screwed up the _TileGrid_ for those layers.

To keep some predictability I implemented the entire function like this:
1. When the resolution (DPI) in print view is changed the map's resolutions array is recalculated to match the DPI setting. If not using a _scales_-configuration hopefully the scales dropdown and the scalebar will show the same scale even when `"constrainResolution": true`. If a _scales_-configuration is used it should contain scales that correspond to resolutions contained in the resolutions array if `"constrainResolution": true`. Otherwise it will still not match. Also the scale denominator must be dividable by 10, due to some rounding function in the resolution to scale calculation.
2. Tiled layers keep their original grid, regardless if it was shared with the map or not. (If it was shared a temporary deep copy is made). This is because otherwise the tiling scheme will not match the tiles of the service. OL will fetch the closest resolution and resample it to correct scale. This is consistent with the functionality before this PR for tiled layers using a separate grid from the map (e.g OSM which have a fixed grid)
3. WMS layers using `renderMode = 'tile'` will get a new temporary _TileGrid_ that reflects the new resolutions. This is because a tiled WMS is actually a dynamic WMS that can generate new tiles in whatever size. By utilizing that the tiles does not have to be resampled and will in theory be more crisp. If the WMS server is using a tile cache, it will cause cache misses on the server, but still work. 
4. The feature is now opt-in and is activated by setting `"recalculateResolutions": true` on the _print control_. Omitting the setting or setting it to false will only cost three _ifs_ and some lines of dead code.

We have tested against a map containing tiled WMS, image WMS, WFS, Group layers and OSM.

Grammostola:
Have tested with a map with image wms, tiled wms, wfs, geojson and osm and the default gridset, no scales array for the print control and `constrainResolution: false`

* Setting the print scale dropdown to the max scale and changing the DPI does not change the print scale bar scale (original issue fixed)
* Nothing disappeared or looked weird

w. `constrainResolution: true` and a local gridset with even scales
* Scales in the dropdown match the local gridset as well as the print scale bar scale, everywhere except for scales smaller than around 1:500 000 where the print scale bar scale shows 1:499 999.  Whether a scales array matching the local gridset and the map is entered for the print control or not.   This is a major upgrade for constrainResolution enthusiasts as with  `recalculateResolutions: false` the print scale bar never shows what is selected in the dropdown.

No lint errors. To me it looks good, except I'm not sure whether it should be opt-in or opt-out.

steff-o:
Changed it back to opt-out. Use `supressResolutionsRecalculation" : true` in order to turn it off. 

Also I removed an unnecessary deep copy of the tile grids when closing the print view.

MattiasSp:
I just want to say great job with this PR to everyone involved! :+1:

PR #1335, opened 2021-09-10
Adds a Content-Security-Policy

### LINKED ISSUE #1334: Add a Content-Security-Policy
To make the pages more secure and don't leak privacy information a Content-Security-Policy is good to have. Nice page to test your pages: [https://webbkoll.dataskydd.net/sv](url)

Add a Content-Security-Policy to the demo-page to act as a example of how to secure the page.

johnnyblasta:
Closes #1334

### DIFF
--- a/index.html
+++ b/index.html
@@ -4,6 +4,8 @@
 	<meta charset="utf-8">
 	<meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=no">
 	<meta http-equiv="X-UA-Compatible" content="IE=Edge;chrome=1">
+	<meta http-equiv="Content-Security-Policy" content="default-src 'none'; img-src *; script-src 'self' 'sha256-O61tfZp/aDhcL8nyV1f8HcikmNvkoXIOADVYHxC57PA=' 'unsafe-eval'; connect-src 'self'; style-src 'unsafe-inline'; style-src-elem 'self'; base-uri 'self'; form-action 'self';">
+	<meta name="referrer" content="same-origin">
 	<title>Origo exempel</title>
 	<link href="css/style.css" rel="stylesheet">
 	<link rel="shortcut icon" href="img/png/logo.png">


jokd (approved):

jokd:
Tried on our web map and got this:
Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' 'sha256-O61tfZp/aDhcL8nyV1f8HcikmNvkoXIOADVYHxC57PA=' 'unsafe-eval'". Either the 'unsafe-inline' keyword, a hash ('sha256-5GTBAHUtU9TO/CzM287r+9RLjA2U1yINKIMC715Z23M='), or a nonce ('nonce-...') is required to enable inline execution.

Do I have to create a new hash for each build?

jokd:
How about using 
`script-src 'self' 'unsafe-eval' 'nonce-0rigo'` instead and then

```
<script type="text/javascript" nonce="0rigo">
	var origo = Origo(...............
```

but perhaps that is less secure. I'm afraid hashing the build will make lots of applications not work as well as issues keeping index.html in sync with the build. In production, adding the hash can be a recommendation.

jokd:
Found another issue; when changing the opacity of a layer:
image

johnnyblasta:
The hash isn't for the Origo script. It's for the inline script that starts Origo. You could use 'unsafe-inline' instead of a hash, but the script that starts Origo doesn't change that often, so it isn't to much bother changing the hash for the hash which is shown in the error message.

jokd:
Ok. But the issue with the event handling is something else,

johnnyblasta:
> Ok. But the issue with the event handling is something else,

 'unsafe-inline' and removing the hash solves this, but much of the purpose of the CSP is gone then :(

jokd:
Yes, better to rewrite code where necessary but that may take some time. Maybe should have 'unsafe-inline' until that work has been done?

johnnyblasta:
Mad a new PR that adds the 'unsafe-inline' #1339 since this was already merged.

tonnyandersson:
@johnnyblasta @jokd Yo, I noticed something when using the old [Origo 2.2.0 test config](https://www.malardalskartan.se/origo/v2.2.0/test.json):

image

Same issue for all vector type layers, i.e. AGS_FEATURE, WFS and external GEOJSON, TOPOJSON. Tested with a different config as well, same result.

Is this an actual issue or am I missing something?

jokd:
We are creating our own CORS-issues :)

jokd:
You would have to use  `connect-src 'self' kartor.vast......se` in order to get data from another domain. wms services are images so they are allowed from anywhere since `img-src *`

tonnyandersson:
Ah. A classic case of RTFM then.

jokd:
Usually don't go that far but yes indeed.

johnnyblasta:
This Content-Security-Policy should be considered as a good starting point, there is often the case it has to be modified when added to your local environment.

PR #1336, opened 2021-09-14
Adds segment label toggle

johnnyblasta:
Adds a button for toggling the segment labels on measurements.
Discussed solutions in #1035 

SegmentLabelToggle


johnnyblasta:
Made this as a toggle function, but perhaps should have been a on/off switch?
See example below when one segment the labels are turned off and then continue on another segment the the labels will be toggled oposite the other segment, maybe a bit confusing.

SegmentLabelToggle2

jokd:
I think it will confuse people. On/off is probably preferable, though it is a nice feature when you understand how it works.

johnnyblasta:
> I think it will confuse people. On/off is probably preferable, though it is a nice feature when you understand how it works.

Less confusion is good. 
I'll rework it after some more time gone by, so that more can have an opinion.

johnnyblasta:
Changed to on/off type button.
Also added a config option `"showSegmentLabelButtonActive": false` default to true so button is active on start unless set to false in config.

SegmentLabelToggle3

jokd (approved):

PR #1340, opened 2021-09-16
Getcap padlock

### LINKED ISSUE #1341: Support for workspace:layername syntax in secured layer check
When using workspace:layername syntax the secured layer check won't work and this is a small fix for that.

jokd:
Fixes #1341 
When using workspace:layername syntax the secured layer check won't work and this is a small fix for that.


jokd:
@asemoller OK with you if I merge this? Can't see why it would mess anything up.

PR #1362, opened 2021-10-06
Adds focus style on buttons and other that should have focus

### LINKED ISSUE #1361: No focus style
**Describe the bug**
There seems to be no focus style associated with buttons and other things that should be able to draw focus and therefore there is no way to determine where the focus is on the map when using tab to interact with the map.

**To Reproduce**
1. Go to a Origo map
2. Press tab on keyboard
3. Nothing seems to happen
4. After 20-30 press on tab you'll get focus on adress in the webbrowser since all the interactions in the map has been visited

**Expected behavior**
A clearly visible change of focus should appear when tabbing.

**Screenshots**
Fix with clearly visible focus change:
TabbFokus

johnnyblasta:
Closes #1361 
Adds focus style on buttons either change in color and a black border and for smaller icon buttons only a black border. 
Focus style on position control. 
Focus style on links in footer and attribution.
TabbFokus2
There is a problem with the legend when legends get focus, but that was there before so a separete issue.
There is also a problem when minimized the buttons stil gets focused, but it isn't visible and therefore you don't now where the focus is.
Haven't tested with all possible controls added.

### DIFF
--- a/src/controls/print/print-settings.js
+++ b/src/controls/print/print-settings.js
@@ -61,6 +61,26 @@ const PrintSettings = function PrintSettings(options = {}) {
   let rotationControl;
   let setScaleControl;
 
+  // Set tabindex for all settings buttons to include or exclude in taborder depending on if expanded or not
+  const setTabIndex = function setTabIndex() {
+    let idx = -1;
+    if (openButton.getState() === 'hidden') {
+      idx = 0;
+      document.getElementById(closeButton.getId()).focus();
+    } else {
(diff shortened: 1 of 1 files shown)


johnnyblasta:
Updated with hidden buttons which aren't visible from tab order. 
Also removed the sidenav button from tab order which was causing the legend to flip out of view.
TabbFokus3

johnnyblasta:
Also fixed focus for print control

PR #1372, opened 2021-10-19
#1314 Attachments

### LINKED ISSUE #1314: Add attachments i editor
Currently there is no support for uploading arbitrary attachments in the editor. FeatureInfo can however display links to attachments or emedd images from links. But creating the attribute must be done either manually as a text attribute, or using another application/custom dialogs.

In the editor I would like to have an attribute type for attachments. When editing it would be possible to add and remove attachments. When adding an attachment a file picker should open and after selecting a file it should be uploaded to the server and stored as an attachment It should be possible to add any number of attachments of any type of file.

In featureInfo all attachments should be displayed as clickable links and make the bowser decide what to do using the mime type. If the attachment is an image it should be possible to display it as todays _img _ option.

For images there already is an attribute type for uplading images, but that is limited to one image per attribute and stores the image as a base64 string in the feature itself, which is very bad for WFS performance. Displaying the image in feature info would be possible using the _urlPrefix_ attribute to set src to _data_/image/jpg;base64_ or appropriate mime type. To support general attchments a more complex expression must be used to turn the embedded attachment into a clickable link. That could possibly be accomplished using the _html_ option, but that would be tedious work for application developers and error prone.

FeatureInfo has an (undocumented) option to split an attribute by a separator character to create multiple links or images from one attribute, however that still requires some way of actually uploading files and link them to the feature.

All of this could be implemented outside origo using various techniquies, e.g by using the html option for display and custom dialogs for uploading, but that's once again tedious and error prone.

**Additional context**
As WFS is not really suited for embedding large binary objects as attributes and to support any number of attachments without having to resort to use a delimeter or xml/json datatype I suggest implementing the feature using an attachments REST service. The rest endpoint should support
- GET on a collection to retreive a list of attachments for a given feature using its private key
- POST on a collection to add a new attachment
- GET on a resource to retreive the actual file
- DELETE on a resource to delete the file

To make use of the attachment service some new options must be added to featureinfo and editor attributes.

The rest endpoint could probably be implemented in origo-server, but to keep it flexible regarding access control and data storage I suggest that the feature is only implemented as an interface dependency in origo. That makes it possible for each organistaion to implement storage as files or database and using different databases and behaviour. 

To avoid out of sync problems that could arise if two services access the same data or when two transactions must be kept in sync I suggest that links are only created on the attachment by storing its parent's private key (FID), just like normal 1-to-many relations in a database. That means that featureinfo must query the attachments service as there will be no attribute on the feature itself.

steff-o:
Closes #1314 

Intoduces the concept of attachments. Attachments are files that are not a part of a feature, but are related to a feature.  The implementation only includes the client side part of the attachment handling. In order to actually use attachments an attachment server is needed. The attachment handling is pretty much like in ArcGis and the server interface is actually taken from ArcGis server specification.

This PR consists of several parts:
	1. Code that comprises an attachment server client repository class.
	2. Code to include insertion and deletions in the attributes editor
	3. Code to create temporary attributes from the attachments links
	4. A new attribute formatter "linktext"
	5. As a bonus, a utility function groupBy is added.

**Configuration**
Configuration is performed by adding an _attachments_ field to a layer.
```
"attachments" :{
	"url"
	"format" 
	"foreignKey" 
	"stripLayerNameFromId" 
	layerId
	groups [{
		name
		title
		linkAttribute
		fileNameAttribute
		allowedFiles
	}]
}
```

**Configuration explained**
_url_ : The base adress for the attachment server e.g. "http://localhost:1337Myserver"
_format_: "origo" or "arcgis". Defaults to origo
_foreignKey_: Name of the attribute in the parent feature that contains the private key. If omitted feature.getId() i used
_stripLayerNameFromId_: removes layer name from featureid. Defaults to true. Useful when using geoserver wfs services to get only database FID as private key. (Just splits on ".")
_layerId_: The name of the layer as perceived by the attachment server, Defaults to layer name if omitted.
_groups_: Array of groups of attachemnts. A group has its own title and each group is treated as a separate attribute. At least one group has to be specified. If format is "arcgis" exactly one group must be specified.
_name_: Name of the group as perceived by the attachment server. Ignored for format = "arcgis"
_title_: Title displayed in the editor. Ignored for format = "arcgis"
_linkAttribute_: Name of an attribute to create on a feature containing all links as a semicolon separated list. Optional.
_fileNameAttribute_: Name of an attribute to create on a feature containing all filename as a semicolon separated list. Order is same as linkAttribute list. Optional.
_allowedFiles_: comma separated list of valid file extensionas and mime types that are allowed to upload. Format is accordning to the accept attribute of <input type = "file">  https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file

**Usage**
**Edit attachments**
Once a layer is configured using the above configuration, attachments can be viewed, added and removed from a feature using the normal attributes edit form. Due to the "out-of-band" nature of attachments all operations occurs immediately and the feature must exist on the server before adding attachments. If using autoSave that is not a problem.

There are plenty room for improvements here. For example adding support for synchronizing the attachments operations with the save operation and allowing add attachments to newly created features, but that would require a pretty complex state handling so I skipped it. Also it would be possible to implement different strategies on how to handle feature refresh if the relation between attachments and features in the database is known. For instance if the attachment server actually updates the feature with a child relation, a refresh of the feature would be needed, or a field could be updated by the editor. But this implementation only supports a foreign key relation from the attachment. Technically, the attachments form does not have to reside in the edit form, or even require an edit session, but to reuse as much as possible I put it there. It could be a separate tool in the editor or a separate control.

**Display attachments in info window**
If the linkAttribute configuration is specified a temporary attribute is added to each feature when the feature info is displayed. This makes it possible to display attachments as normal attributes by configure an attribute that uses the temporary attribute. If the fileNameAttribute configuration is specified the link text can be set using the attribute option linktext.

Assuming the attributes configuration contains a group entry containing "linkAttribute": "extra" and "fileNameAttribute": "extra_links", the list of links can be displayed by configuring an attribute for it.
```
{
          "title": "Extra",
          "linktext": "extra_links",
          "url": "extra",
          "splitter": ";"
 }
```

If configured in this way, the feature info sends a request to the attachment server each time a feature is displayed. It works regardless of which infoWindow is used (ovelay, sidebar, infoWindow). It also possible to configure attachments to be displayed for WMS-services, although it is not possible to edit attachments as that is performed in edit mode.

If the attachment server updates a real attribute with the links, there is no need for configuring the creation of the temporary attributes. However, as the feature is not automatically refreshed when attachments change it is not a good idea to create an application that uses both editing and feature info

**The attachment server**
The server should implement the interface of the arcgis server rest api:
_List attachments_: https://developers.arcgis.com/rest/services-reference/enterprise/attachment-infos-feature-service-.htm
_Get attachment_: https://developers.arcgis.com/rest/services-reference/enterprise/attachment-feature-service-.htm (Documentation says both returns an attachemntInfo object and streams the file. As I have no arcgis server to test against, I assumed that what you get is actually just the file downloaded as its mimetype as that would be the most appropriate thing to do)
_Add attachment_: https://developers.arcgis.com/rest/services-reference/enterprise/add-attachment.htm 
_Delete attachment_: https://developers.arcgis.com/rest/services-reference/enterprise/delete-attachments.htm
Update is not used.
I have not tested against a arcgis server as I don't have one.

If mode is set to "origo" the server should support additional functionality:
1. The attachmentInfo object should contain another field called "group", containing the name of the group the attachment belongs to
2. When adding an attachment the server should support another post variable called "group" containing the name of the group the attachment will belong to.

Reference implementation
There is a zero configuration reference implementation for test purposes here: https://github.com/ornskoldsvikskommun/attachmentServer-reference

**Security considerations**
The interface does not specify och explicitly support any authentication or authorization mechanisms. Instead it relies on the principle of the browser handling user sessions if the attachment server is deployed to the same server (as visible to the client) as the application. By deploying the attachment server to the same server as the application will make the browser send session cookies and basic authentication headers from the application that can be used for authorization. Another option for intranets would be using windows integrated authentication in the attachment server. To separate public access and internal use, two separate instances of the attachment server using different configuration but accessing the same data can be used, or use a server that updates the feature with a pre-cooked attribute for the public application.

### DIFF
--- a/src/controls/editor/attachmentsform.js
+++ b/src/controls/editor/attachmentsform.js
@@ -1,97 +1,89 @@
+import createElement from '../../ui/dom/createelement';
 import attachmentclient from '../../utils/attachmentclient';
 
 /**
  * Creates complete HTML for the attachment edit form including eventhandlers and what not. Asynchronously populates its container when result is received from the server
  * @param {any} layer The layer to create form for
  * @param {any} feature The feature to create form for
- * @param {any} containerId DOM id for a tag that this function will set the innerHTML to.
+ * @param {any} el DOM element that this function will set the innerHTML to.
  * @returns A Promise when resolved returns nothing at all really.
  */
-const attachmentsform = function attachmentsform(layer, feature, containerId) {
-  let str = '<h4>Bilagor</h4><br/>';
-
-  // Create the repo. No network traffic yet
+const attachmentsform = function attachmentsform(layer, feature, el) {
+  // Create the repo. No network traffic yet it just reads config
   const ac = attachmentclient(layer);
   const groups = ac.getGroups();
 
+  // Shortest code to clear a node
+  // eslint-disable-next-line no-param-reassign
+  el.innerHTML = '';
+  el.appendChild(createElement('label', `${layer.get('attachments').formTitle || 'Bilagor'}`));
+  el.appendChild(document.createElement('br'));
+  el.appendChild(createElement('p', 'Hämtar ...'));
+
   // Get from repo
-  return ac.getAttachments(feature).then(data => {
+  ac.getAttachments(feature).then(data => {
+    el.removeChild(el.lastChild);
     groups.forEach(group => {
       if (group.title) {
-        str += `<b>${group.title}</b>`;
+        el.appendChild(createElement('b', `${group.title}`));
       }
+
       if (data.has(group.name)) {
         const value = data.get(group.name);
         value.forEach(link => {
           // Display the link and add a delete button
-          str += `<div style="display: flex;align-items: center;">
-                    <a style="flex-grow: 1;" href="${link.url}" target="_blank">${link.filename}</a>
-                    <button class="o-button-lg" id="o-attachid-${link.id}"><svg class="o-icon-24"><use xlink:href="#ic_delete_24px"></use></svg></button>
-                  </div>`;
-        });
-      }
-      // Add an add button. Do some silly css stuff to hide the actual file input as it is very ugly and use the label instead
-      // and even stupidier css stuff to make the button inside the label not act like a button. But it's there to steal the button style
-      let acceptstring = '';
-      if (group.allowedFiles) {
-        acceptstring = `accept="${group.allowedFiles}"`;
-      }
-      str += `<div>
-                <label  for="o-attach-${group.name}" style= "cursor: pointer;display: block;">
-                  <button class="o-button-lg" style="pointer-events: none;">
-                    <svg class="o-icon-24"><use xlink:href="#ic_add_24px"></use></svg>
-                  </button>
-                </label>
-                <input type="file" id="o-attach-${group.name}" ${acceptstring} style="opacity: 0; width: 0; height:0; padding: 0; border:0;"/>
-              </div>`;
-    });
-
-    // Set the content on our container. It just happens to acually create the DOM objects as well
-    document.getElementById(containerId).innerHTML = str;
-    // Elements exist now, add handlers.
-    // Maybe build content using Element instead to make it easier to attach handlers
-    groups.forEach(currGroup => {
-      const key = currGroup.name;
-
-      // Add an add event listener for each group
-      const elm = document.getElementById(`o-attach-${key}`);
-      elm.addEventListener('change', ev => {
-        ac.addAttachment(feature, ev.target.files[0], key)
-          .then(() => {
-            // Kinda looks like it is recursive, but it's not. It just schedules a promise to refresh the from content
-            attachmentsform(layer, feature, containerId);
-          })
-          .catch(err => {
-            const errorMessage = `<p>Misslyckades med att spara: ${err}</p>`;
-            document.getElementById(containerId).innerHTML = errorMessage;
-          });
-      });
-      // Only look for remove buttons in groups that actually have attachments
-      if (data.has(key)) {
-        const links = data.get(key);
-        // Add a delete eventhandler for each attachment
-        links.forEach(link => {
-          const deleteButtonElm = document.getElementById(`o-attachid-${link.id}`);
-          deleteButtonElm.addEventListener('click', () => {
+          const rowEl = createElement('div', `<a class="grow" href="${link.url}" target="_blank">${link.filename}</a>`, { cls: 'flex row padding-x padding-y-smaller' });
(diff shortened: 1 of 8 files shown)


tonnyandersson:
Only tested against ArcGIS Server and it works like a charm!

A few suggestions, though. Feel free to argue. :)

- Make attachmentsform title configurable ('Bilagor' may be set as default).
- Change attachmentsform title element to `<label>` instead of `<h4>` for conformity. For the record, I do like the `<h4>` style better, but it doesn't work well with the rest of the edit form.
- Generally, don't use headings. Use element tags already in the edit form. Major deviations from the general UI style is frowned upon.
- IMO, it looks better with smaller buttons, like the rest of the UI (not including editor controls).
- Add some vertical space between buttons.
- Move Add button to the right, in line with the Delete buttons. May be unwise from a usability standpoint, but I think it looks better. Or change it into a text button, like 'Add attachment'.

Before:
attachmentsform_before

After:
attachmentsform_after

- Perhaps an unnecessary console.log() on line 1002 in edithandler.js? Not sure if anything other than `undefined` will ever be logged there.
- Needs to be documented thoroughly, with examples. The configuration is not 100% intuitive.



Other considerations, perhaps for future improvements:
- Implement splitter by default. If a layer has a maximum of ONE attachment per feature, using the attachments attribute type is a bit overkill.

steff-o:
Of course you are right @tonnyandersson. I have updated now. Visual representation just isn't my strong side, but I have tried to steal the look from other menus using smaller icons and adding tool-tips. Also made section title configurable.

While at it I also reworked the form to not depend on id-attribute and getElementById for setting up eventhandlers.

tonnyandersson:
@steff-o For once, I remembered to run a test on an actual mobile device as well... It turns out the Add button is not showing in IOS Safari. It seems to be the "hidden" input element that is causing this. Setting element `width` and `height` to `1px` instead of `0` seems to do the trick.

Other than that, I think this is good to go. Just a few visual things that still bug me, but those are not caused by this PR.

So, just fix the IOS thing and I'll approve.

steff-o:
Input is now 1x1 px. Also made tooltip visible for add button. It didn't show as mouse events were disabled.

tonnyandersson:
LGTM.

tonnyandersson (approved):

PR #1381, opened 2021-11-08
Search function for legend control

### LINKED ISSUE #1330: Legend search
With a legend (or layer list) that contains many layers, it can be difficult for the user to find the right layer in the list. It should be possible to search for layers (and groups?) in the legend. 

One solution is to have a search function in the legend. Another solution could be to improve the search control so that it also includes layers which, when clicked in the results list, make them visible in the map.

johnnyblasta:
Closes #1330 

Introduces 3 new configurations options:
```
{
  "name": "legend",
  "options": {
      "searchLayersControl": true,
      "searchLayersMinLength": 2,
      "searchLayersLimit": 10
  }
}
```

### DIFF
--- a/scss/externs/_awesomplete.base.scss
+++ b/scss/externs/_awesomplete.base.scss
@@ -32,6 +32,21 @@
   z-index: 1;
 }
 
+.awesomplete.black > ul {
+  background: #000 !important;
+  box-sizing: border-box;
+  color: #fff;
+  left: 0px;
+  list-style: none;
+  margin: 0;
+  min-width: 100%;
+  padding: 0;
+  position: absolute;
+  top: auto;
+  bottom: 100%;
+  z-index: 9999999;
+}
+
 .awesomplete > ul:empty {
   display: none;
 }
--- a/scss/externs/_awesomplete.theme.scss
+++ b/scss/externs/_awesomplete.theme.scss
@@ -69,8 +69,16 @@
   font-weight: bold;
 }
 
+.black mark {
+  color: #aaa;
+}
+
 .awesomplete li .suggestion:hover mark {
   background-color: $search-hover-backround;
+}
+
+.black li .suggestion:hover mark {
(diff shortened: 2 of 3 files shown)


jokd:
I get a list but nothing seems to happen when I click one of the results?
Shall layers not visible in the legend be excluded from search results? When searching origo in default config i get both Origokommuner and origo-mask (who belongs to group 'none').

johnnyblasta:
> I get a list but nothing seems to happen when I click one of the results? Shall layers not visible in the legend be excluded from search results? When searching origo in default config i get both Origokommuner and origo-mask (who belongs to group 'none').

If layer is already visible nothing happens when clicked. Should it be toggled?
As it is now it searches on layer name and title, should it also search on abstract?
What is the opinion on whether certain layers should be excluded?
The search function makes most sense when the legend contains a lot of layers in multiple groups.

jokd:
I think I expected a toggle or zoom-to-layer-in legend-function but maybe just turning on is enough. Or if it in the search results was indicated whether its turned on or off, then a toggle function would be the best. Otherwise you must find the layer in the legend to be able to turn it off if that's what you want....
Also inclulde abstract is not a bad idea, like a light metadata search.
I think that layers that are not visible in the legend should be excluded.

johnnyblasta:
Added the possibility to search on abstract for the layer and make it configurable, default is only to search in name and title. 
Excluded layers that belongs to group none in the search.

```
    {
      "name": "legend",
      "options": {
        "searchLayersControl": true,
        "searchLayersMinLength": 2,
        "searchLayersLimit": 10,
        "searchLayersParameters": ["name", "title", "abstract"],
      }
    }
```

johnnyblasta:
I've had plans to include a zoom-to-layer-in legend-function but haven't had time to do it so thought it might be best to release this now and improve it later.

jokd (approved):
I think this is good to merge

Issue #1108, opened 2021-01-20
Error with the carousel


jokd:
**Describe the bug**
When emulating a touch screen in Edge, the carousel renders an error. Works as expected though

**To Reproduce**
Click on at least two items and use the arrows in the carousel. An error is shown in the console.

**Screenshots**
![image](https://user-images.githubusercontent.com/6848075/105142969-833c7280-5afb-11eb-9255-dbd13d5e0184.png)

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Edge touch emulation mode

jokd:
Turns out it is a known bug in GlideJS and probably solved in the master branch but not the latest release.

jokd:
Fixed with #1392

PR #1392, opened 2021-11-26
Update Glide js

jokd:
Solves #1108
Tested with default config as well as @johnnyblasta 400 toppar and seems to work well.


johnnyblasta (approved):
Tested and LGTM

PR #1393, opened 2021-11-29
Webpack 5

### LINKED ISSUE #1355: Webpack 5
It's about time we migrate to webpack 5. That would probably ease future version bumps of dependencies as well as get rid of a couple of vulnerabilities.

jokd:
Fixes #1355 
Also removes Babel so testing it in different setups is appreciated.


jokd:
Tested and the build works with default config in:
Chrome/Edge v.96
Firefox v.89
Safari in iOS 15.1

Does NOT work in:
Internet Explorer 11

tonnyandersson:
Do we still care about IE11 at all?

johnnyblasta:
> Do we still care about IE11 at all? If we do, [this](https://webpack.js.org/migrate/5/#need-to-support-an-older-browser-like-ie-11) is supposed to fix that, but all I can see from a super quick test with `target: ['web', 'es5']` in the dev environment is that it comes with a different set of issues. Works fine in Chrome, though.

I checked my webstats for november and it was 1,2 % that still used IE. Going down but not eradicated yet.

tonnyandersson:
IIRC, we decided to drop IE11 support a while back. So there's that.

Sadly, I don't have any cool, avantgarde webpack setups to test with, nor can I come up with a credible real-world example. But I haven't found any issues other than the ones mentioned. So, LGTM.

Oh. Perhaps we no longer need the root/babel.config.js file if Babel is out?

jokd:
For that ~1.2% with unsupported browsers we can add a text saying that the map requires a modern browser?

johnnyblasta:
> For that ~1.2% with unsupported browsers we can add a text saying that the map requires a modern browser?

It was a steady declining curve, a couple of months before it was 3%, so hopefully it's gone completely in another months. Don't know if a text is necessary, since they should already know that things aren't working so good on IE nowadays just by surfing a bit. 😵‍💫

tonnyandersson:
Getting a _bunch_ of these when loading map from builds. The screenshot is from a build-dev build. The build build only produces one line. Please note that this only occurs when loading the unminified origo.js file. The minified origo.min.js does not produce any errors.

Is this expected? Source maps are not my strong suit.

steff-o:
Not sure what the intention of copy.js and copy-dev,js is, but the missing source maps that @tonnyandersson  is experiencing is because:
1. The mode in both files is 'development', which implies that source maps is created in 'eval' mode unless otherwise specified.
2. The 'devtool' setting is missing in copy-dev.js, which means that devtools option defaults to 'eval', which seems to not generate source maps for OL.
3. The 'devtool' setting is `false` in copy.js, which turns off source maps all together.

How to fix this depends on what the intention of the differnent files are. 

Assuming that the intention of copy.js is to create a bundled file that is not minified for testing without source maps the settings should be fine. When running through devserver it however seems to expect source maps.

Assuming that the intention of copy-dev.js is to produce a bundled file for testing with source maps just like when running the dynamic output of devserver, another source map strategy should probably be chosen that produces valid soure maps for OL as well, e.g using the same option like dev.js: `devtool: 'eval-cheap-source-map'`.

One thing that I don't understand is why copy.js and copy-dev.js copies the origo.min.js file from dist/ to js/ 

This PR updates the rules for lint to airbnb 15.0.0, which makes the existing code fail lint test.

tonnyandersson:
@steff-o The only difference between copy and copy-dev is that the latter exports build files to a _build_dev_ folder, which is ignored by git. This gives you a way to create builds without git registering the build files as changes.

jokd:
I think this is ready for merge, can someone please have another look?

johnnyblasta:
Tested and got this when running the build
image
Otherwise it looks good.

jokd:
Yeah I saw that but I think we just ignore that. That js file shouldn't be used that often anyway.

PR #1395, opened 2021-11-30
Tool for showing buffer around point

### LINKED ISSUE #1386: Show buffer around point
It would be useful to be able to show a buffer around a point and exactly specify the radius of the buffer.

I'm thinking it would be a tool in measure control and click on a point and ask for a radius and the buffer is drawn.

A buffer around other geometries such has lines could also be useful.

johnnyblasta:
Closes #1386 
Example animation:
BufferPromptNew
Config code:
```
      "default": "buffer",
			"measureTools": [
			  "area",
			  "length",
			  "buffer"
			]
```

There is also a fix for bug for selecting a default tool in config which was not working.


tonnyandersson:
Drop the "Radie" part of the label. Or make it configurable, with an empty string as default. It would also look so much cooler if the label was curved along the edge of the buffer. Just my opinion.

johnnyblasta:
I tested to set the text along the curve earlier, but didn't get it to work, so did it like this instead.

tonnyandersson:
It's quite possible to achieve. Just make a polygon from the circle geometry, like `Polygon.fromCircle(circleGeometry);` and use the polygon as geometry for the label. With text `placement` set to `line`, the label will follow the polygon edge. It's not perfect and it takes som fiddlin' to make it look right, but it can be done.

However, if you just drop the "Radie` prefix, the label will be short enough to look okay without the curvature in most cases.

johnnyblasta:
Made a test with the text on the circle, good enough or weird? @tonnyandersson 
BufferPromptCircleText2

tonnyandersson:
Well, what do you think?

johnnyblasta:
> Well, what do you think?

Kind of cool, but looks a bit strange in some cases. Special when zooming and the text skips around.

steff-o:
It would look a lot nicer if the dialog wasn't a window.prompt(). But maybe its not worth the extra work and additional code when prompt() is there and works.

johnnyblasta:
> It would look a lot nicer if the dialog wasn't a window.prompt(). But maybe its not worth the extra work and additional code when prompt() is there and works.

Yes it would be nicer without window.prompt, but it is such a simple solution that only requirs a single line of code.
If there was several use cases that needs a prompt function, then it would be good idea to do a general util prompt function. As of now this is the only use of prompt in Origo what I can see.

johnnyblasta:
Rework with modal instead of window.prompt:
BufferPromptCircleText3

asemoller:
Ah, looks like multiselect buffer tool! 😃

johnnyblasta:
@asemoller yes that worked nice in this case, thanks for the hint :)
With this last commit I feel satisfied with this PR.

src/controls/measure.js
```diff
@@ -65,6 +68,8 @@ const Measure = function Measure({
   let measureButton;
   let lengthToolButton;
   let areaToolButton;
+  let bufferToolButton;
+  let bufferSize = 1000;
```
tonnyandersson, line 72: No need to set a default value if the default value is never used.

src/controls/measure.js
```diff
@@ -399,6 +437,40 @@ const Measure = function Measure({
...
+    });
+    const bufferradiusEl = document.getElementById('bufferradius');
+    bufferradiusEl.focus();
+    const bufferradiusBtn = document.getElementById('bufferradiusBtn');
+    bufferradiusBtn.addEventListener('click', (e) => {
+      const radiusVal = bufferradiusEl.value;
+      const radius = parseFloat(radiusVal);
+      if ((!radius && radius !== 0)
+        || (radius <= 0)) {
+        bufferradiusEl.classList.add('unvalidValue');
```
tonnyandersson, line 460: Misspelling. Should be '**in**validValue'.

Also, can't see the point of adding this class since it's not used anywhere else, nor is there any CSS stuff implemented for it. Is this a "might come in handy down the line" type thing?

src/controls/measure.js
johnnyblasta: That was a remnant from before reworking code. Good catch.

src/controls/measure.js
johnnyblasta: Don't remember what the thinking was here so I'll remove it.

tonnyandersson (approved):

PR #1396, opened 2021-11-30
Changed to focus-visible

### LINKED ISSUE #1394: Buttons are incorrectly highlighted
**Describe the bug**
There is a black circle around the buttons when you click on them. And when you close a tool, the button is marked incorrectly.

**To Reproduce**
1. Download latest Origo
2. Click on any button
3. See error

 - Browser [chrome, edge]
 
marker_icons

johnnyblasta:
Fixes #1394 
Changed to focus-visible from focus to only display clear focus when using keyboard navigation and not on mouse clicks. 
One problem is that focus-visible is a new feature in web browser and only got adopted as of 2021. Thoose with old browser will not see clearly focused buttons when tabbing.

### DIFF
--- a/scss/_bookmark.scss
+++ b/scss/_bookmark.scss
@@ -1,27 +1,55 @@
-.o-bookmark-button:focus-visible {
+.o-bookmark-button:focus {
   background-color: $grey-lighter;
 }
 
-.o-bookmark-button.light:focus-visible {
+.o-bookmark-button.light:focus {
   background-color: $grey-lighter;
 }
 
-.o-bookmark-button.primary:focus-visible {
+.o-bookmark-button.primary:focus {
   background-color: $primary-color-dark;
 }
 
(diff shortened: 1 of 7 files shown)


asemoller:
Wow, that was really fast 👍 and it works fine!
Good work @johnnyblasta and thanks!

johnnyblasta:
Doesn't work in Safari and other web browser older than about a year. Is that a big problem? Could do a css hack which adds visible focus for those browser that don't support `focus-visible`, but then thoose would have the undesired behaovior that shows focus even on clicking.

jokd (approved):

jokd:
Excellent!

johnnyblasta:
Made some test on Safari and older browser, seeing focus and LGTM.

jokd:
How about IE?






Kidding, you can merge this.

PR #1398, opened 2021-12-02
Related tables

### LINKED ISSUE #1315: Edit related objects
Sometimes data for an object is distributed over serveral tables in the database. When editing attributes, origo has no perception of when objects from different layers have a realtion in the database. It prevents a data model to have 1-to-many relations if they are to be edited in origo or forces the user to manually edit foreign keys attributes or use a flattened data model which requires a fixed maximum number of columns.

In featureInfo related tables should be displayed as an attribute with an expandable tree to show the attributes from the related table. This would be recursive so that a related table can have related tables, and so on.

When editing, related tables should be editable from the parent's attribut editor by the means of an expanding tree or similar. It should also be possible to add and remove new instances in the child layer.

Flatten the datamodel. But that does not scale well for multi-level relations and child tables with many attributes. Also it would not look appealing.

**Additional context**
This feature would resemble the functionality in ArcMap which has the capability to follow relations. In the ESRI-world it does that from defined relations in the database. But since we're using WFS that has no perception of relations, this feature would require manually configuring origo with the foreign key relation.

### LINKED ISSUE #1391: Error in attachment error message
The error message below is displayed whenever there's a featureinfo request promise error, even if the layer in question is not configured for attachments at all.

This error message should only be triggered if the promise error is related to attachments.

image

I this this is the culprit, alerting the world about attachments regardless of what the catch of the day is.

https://github.com/origo-map/origo/blob/73b4f93748e01a3d3cea3a470fb21083601439e9/src/featureinfo.js#L412-L420

Ping @steff-o :)

steff-o:
Resolves #1315 by adding the concept of related layers and fixes #1391 by sheer luck.

Implementation is limited to provide the basic use case to _edit attributes in a related table when the base table is considered read only because it is owned by another party_. It can also display information from the related table in the featureInfo window by aggregating attributes from all children to a temporary attribute on the parent.

The feature is only implemented for WFS. Mainly because we don't use AGS and I don't have access to an AGS. If implementing for AGS it would be possible to fetch the configuration from AGS as AGS already have a concept of relations in the database which is exposed through the rest interface.

__Added Functionality__
By configuration it is possible to add a 1-n relationship between a parent layer and a child layer. It can be done recursively to support any depth.

image

Editing the child layer attributes is performed from the parent's attribute editor. It is possible to limit which operations are allowed on each child layer. In the example above it is possible to edit, delete and add "Bilder" as entries in the related table "metadata", but the layer itself is not configured as editable. It is also possible to edit the entry "Ansvarig". It is not possible to add more entires to "Ansvarig" as it is configured for max 1 entry. The text for each child entry is either the _fid_ ( as in metadata example) or a configured column (as in ansvarig example)

If clicking the plus sign under "Bilder" a new entry is created (without geometry) in the child table and if the editor is configured with `"autoSave" = true` the new entry is saved. If the editor is configured with `"autoForm" = true` the attribute editor form for the newly created is displayed.

If clicking the edit pen on "Stefan", the attribute editor for the "Stefan" record is opened and can be edited. When closing that form the user returns to the parent attribute form ("Byggnad"):

image

By configuring the attribute aggregation, the information from the related child objects can be displayed in the infowindow:
image

In this example the parent table is "Byggnad", which has two "Public bilder". Each "Publik bild" has an attachment. The first "Publik bild" also have two children of its own, which are comments on that "Publik bild" and are thus on the third level in the hierarchy.

__Bonus features as a result of implementation__
- The Modal component emits a new "closed" event when closed.
- The WFS source is implemented as a _class_ that extents _VectorSource_ instead of only implementing the _loader_ callback. This opens up for implementing more functions on the source like programmatically fetch and refresh individual objects, and changing the cql-filter to create a server side filtering based on user actions etc.
- The wfs transaction handler is _awaitable_ so the _edithandler_ can wait for a save operation to finish instead of relying on some shady events.
- A new option _isTable_ is added to WFS layers indicating that the layer does not have a geometry.

__Features considered but not implemented__
- Support for child layers with geometry. It is possible to edit the attributes of children with geometries,  but it is not possible to create new geometries or select a child from parent's attribute editor to edit the geometry.
- More aggregate functions, like min(), max().
- More ways of extracting attributes from child before aggregating to parent
- Handle adding child elements to newly created parents. If the parent has not been saved to database yet it has a guid as id and if _fid_ is used as PK the relation must be updated after save. As long as the edit store doesn't have an undo feature most applications will probably run in auto save mode anyway.  For the implemented  use case where the parent is considered read only, it is not possible to add new entries at root level anyway.
- Composite keys in the relation.

__Implementation details__
As this is a pretty complex PR and requries quite hefty configuration in order to work, it can be hard to thoroughly test. But it would be nice if someone could aid in testing that it doesn't break anything else as there are big changes in the edithandler.

Below is the configuration that creates the dialogs in the example. Only parent and person configuration is included, as they demonstrate all possible configurations. All relevant options are described in the comments, and all existing but not used options are also in the comments.

```javascript
{
  "layers": [
    {
      // This is the child layer. It must be configured as a normal layer. Only wfs is supported.
      // isTable should be set to true to ignore geometry and it should not be added to legend and
      // it should not be configured as editable. It becomes editable from parent's configuration
      "name": "sf:person",
      "title": "Ansvarig",
      "source": "local_wfs",
      "type": "WFS",
      // Set isTable to true in order to ignore geometry.
      // Also sets layer to visible to make OL actually query the layer.
      "isTable": true,
      // Works with both bbox and all. If isTable is true and strategy is bbox, requests will only be performed when using featureinfo 
      // or edit form as no features will be fetched using bbox if the features have no geometry.
      "strategy": "bbox",
      // Attributes are as normal, this defines how the child's attribute edit form will look like.
      // No need for configuring the attributes for featureinfo as this layer will not have its own featureinfo
      "attributes": [
        {
          "name": "namn",
          "title": "Namn",
          "type": "text"
        },
        {
          "title": "Paxtyp",
          "name": "typ",
          "type": "dropdown",
          "options": [
            "Mäta",
            "Riva",
            "Flytta in"
          ]
        },
        {
          "name": "typ2",
          "title": "Varför",
          "type": "dropdown",
          "constraint": "change:typ:Riva",
          "options": [
            "kackerlackor",
            "sumpgas"
          ]
        }

      ]
    },



    {
      // This is the parent layer
      "name": "sf:SketchMultiPoint",
      "queryable": true,
      "title": "Byggnad",
      "group": "root",
      "source": "local_wfs",
      "style": "origo-cities",
      "type": "WFS",
      "geometryName": "the_geom",
      "geometryType": "MultiPoint",
      "visible": true,
      "strategy": "bbox",
      "editable": true,
      "allowedEditOperations": [
        //"updateGeometry",
        //"create",
        //"delete",
        "updateAttributes"
      ],
      "attributes": [
        {
          "name": "Fritext",
          "title": "Fastighetsbeteckning",
          "type": "text",
          "readonly": true
        },
        {
          // This is the temporary attribute created by related table configuration
          // It can be used using any normal attribute option like name, url, html etc or in custom render function.
          "name": "hoist_person",
          "title": "Paxad"
        }
      ],
      // A list of all related layers. The related layers must be configured as usual layers.
      // Related layers will always be displayed in the parent's attribute edit form if the parent layer is editable.
      "relatedLayers": [
        {
          // Name of the child layer as already configured as a layer
          "layerName": "person",
          // The field that holds the private key in parent. defaults to fid (omit for fid as that can not be explicitly set)
          //"PK": "gid",
          // If PK is fid, remove layer name from fid. (default true)
          //"stripPK": true,
          // Name of field in child that holds the foreign key
          "FK": "parent_fid",
          // If the FK field is string. Can not always be determined in runtime so it has to be specified. Default false.
          "FKIsString": false,
          // The title that will be displayed above all child elements in the parent's attributes editor
          // Defaults to the child layer's "title" property.
          "childTitle": "Ansvarig",
          // Name of the attribute that will be displayed in the parent's attribute form. Defaults to fid.
          "featureTitle": "namn",
          // Limit the number of children that can be added. Only affects if the add button is visible. If there are more items in db, they are shown.
          "maxChildren": 1,
          // How to handle if parent is deleted. Can be
          // "cascade" : Delete recursively, both in db and locally.
          // "db" : Assume database has cascading delete and only recursively delete locally
          // "none" : Do nothing. Child remains both locally and in db, orphaned.
          // defaults to "none"
          "cascadingDelete": "cascade",
          // If the add button should not be visible
          //"disableAdd": true,
          // If the delete button should not be visible
          //"disableDelete": true,
          // If the edit button should not be visible
          //"disableEdit": true,
          // List of attributes that will be available on the parent. They are only used when displaying featureinfo and can be omitted if you 
          // don't want to show anything from child records.
          "promoteAttribs": [
            {
              // The parent will magically get an attribute by this name when featureinfo is displayed. It can be used
              // in parent attribute configuration like any other attribute.
              "parentName": "hoist_person",
              // How to extract attributes from child before aggregating.
              // Html is the only supported option. Works like the usual html attribute option.
              "html": "<p><b>{{namn}}:</b>{{typ}}</p>"
              // Separator is used when aggregating using string.join (default and only implemented) Defaults to ''
              // "separator" : "<br>"
            }
          ]
        }
      ]
    }
  ]
}

```


jokd:
Has anyone tested this? Or want to?
Would be sweet to merge it before next release.

Owemyr:
I did some testing with parts from provided example config and in my usecase it works good. Thanks @steff-o 

Initially had some problems with PK/FK when i tried different settings for `"PK":`
Failed on this `const childFeatures = childLayer.getSource().getFeatures().filter((f) => parentPK.toString() === f.get(FKField).toString());`

Probably just did something wrong related to config/backend.
With `"relatedLayers":` config as in example  it works without any issues.
Only tried with `"cascadingDelete": "cascade"`

I have not yet tested how featureinfo works with `"promoteAttribs": [`

steff-o:
@Owemyr It could be that the FK is null if the table contains orphaned children, which would make _toString()_ fail. I have added a check to allow orphaned entries in the child table. While at it I also added a check to allow empty PK ( which could happen when not using FID as PK).

Last commit also adds a check in featureinfo to ensure that it won't fail if someone created a SelectedItem without specifying a layer.

jokd:
I haven't thoroughly tested this but it looks like a great addition and according to previous comment  works good.
I suggest removing the changes in index.json and then we can merge this. Anyone disagree?
If so, @steff-o please add some rows in the documentation about the new feature.

steff-o:
Fixed the unintentional index.json update. 

Documentation coming up.

jokd (approved):

PR #1401, opened 2021-12-08
Snapping in measure feature

### LINKED ISSUE #811: Snap to objects with measure tool
Would be useful if one can snap to objects when using the measure tool. Perhaps as an option in the config just like the editor.

Polleryd:
Resolves #811 

Example config:
```
"controls": [ 
  {
    "name": "measure",
    "options": {
      "snap": true,
      "snapIsActive": true,
      "snapLayers": [
        "origo-cities",
        "osm"
      ],
      "snapRadius": 15
    }
  }
]
```
**snap**: enables snapping.
**snapIsActive**: sets the initial state of the snap (on/off).
**snapLayers**: list of layers that snap is enable for. If undefined, snap will be enabled for all layers.
**snapRadius**: distance from point where snap is triggered.


src/controls/measure.js
```diff
@@ -734,6 +937,20 @@ const Measure = function Measure({
           });
           buttons.push(clearButton);
         }
+
+        if (snap) {
+          toggleSnapButton = Button({
+            cls:
+              `o-measure-elevation padding-small margin-bottom-smaller icon-smaller round light box-shadow hidden activ ${snapActive && 'active'}`,
```
tonnyandersson, line 944: Should be of class `o-measure-snap`, not `o-measure-elevation`.

tonnyandersson:
Works fine!

Just a few things I noticed:

- No need to add the `ic_linear_scale_24px` icon since it's already in the material-icons.svg.
- The snap button icon would probably benefit from a 0 border radius since the default border radius cuts off the top corners of the icon. Look at `.o-measure-length .icon svg` and `.o-measure-area .icon svg` in scss/_measure.scss for reference.
- A _lot_ of (in my opinion unnecessary) changes in the general code structure, which makes it hard to see what _actual_ changes has been made. Minor cleanup efforts are always welcome, but I think major overhauls like this should be done separately. Again, my opinion.
- How much of an effort would it take to make it work on touch devices as well? Just from a quick glance, my impression is that it would be a bit tricky. Or?

Polleryd:
Closing this PR and replacing with #1410.

PR #1403, opened 2021-12-09
Allow multiple instances of a control

### LINKED ISSUE #1400: Several links and abouts in the config
**Describe the bug**
Previously, it was possible to add several link controls in the config where each link was displayed under the menu. This is no longer possible, only one of the links is displayed under the menu. This also applies to the about control.

**To Reproduce**
1. Add at least two links or abouts in your config file.
2. In Origo - Click on the menu.
4. Only one link/about appear.

**Expected behavior**
Every link/about should be displayed under the menu.

tonnyandersson:
Fixes #1400 in a very simple way. For example, it makes it possible to add multiple About or Link controls with different settings in the config. It also makes it possible to add multiple instances of _any_ control that is not implemented as a default control in origo.js, which might be considered as a negative side effect.

A different approach could be to modify the Link and About controls (and possibly others) to accept multiple links/abouts. However, making such an implementation backwards compatible would probably be a bit messy since old style single links would have to be handled one way, and new style multiple links a different way.

### DIFF
--- a/src/loadresources.js
+++ b/src/loadresources.js
@@ -125,10 +125,10 @@ const loadResources = async function loadResources(mapOptions, config) {
             }
 
             map.options = Object.assign(config, data);
-            map.options.controls = config.defaultControls || [];
+            map.options.controls = config.defaultControls.slice() || [];
             if (data.controls) {
               data.controls.forEach((control) => {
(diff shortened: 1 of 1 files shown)


johnnyblasta (approved):
Tested and LGTM

PR #1404, opened 2021-12-14
fix: make featureinfoLayer layer property function

### LINKED ISSUE #1384: featureinfoLayer layer property fails in use
**Describe the bug**
When the featureinfoLayer property is employed for a layer, referencing another layer of another kind with features in the same positions, for instance a WMS layer with featureinfoLayer: an AGS_TILE/MAP layer (that is queryably and works fine with infoclicks on its own), then info clicks/taps on features of the WMS layer fails to produce an info popup.

**To Reproduce**
With a popular LST ArcGIS Server WMS/Map service
```    {
      "name": "Fororenade",
      "id": "0",
      "source": "LST",
      "group": "root",
      "title": "Förorenade",
      "type": "AGS_TILE",
      "style": "mask"
    },
    {
      "name": "LST_Potentiellt_fororenade_omraden",
      "queryable": true,
      "source": "LST_WMS",
      "group": "root",
      "featureinfoLayer": "Fororenade",
      "title": "Förorenade WMS",
      "type": "WMS"
      },
```

**Expected behavior**
Info Popup

**Actual behaviour**
Crash in getfeatureinfo.js

**Additional context**

Grammostola:
Proposes to fix #1384. Thanks for suggestion toward solution @tonnyandersson , I've tested AGS_TILE featureinfoLayer: wms layer and vice versa, both seem to work.

### DIFF
--- a/src/getfeatureinfo.js
+++ b/src/getfeatureinfo.js
@@ -134,14 +134,14 @@ function getGetFeatureInfoRequest({ layer, coordinate }, viewer) {
       if (layer.get('featureinfoLayer')) {
         const featureinfoLayerName = layer.get('featureinfoLayer');
         const featureinfoLayer = viewer.getLayer(featureinfoLayerName);
-        return getGetFeatureInfoRequest({ featureinfoLayer, coordinate }, viewer);
+        return getGetFeatureInfoRequest({ layer: featureinfoLayer, coordinate }, viewer);
       }
       break;
     case 'WMS':
       if (layer.get('featureinfoLayer')) {
         const featureinfoLayerName = layer.get('featureinfoLayer');
         const featureinfoLayer = viewer.getLayer(featureinfoLayerName);
-        return getGetFeatureInfoRequest({ featureinfoLayer, coordinate }, viewer);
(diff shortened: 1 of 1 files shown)


tonnyandersson:
LGTM.

However, it doesn't work with other layer types. WMS/AGS_TILE works both ways, but, for example, WMS with a WFS layer as `featureinfoLayer` doesn't. I think this is because the `getGetFeatureInfoRequest` function returns `null` if the `featureinfoLayer` is of any other type than WMTS/WMS/AGS_TILE. I always though that would be the default use case, but I can't find any evidence that suggests it has ever worked with WMS/WFS before...

...which begs the question if that should be treated as a different issue or not.

Grammostola:
Yes, it looks as if AGS_TILE and WMS and WMTS type layers can have a featureinfoLayer and AGS_TILE and WMS can be featureinfoLayers. This change should make that work. In my opinion a separate issue would be to add to the supported list of featureinfoLayer types. Though one can certainly see the whole thing as one "Make it work _well"_ issue.

tonnyandersson (approved):

PR #1408, opened 2021-12-22
Legend for print map

### LINKED ISSUE #1382: Prints should be able to include a legend
The print control incorporates no map legend in the printed pdf/image and this hinders what kind of prints can reasonably be interpreted without ambiguity

A limited legend in the print layout

Endless legend on additional pages

**Additional context**
Map legend layouts on prints can be the subject of some debate as they are almost always a compromise of some sort. Also for Origo there is a need for it to work with several types of layers/services.

sweco-semvtw:
Proposes to fix #1382 

Adds a legend to the print maps top left corner that displays legends for all selected layers. Styles from index.json will be considered first. If a style is not found in index.json and layer is not of an AGS type then a call to getLegendGraphic will be made to the layer's source.

The legend can be hidden with the control menu. 

Tested using:
`WMS`, `WFS`, `GeoServer`, `ArcGIS`

Screenshot 2021-12-22 at 11-23-16 Origo exempel


steff-o:
I have tested against WMS and WFS using Geoserver and it works like a charm. But I have some observations:

1. There are some unnecessary calls to _console.log()_
2. This PR does not use the high resolution prints from #1407, and that PR does not include the legend. So my suggestion is to wait until that has been merged and update this PR.
3. As with #1407, if a layer lacks style it crashes.

I noticed that _onRender_ is now _async_. Nothing wrong with that, I'm a huge fan of _async_ myself and something has to be a _promise_ when calling the server. But as there is no _await_, all errors except those swallowed when the remote call is performed, will end up as _uncaught (in promise)_. Maybe that is not a problem as previously I assume all errors would have ended up as _uncaught_ anyway as the way I understand it, the code that calls _onRender_ did not catch anything. But it could be worth contemplating if it affects anything else.

jokd:
Works well for me

Grammostola:
It doesn't crash if a feature layer lacks a style property in index.json, instead a wfs layer will be present and without an icon, an AGS_FEATURE layer will not be present (in the legend). It does seem to crash if a WMS layer does not have a style.

Grammostola:
Correction: it does seem to crash if a WMS layer does not have a style property in index json and its source is local but defined in absolute terms, was getting 'tainted canvas' on toBlob() https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_enabled_image . I tried a WMS service from another place (from a source defined in absolute terms) and it worked without a style (no legend icon but present in the legend). The local source expressed as a relative path also worked.

The legend icons do seem to match the rest of the print, resolutionwise (and less console.log after the last commit) so I think it works well enough. We may choose to merge the other print PR before this one.

jokd:
@Grammostola Is this good to merge before 2.5? If so please resolve the conflicts.

Grammostola:
From what I've seen yes, it's mergable. The legend is optional in the printing pane and onRender being async hasn't produced anything unforseen while testing.

PR #1410, opened 2022-01-10
Origo master feature measure snapping

### LINKED ISSUE #811: Snap to objects with measure tool

Polleryd:
Resolves #811 

Example config:
```
"controls": [ 
  {
    "name": "measure",
    "options": {
      "snap": true,
      "snapIsActive": true,
      "snapLayers": [
        "origo-cities"
      ],
      "snapRadius": 15
    }
  }
]
```
**snap**: enables snapping.
**snapIsActive**: sets the initial state of the snap (on/off).
**snapLayers**: list of layers that snap is enable for. If undefined, snap will be enabled for all layers.
**snapRadius**: distance from point where snap is triggered.

Note:
Snapping works with vector layers.
The snapping feature works a bit differently for mobile devices. The snap occurs when placing the measure point and not while dragging the map.


johnnyblasta (approved):
Thoroughly tested on desktop and Android.
LGTM!
Nice work!

PR #1412, opened 2022-01-12
Open current suggestions on focus in Search control

### LINKED ISSUE #1141: Trigger search when placing cursor in search bar
After getting suggestions and then deselecting the search bar, the suggestions disappear. This is great, but if the user then wants to see the list of search results again there is no other way than to change the search text, e g adding and removing a random character.

I think triggering the search every time the cursor is placed in the search bar would be a an intuitive way for the user to hide/show the search results. The search still shouldn't trigger unless the specified minimum number of characters have been entered.

Approaches used in other systems include a dedicated "trigger search" button, or a hotkey such as the Enter key on a keyboard, but those seem inferior.

image

MattiasSp:
Fixes #1141.

Piggybacks on the listener on the `focus` event to show any suggestions stored from a completed search.

### DIFF
--- a/src/controls/search.js
+++ b/src/controls/search.js
@@ -203,6 +203,9 @@ const Search = function Search(options = {}) {
     });
     document.getElementsByClassName('o-search-field')[0].addEventListener('focus', () => {
       document.getElementById(`${wrapperElement.getId()}`).classList.add('active');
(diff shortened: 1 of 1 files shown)


MattiasSp:
Here's a map for testing the new bahaviour: https://karta.haninge.se/devmaps/PR1412/

jokd (approved):

PR #1415, opened 2022-01-13
Fix for search incorrect zoom

### LINKED ISSUE #1405: Search control zoom to wrong location
**Describe the bug**
In latest origo-master search control fails to zoom to the correct location/feature.
It fails only when feature location is not in current extent. 
When i revert Attachments PR #1372 it works fine again.

**Screenshots**
As seen in the gif, first search when feature is inside extent works but second fails.
search_probs_2021-12-15_6

**Additional context**
Origo-server is used as backend for search control.

steff-o:
Fixes #1405 by disabling the pan to overlay before search zooms to the search result feature, as the pan will be overrun anyway.

Added an option to disable panning when displaying popup as the panning animation was created using relative coordinates which became incorrect when immediately zooming to another extent. Also changed all known places where render() is called to support or make use of the option too keep old behavior.


Owemyr:
PR working fine for us, we use origo-server search option 3, thanks @steff-o for fix and explanation of the issue

jokd (approved):

Issue #1320, opened 2021-08-20
Make it more clear that legend has more information to show


johnnyblasta:
**Is your feature request related to a problem? Please describe.**
It is not obvious for all users that they can click on the layer to see the complete legend for that layer when it contains other than one symbol. It should be made more clear that there is more to see.

**Describe the solution you'd like**
Perhaps it should be some info icon to make it clear that more info is hidden.

**Describe alternatives you've considered**
A popup when hovering over the legend

asemoller:
What about having a "More option"-button (three small points) that is very common? Either vertically or horizontally. When you click on it, it can either display a small menu where you can select things like "Läs lagerinformation", "Zooma till hela lagret", "Exportera lager" or go directly to the layer information.  

I'm not sure where the icon should be placed. Does anyone have a suggestion?

jokd:
@asemoller Good idea. Left of the circle maybe?
![image](https://user-images.githubusercontent.com/6848075/143202341-e966d0c8-52ab-4569-b0ce-df8be293b962.png)

jokd:
Whether to display a menu or slide to the layer information I am not sure. The easiest way is to have all stuff under layer information, but if eg zoom to layer is an often used function it would be smoother to have it in a menu. But I'm dubious about that.

asemoller:
> @asemoller Good idea. Left of the circle maybe? ![image](https://user-images.githubusercontent.com/6848075/143202341-e966d0c8-52ab-4569-b0ce-df8be293b962.png)

Looks good!

But what does it look like if you also have a remove button there as well? And if the layer name contains many characters?

jokd:
In case we go with the menu option the remove layer function could be put there. There's definitely a risk the text will overflow and should be handled some way but that is probably a present problem as well.

asemoller:
Yes, of course it should be put there! 

This is how it looks now if a layer name is very long and have a remove button:

![image](https://user-images.githubusercontent.com/17123002/143208012-788df845-2a6a-4579-af43-d088043842fb.png)

So maybe it will work!

asemoller:
Oh, one thing, the "More Options" button is usually the last button on the right. Maybe like this?

![image](https://user-images.githubusercontent.com/17123002/143213331-02cf68ae-3744-408d-93c0-e9fa1d403e43.png)



I have always found it a bit inconsistent that the arrows for groups and subgroups are on different sides. Maybe the legend needs an overhaul. But I guess that's a antoher issue.

[...]

PR #1425, opened 2022-01-17
Origo master feature more option button

Polleryd:
Change as detailed in #1320

As an additional change, when layer is removable, a popup menu will open:
image


tonnyandersson:
Nice work!

I noticed a few small things, though. Possibly more to come, so don't update this PR just yet.

- We mainly use Material Icons icons in the UI. I suggest using the `#ic_more_vert_24px` icon and `icon-smaller` button class. The icon is available in _css/svg/vendors/material-sprites/svg-sprite-navigation-symbol.svg_. Just copy it to the main _material-icons.svg_.
- Personally, I think the popup menu would look a lot nicer with a `box-shadow: $shadow`. I would have removed the border as well, but perhaps some users wouldn't like that.

steff-o:
I have tested against WMS and WFS using Geoserver and it works like a charm. But I have some observations:

1. There are some unnecessary calls to _console.log()_
2. This PR does not use the high resolution prints from #1407, and that PR does not include the legend. So my suggestion is to wait until that has been merged and update this PR.
3. As with #1407, if a layer lacks style it crashes.

I noticed that _onRender_ is now _async_. Nothing wrong with that, I'm a huge fan of _async_ myself and something has to be a _promise_ when calling the server. But as there is no _await_, all errors except those swallowed when the remote call is performed, will end up as _uncaught (in promise)_. Maybe that is not a problem as previously I assume all errors would have ended up as _uncaught_ anyway as the way I understand it, the code that calls _onRender_ did not catch anything. But it could be worth contemplating if it affects anything else.

asemoller:
> I have tested against WMS and WFS using Geoserver and it works like a charm. But I have some observations:
> 
> 1. There are some unnecessary calls to _console.log()_
> 2. This PR does not use the high resolution prints from [Printing in higher and lower resolution #1407](https://github.com/origo-map/origo/pull/1407), and that PR does not include the legend. So my suggestion is to wait until that has been merged and update this PR.
> 3. As with [Printing in higher and lower resolution #1407](https://github.com/origo-map/origo/pull/1407), if a layer lacks style it crashes.
> 
> I noticed that _onRender_ is now _async_. Nothing wrong with that, I'm a huge fan of _async_ myself and something has to be a _promise_ when calling the server. But as there is no _await_, all errors except those swallowed when the remote call is performed, will end up as _uncaught (in promise)_. Maybe that is not a problem as previously I assume all errors would have ended up as _uncaught_ anyway as the way I understand it, the code that calls _onRender_ did not catch anything. But it could be worth contemplating if it affects anything else.

Does this apply to this PR?

steff-o:
@asemoller My bad. No it doesn't.

tonnyandersson:
@Polleryd What's the reason for making the popup menu a component under the layer switcher (_overlays_) and render it on load (albeit hidden), rather than creating popup menu components on demand - and possibly add them to the respective layer component (_overlay_) - when clicking the _More_ button?

Personally, I don't see why the layer switcher need to be aware of the popup menu(s) at all. On the other hand, I haven't analyzed every aspect of this problem so maybe I'm missing something crucial here. Hence the question.

tonnyandersson:
Also, rendering this stuff on load may make it more difficult to implement dynamically changing layer properties affecting the contents of the popup menu. Just a thought.

Polleryd:
> @Polleryd What's the reason for making the popup menu a component under the layer switcher (_overlays_) and render it on load (albeit hidden), rather than creating popup menu components on demand - and possibly add them to the respective layer component (_overlay_) - when clicking the _More_ button?
> 
> Personally, I don't see why the layer switcher need to be aware of the popup menu(s) at all. On the other hand, I haven't analyzed every aspect of this problem so maybe I'm missing something crucial here. Hence the question.

I know the code is a bit confusing. The reason I did it this way is that the overlay component has position relative and therefore the popup cannot be shown outside the overlay component. But maybe I am missing some other way of doing this. I could maybe add the popup directly to <body> from the overlay component. If you have a good way of solving this I am all ears :)

tonnyandersson:
I think it would be nice if the popup menu closed when clicking the _More_ button a second time.

Other than that, I think it all works quite well. I still have some issues with the implementation in general, though. For example, the _PopupMenu_ should probably have been implemented as a UI Component, with all the bells and whistles that comes with it. I know that the feature info popup is not implemented that way, but that doesn't mean this one shouldn't either.

However, since we are about to make a new release and this feature actually looks and works quite well, I think we can merge this anyway, given that you fix the popup menu closing issue i mentioned first.

johnnyblasta:
When using tab to navigate it's possible to make the legend to freak out a bit, see animation.
LayerPropFocus3

asemoller:
@johnnyblasta I get the same behavior on todays master when using tab.

johnnyblasta:
> @johnnyblasta I get the same behavior on todays master when using tab.

Will have to make a issue for that when this is merged than. No point of fixing it before this is in since it concerns the same files.

jokd:
Then let's merge this

tonnyandersson (approved):

PR #1426, opened 2022-01-17
feat: add the four new participating municipalities as of January 202…

### LINKED ISSUE #1411: Origo mask and point layer should include the new members of the project
Four new members have joined the project. Their geometries should be included in origo mask and point layer.

Grammostola:
…2 to the cities and mask geojsons. Aims to fix #1411


johnnyblasta:
A bit too much decimals compared to previous perhaps.

tonnyandersson:
Perhaps the default zoom level and center should be adjusted as well.

Grammostola:
Yeah, may as well include in this PR. What say you?
default_view

tonnyandersson:
I don't know what screen size is considered being "standard" these days, but the north/south extremes are cut out on older smart phones like Iphone 7/8. Which I certainly don't use myself. Only for work. My personal smart phone is the very, very latest.

Not sure if this is something we need to consider, though, since it would require the map to be zoomed out yet one more step. A bit much, IMO. Half a step would've been perfect.

Grammostola:
By default (constrainResolution: false) there could be a decimal starting zoom level, 6.5 for instance. The slight trouble in my eyes is that then OSM starts out looking slightly off, which it doesn't with a whole number (when staying on the precached zoom lvls). The counterpoint could be that it starts looking iffy as soon as you start zooming anyway (by default). Maybe we could find another kind of default background map.

tonnyandersson:
Yeah, my half step comment wasn't 100% serious. But it would have been a better fit for sure.

johnnyblasta (approved):
LGTM

Grammostola:
iphone_se
The decimal number isn't very even but the starting view from one of the more popular smaller mobile screens as of late looks rather nice I think

PR #1427, opened 2022-01-18
Package update

### LINKED ISSUE #1423: OL 6.12
Bump Origo to latest OpenLayers version.

johnnyblasta:
Closes #1423 
Did a npm update which brings the packages up to date, but it introduced a bug which is also fixed with this PR


tonnyandersson:
General question - what's our policy regarding `for` loops? Airbnb Style Guide states ["no iterators"](https://github.com/airbnb/javascript#iterators--nope).

I know there's a bunch of `for` in the code, but `forEach` is way more common (164 `forEach` vs 29 `for`). I see no reason to try to even that score.

In this case, the problem could be solved by using `Array.from(elements).forEach(...` or `[...elements].forEach(...` to turn the HTMLCollection into an array. Not saying it's better in any way, but it's more consistent. I like consistency.

johnnyblasta:
> General question - what's our policy regarding `for` loops? Airbnb Style Guide states ["no iterators"](https://github.com/airbnb/javascript#iterators--nope).
> 
> I know there's a bunch of `for` in the code, but `forEach` is way more common (164 `forEach` vs 29 `for`). I see no reason to try to even that score.
> 
> In this case, the problem could be solved by using `Array.from(elements).forEach(...` or `[...elements].forEach(...` to turn the HTMLCollection into an array. Not saying it's better in any way, but it's more consistent. I like consistency.

Sorry I was a bit old school there!

steff-o:
I don't think the Style Guide is against an old school _for_ loop. It's the _for in_ https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in loop they don't like.

Personally I think there is a place for old school for loops in modern programming as it is fast and has no side effects like introducing a new this-scope etc. Also _forEach_ and _async_ don't mix.

tonnyandersson:
I actually think they are against `for` loops:

image

I don't have any issues with `for` loops per se, or with old school stuff in general. Whatever works works, and sometimes old school stuff works even better. What I don't like is inconsistency. If everyone would go by their own personal preferences at any given time, we would end up with an [exquisite corpse](https://en.wikipedia.org/wiki/Exquisite_corpse) style codebase. The codebase is rather messy as it is (I've probably added a fair share to it myself). No point in messing stuff up even more for no apparent reason.

With that said, this particular issue isn't much of an issue at all. I just noticed it because of the (kinda) unnecessary change from `forEach` to `for`.

steff-o:
I just assumed _for_ was OK as lint complains about _for in_ but not old school _for_

I also agree that personal preferences should not have precedence over guide lines.

tonnyandersson:
OK, but frowned upon. :)

jokd (approved):

PR #1428, opened 2022-01-19
HTMLElement as modal content

### LINKED ISSUE #1406: Allow html element as content in Modal
The Modal can only take a html string as content. This is annoying when the elements in the modal should have eventhandlers. First the Modal have to be created and all items that should have handlers must have an id in order to get them as elements using document.getElementById() and then add the event handler. Besides being annoying, its error prone when data needed for identification is round tripped in html text.

The _content_ parameter should be either a string or a html element. In this way its still possible to use html text and use DOM elements as content.

**Additional context**
The underlaying method (createElement) already handles content as both text and elements. Its just when the Modal is created the content is stringified, which is unnecessary anyway, since either it is already a string, or it will contain _content.toString()_, which is most likely never what was intended.

steff-o:
Closes #1406 

Added possibility to specify the contents of a Modal as a `HTMLElement` object in order to pre build the contents complete with eventhandlers and all.

This PR adds a `contentElement` parameter to Modal constructor options. The `contentElement`  should be a  `HTMLElement` containing the content. If present, it takes precedence over the `content` option.

Hard to test, as it is only an api change and not used anywhere, but can be tested by calling `Origo.ui.Modal()` from console and also by regression tests of existing modals.

As a note this feature could probably have been implemented by allowing either text or HTMLElement in the content option, but in order to keep absolute backwards compability it is implemented as a new option. There is a small risk that the implementation of performing a toString() operation on the content value serves some purpose with certain content if it in fact is not a string already.

### DIFF
--- a/src/ui/modal.js
+++ b/src/ui/modal.js
@@ -3,10 +3,25 @@ import Element from './element';
 import Button from './button';
 import { html } from './dom/dom';
 
+/**
+ * Creates a modal and displays it. The modal is created as a div tag that is attached to the DOM as a child of the options.target element.
+ * @param {any} options Object containing options for the modal. Valid options are:
+ *  title: Title of modal
+ *  content: The content to display. String containing html
+ *  contentElement: HTMLElement containing the content. If specified the 'content' param is ignored.
+ *  cls: Optional classes to append to the modal container div
+ *  static: true if modal must be explicitly closed
+ *  target: DOM element to wich modal div is appended as a child. Required. Most likely the viewer element.
+ *  closeIcon: name of icon to use as close button. Defaults to a cross.
+ *  style: html style attributes to add to modal div
+ *  newTabUrl: URL to create a link button to
+ *  @returns {any} A Component representing the modal.
+ */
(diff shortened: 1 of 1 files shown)


tonnyandersson:
@steff-o Can you provide a simple working example?

Starting to feel a bit stupid here because all my endeavors end up with event handlers being thrown out the window when the new HTMLElement is added and the Modal is displayed. I'm probably doing something wrong, but I can't figure out what.

steff-o:
No, I can't. Maybe it only worked in my mind, there's no way that would have worked as the content was serialised and deserialised, which in effect created a deep copy of the DOM element. So it worked for elements, but event handlers were dropped. I did a test a long time ago, and it worked, but probably I circumvented all the render stuff back then and forgot that I cheated.

I have pushed an updated version that actually works.

let el = document.createElement('div');
el.innerHTML = '<b>This is a button</b>'

tonnyandersson:
LGMT

tonnyandersson (approved):

Issue #1420, opened 2022-01-14
Search configuration revamp


MattiasSp:
**Description**
When configuring the Search control, the five current options are an emergent behaviour rather than explicitly configured. Choosing a search mode is done by selecting the correct combination of config parameters, some of which have sort of the same meaning but different names - in order to differentiate which search mode is to be used.

This is, in my opinion, a rather indirect and less than user (admin) friendly way of controlling the Search control's behaviour.

**Proposed solution**

In essence, I see the Search control config as having two different parts;

1. A search endpoint
_that is queried and returns a JSON array_
    - `url`
    - `searchAttribute`
    _specifies which attribute in the JSON results that should be displayed in the search result list_
    - `title`, `titleAttribute` or the title of the map layer referenced by the attribute named with `layerNameAttribute` or `layerName`
    _used to group results under different headings when `groupSuggestions` is true_
2. Feature info configuration
_config parameters that define where to get the geometry and attributes to be able to show or pan to a search result in the map_
    - `geometryAttribute` (or `northing` plus `easting`)
    - `title`, `titleAttribute` or the title of the map layer referenced by the attribute named with `layerNameAttribute` or `layerName`
    _the title is used again as a heading for the feature info popup when a result is clicked_
    - `idAttribute`
    _only used for option 1, to get the id of the feature for which to fetch feature info from the layer_
    - `contentAttribute`
    _only used for option 3, to use another attribute for the feature info popup text, separate from `searchAttribute`_

The current complexity is due to the feature info configuration and stems from the need to fetch information from different sets of sources;
- From the search result itself,
- from named map layers,
- or from text value in the Search config.

So why not go all in and make the configuration both fully flexible (allowing any combination of sources instead of just the current five) and more explicit at the same time?

Based on the above analysis I suggest a revamp of the configuration where we explicitly state both the source reference and its source type for each required information. Also, the `searchAttribute` should be renamed to something like `displayAttribute` since it doesn't actually control which attribute is searched - only which attribute in the results that should be displayed.

I'll try to add suggestions for what the configs would look like for the five different options soon.

jokd:
This will be a nice improvement. Can a breaking change with older configs be avoided?

MattiasSp:
> This will be a nice improvement. Can a breaking change with older configs be avoided?

[...]

PR #1436, opened 2022-01-27
Searchlist in infowindow

### LINKED ISSUE #1413: Improve how the search results are presented
We need to be able to see all hits when we do a search in Origo. One suggestion is to design the  window  for the search results in the same way as the information window. That way it is possible to get a nice overview with all results presented in bundles and we will keep a consistent look.

jokd:
Fixes #1413 
Adds a infowindow-component which can be used to create a floating infowindow or aligned to the left.
Search control is configured ike this:
{
"name": "search",
"options": {
	...
	"autocompletePlacement": "search", //can be search, left or floating (defaults to search)
	"searchlistPlacement": "left" //can be search, left or floating, activated by hitting enter (defaults to nothing)
}
}


tonnyandersson:
Tested with old default test config (v. 2.2.0). Using anything other than `"search"` for `"autocompletePlacement"` or `"searchlistPlacement"` results in this error:

image

image

jokd:
I suppose the structure of our search responses differ. I will look into it.

jokd:
This feature works when using Option 1 for search:
```
   * Option 1. Feature info is requested from a map service.
   * In this case idAttribute and layerNameAttribute must be provided.
   * A map service is used to get the geometry and attributes. The layer is defined
   * as an ordinary layer in the layer config section.
```
Will have to make adjustments to support the other variants as well. But maybe I await this PR first: https://github.com/origo-map/origo/issues/1420

@MattiasSp Are you looking into this?

jokd:
Replaced with #1751 to get a clean commit history

PR #1439, opened 2022-01-28
Adds share map functionality when using infowindow

### LINKED ISSUE #928: When infowindow is infowindow no sharelink object zoom.
Found one more bug, when infowindow is infowindow the zoom to object wont work at all. No id in url is present

_Originally posted by @mulfvik in https://github.com/origo-map/origo/pull/917#issuecomment-627834451_

jokd:
Fixes #928

### DIFF
--- a/src/permalink/permalinkstore.js
+++ b/src/permalink/permalinkstore.js
@@ -48,9 +48,19 @@ permalinkStore.getState = function getState(viewer, isExtended) {
   } else {
     const selectedItems = viewer.getSelectionManager().getSelectedItems().getArray();
     if (selectedItems.length > 0) {
-      const layerType = selectedItems[0].getLayer().getProperties().type;
-      if(layerType === 'AGS_FEATURE' || layerType === 'WFS' || layerType === 'GEOJSON' || layerType === 'TOPOJSON') {
-        state.feature = selectedItems[0].getId();
+      const layer = selectedItems[0].getLayer();
+      const layerType = layer.get('type');
+      const layerName = layer.get('name');
+      if (layerType === 'AGS_FEATURE' || layerType === 'WFS' || layerType === 'GEOJSON' || layerType === 'TOPOJSON') {
+        const id = selectedItems[0].getId() || selectedItems[0].ol_uid;
+        if (layerType === 'WFS') {
+          const idSuffix = id.substring(id.lastIndexOf('.') + 1, id.length);
+          state.feature = `${layerName}.${idSuffix}`;
+        } else if (layerType !== 'WFS') {
+          state.feature = `${layerName}.${id}`;
+        } else {
+          state.feature = id;
+        }
       }
     }
   }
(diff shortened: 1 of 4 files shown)


src/permalink/permalinkstore.js
```diff
@@ -45,6 +45,14 @@ permalinkStore.getState = function getState(viewer, isExtended) {
 
   if (featureinfo.getSelection().id && (type === 'AGS_FEATURE' || type === 'WFS' || type === 'GEOJSON' || type === 'TOPOJSON')) {
     state.feature = featureinfo.getSelection().id;
+  } else {
+    const selectedItems = viewer.getSelectionManager().getSelectedItems().getArray();
+    if (selectedItems.length > 0) {
+      const layerType = selectedItems[0].getLayer().getProperties().type;
+      if(layerType === 'AGS_FEATURE' || layerType === 'WFS' || layerType === 'GEOJSON' || layerType === 'TOPOJSON') {
+        state.feature = selectedItems[0].getId();
```
tonnyandersson, line 53: `getId()` only returns id without layer name, at least when testing with default config. Feature part of sharemap url turns out like `&feature=8`.

jokd:
You are right, I just tested our production config. A local geojson file obviously needs some other treatment.

jokd:
Haven't tested with AGS or TOPOJSON services

steff-o:
I don't think that this solution will work if using the _id_ parameter on WFS layers.

When creating the link the string is set to `&FEATURE=<layerName>.<PK>` and then when the link is restored it is assumed that `layerName` is in fact the the name of the layer on the map server when creating the id of the feature as `layername` is used both for identifying the layer and creating the id of the feature. If using the _id_ parameter on the layer this assumption will fail.

This is however probably not a new issue and consistent with current implementation. 

The simple solution would of course be to ignore the significance in WFS id and just use `<layerName>.<id>` for all layer types. That would unfortunately break old share links.

To solve it it would be necessary to check if the layer has an _id_ parameter and use that as first part of id instead of layerName when the link is restored.

jokd:
I'd say we have always used the layername for this feature so I think it'll work with old links as well.

steff-o:
Yes it will work with old links the way it always have worked. Just thought it would be a great opportunity to fix it for when using the _id_ parameter on WFS layers as that probably never have worked. Haven't tested though so I will drop it for now and maybe come back with a new issue.

tonnyandersson:
The _zoom to_ animation is not what it used to be.

**Before**


**After**


I think this is caused by two `fit()` functions working against each other.

jokd:
> The _zoom to_ animation is not what it used to be.
> 
> **Before** 
> 
> **After** 
> 
> I think this is caused by two `fit()` functions working against each other.

Yeah I am planning on making the one in showFeatureInfo optional when this and #1436 are merged. Then it will hopefully be a bit more smooth.

tonnyandersson:
Ok, I'll allow it. For now.

tonnyandersson (approved):

PR #1440, opened 2022-01-28
Monkey patch OL to correct position of mouse events in print preview

### LINKED ISSUE #1422: Wrong place for selected polygon in print preview
**Describe the bug**
In print preview it is possible to select a polygon (or area) by clicking on the map while highlighted area after the click, which will be shown in blue color, doesn’t match with selected polygon, i.e. the area that was clicked on the map.

**To Reproduce**
1. Go to 'print preview'
2. Click on a polygon (which is selectable)
3. Another polygon on the map will be selected

**Screenshots**
image
Here a property in the map is selected but the selected property is in the wrong place. The property which was clicked is at the corner of the map.

steff-o:
Fixes #1422

__Beware__: this is a monkey patch that overrides an internal OL function and may very well break when OL is updated.

In order to minimise side effects, the monkey patch is only active when print preview is active.

Not all events are fixed with this PR as they internally in OL uses another way of calculating the position of the event. Pinch zoom and pan are among those interaction that is not fixed.

### DIFF
--- a/src/controls/print/print-component.js
+++ b/src/controls/print/print-component.js
@@ -5,6 +5,7 @@ import TileImage from 'ol/source/TileImage';
 import TileWMSSource from 'ol/source/TileWMS';
 import TileGrid from 'ol/tilegrid/TileGrid';
 import { Group } from 'ol/layer';
+import PluggableMap from 'ol/PluggableMap';
 import {
   Button, Component, cuid, dom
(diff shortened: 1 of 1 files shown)


johnnyblasta (approved):
Working fine, tested with one of my configs.
LGTM

jokd (approved):

PR #1445, opened 2022-01-28
Adds style filter DATETIMEDIFFNOW

### LINKED ISSUE #1433: Feature style based on time
When the data contains a datetime component it would be nice to be able change the style based on the datetime and specifically on the difference from now. A use case can be that there is data when a ski track is last preperaed and the style should have different colors f.e. if it was preperaed in the last 3 hours in green and last 24 hours in yellow.

I would like to setup a filter that takes a attribute and compares to how many seconds from now has past since the value specified.

```
[
  {
    "fill": {
      "color": "rgba(170, 255, 0,1)"
    },
    "stroke": {
      "color": "rgba(170, 255, 0,1)",
      "width": 2
     },
     "filter": "[dateLastPreparation] DATETIMEDIFFNOW 10800",
     "label": "Prep senaste 3 timmar",
     "zIndex": 90
  }
]
```

More options to filter on datetime could be that you want to specify a specific date, month or year, but don't have any use case for that for now so leave it for future development.

johnnyblasta:
Closes #1433 

Adds a style filter that can filter on how long from now has past since the attribute value which contains a date time.

Config example:
```
[
  {
    "fill": {
      "color": "rgba(170, 255, 0,1)"
    },
    "stroke": {
      "color": "rgba(170, 255, 0,1)",
      "width": 2
     },
     "filter": "[dateLastPreparation] DATETIMEDIFFNOW 10800",
     "label": "Prep senaste 3 timmar",
     "zIndex": 90
  }
]
```

### DIFF
--- a/src/style.js
+++ b/src/style.js
@@ -172,6 +172,7 @@ function checkOptions(feature, scale, styleSettings, styleList, size) {
         // find attribute vale between [] defined in styles
         let regexExpr;
         let regexFilter;
+        let dateDiffNowFilter;
         let featMatch;
         let filters = [];
         let filtering = '';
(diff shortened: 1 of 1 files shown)


tonnyandersson:
IMO, this caters to a too specific use case with basically no flexibility. I think it would be a whole lot more useful to implement an expression handler of some sort, kinda like the way it is done in ArcGIS/QGIS. Adding various one-off functions like this here and there will only result in a lot of headaches down the line.

This could have been done on the backend side (as mentioned in the corresponding issue) or as a custom style function.

Just my opinion.

johnnyblasta:
When not having control of the backend it seems a hassle to set up a proxy to solve it there.
Yes a custom style function could do the trick also.
Yes it wasn't the most refined solution to a problem and in my opinon how the whole style filter works could need a complete rewrite.

johnnyblasta:
Closing this to rework it to a more general function

PR #1448, opened 2022-02-01
Account for layer in alreadyExists

### LINKED ISSUE #1444: Infowindow only shows first feature if multiple features with same ID is selected
**Describe the bug**
Infowindow only shows first feature if multiple features with same feature ID number is selected.

**To Reproduce**
1. In default config, use `"infowindow": "infowindow"`
2. Duplicate the `origo-cities` layer and rename both like `origo-cities-1` and `origo-cities-2`
3. Load map and click on any feature

**Expected behavior**
Infowindow should display (at least) two hits from two different layers, but only one hit (the first) is displayed.

image

Apparently, this is a known issue since it is mentioned in the code. The quick and dirty way to fix it would be to concatenate layer name and feature ID in the `some` test below (from _selectionmanager.js_). Any other suggestions?

``` javascript
function alreadyExists(item) {
  // FIXME: Take into consideration which layer? Also affects remove and all other by id functions.
  // Right now, if several layers use the same source, a feature can only be selected in one layer (the first attempted)
  return selectedItems.getArray().some((i) => item.getId() === i.getId());
}
```

tonnyandersson:
Fixes #1444.

Thanks for the `selectionGroup` tip, @jokd.

### DIFF
--- a/src/selectionmanager.js
+++ b/src/selectionmanager.js
@@ -23,9 +23,7 @@ const Selectionmanager = function Selectionmanager(options = {}) {
   const isInfowindow = options.infowindow === 'infowindow' || false;
 
   function alreadyExists(item) {
-    // FIXME: Take into consideration which layer? Also affects remove and all other by id functions.
-    // Right now, if several layers use the same source, a feature can only be selected in one layer (the first attempted)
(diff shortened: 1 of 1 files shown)


jokd (approved):
Works well and shouldn't break the multiselect plugin

PR #1450, opened 2022-02-02
Fix to support id parameter on wfs layers in sharemap

### LINKED ISSUE #1446: Share map does not show selected feature when id option is used
**Describe the bug**
If a share map link or mapstate includes a feature, the feature will not be displayed if it belongs to a WFS layer that is configured using the _id_ option to specify the layer identity on the server different from the _name_ option. Most likely it only fails for WFS.

**To Reproduce**
1. Configure a WFS layer to have the _id_ option set to the name of the layer in the WFS server and the _name_ option to something completely different.
2. Click on a feature from aforementioned WFS layer to open the featureInfo.
3. Use the share map control and copy the link
4. Open the link in your favorite browser, or edge.
5. The feature is not zoomed to nor its pop up visible

**Expected behavior**
The feature selected before sharing the map should be zoomed to and its pop up visible.

**Additional context**
The problem is that share map uses layer name as identifier for both layer name and featureId in the mapstate by assuming that featureId is `<layerName>.<PK>` , but that does not hold when _id_ option is used, as it is the WFS server that sets the featureId, which in this case would be `<layerId>.<PK>`. It is an easy fix i viewer.js to check if the layer has the _id_ option set and set the _featureId_ accordingly before fetching the feature.

steff-o:
Fixes #1446 

Uses the value from the id-option (if present) on wfs layers to construct the featureId when opening a shared map. The id option takes precedence over layer name as that's how the layer is created,

### DIFF
--- a/src/controls/print/print-component.js
+++ b/src/controls/print/print-component.js
@@ -253,9 +253,11 @@ const PrintComponent = function PrintComponent(options = {}) {
     const widthInMm = orientation === 'portrait' ? sizes[size][1] : sizes[size][0];
     widthImage = orientation === 'portrait' ? Math.round((sizes[size][1] * resolution) / 25.4) : Math.round((sizes[size][0] * resolution) / 25.4);
     heightImage = orientation === 'portrait' ? Math.round((sizes[size][0] * resolution) / 25.4) : Math.round((sizes[size][1] * resolution) / 25.4);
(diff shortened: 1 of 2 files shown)


jokd:
Tested a coouple of configs and seems to work well

jokd (approved):

Issue #1402, opened 2021-12-08
Tab navigation in list of layers


Maryzandi:
After using "Tab" key to navigate in list of layers, appearance of the list changes and the only way to come back to the previous appearance is to refresh the map.  

See the images:
![image](https://user-images.githubusercontent.com/29747817/145202739-f254a552-03da-4bbc-86dc-1ae99aa1f92d.png)
![image](https://user-images.githubusercontent.com/29747817/145202777-775e1be2-b842-4731-a65b-a10c875786e4.png)

johnnyblasta:
This is something that has been around for a long time, it's a hidden button that takes focus and messes up the legend.

johnnyblasta:
@Maryzandi Which version of code do you use? 
I tested now and didn't see this bug. 
Might have been fixed with PR #1374

Maryzandi:
@johnnyblasta We downloaded the code last week.

johnnyblasta:
> @johnnyblasta We downloaded the code last week.

The 2.4.0 release or the master branch? The master branch is the most current code.

Maryzandi:
Yes, We use the master branch.

Maryzandi:
@johnnyblasta I wonder if you could reproduce this bug? Otherwise I can add more screenshots.

jokd:
@Maryzandi If you can provide a config or link to a map where this happens it would be easier to test/debug.

Maryzandi:
In this video I have tried to demonstrate how the list of layer changes after tab navigation. We downloaded the code yesterday.
![tab4](https://user-images.githubusercontent.com/29747817/152365281-b031b91f-bc41-4754-ab8c-1eb301bac6b4.gif)
.

jokd:
Is it all layers or just the ones with layer names on multiple rows?

Maryzandi:
The appearance of the list changes for all layers and its impossible to see the full names of layers in the list any more.

johnnyblasta:
It's some button that is hidden and should have a negative tabIndex so that it is excluded in the tab order.

jokd:
@Maryzandi @johnnyblasta Is this solved now?

Maryzandi:
@jokd We downloaded the latest code and it seems that it is solved now.

PR #1454, opened 2022-02-04
Fixes tab and focus on layer properties

### LINKED ISSUE #1452: No close for layers properties with keyboard
**Describe the bug**
When tabing and opening the layers properties the back button isn't reachable and there is no focus on the layer properties, so it acts a bit strange.

**To Reproduce**
See animation:
LayerPropFocus

**Expected behavior**
When opening layer properties with keyboard it should be in focus on back button and when closed the it should be exluded from the tab order.

**Additional context**
There is a issue #1320 with a PR #1425 that is adressing layer properties so maybe it will be fixed there or it has to be fixed after that is merged.

johnnyblasta:
Closes #1452 and fixes partly or completely #1402

### DIFF
--- a/src/controls/legend/overlays.js
+++ b/src/controls/legend/overlays.js
@@ -267,6 +267,17 @@ const Overlays = function Overlays(options) {
           });
           slidenav.setSecondary(layerProperties);
           slidenav.slideToSecondary();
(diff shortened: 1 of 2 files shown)


johnnyblasta:
LayerPropFocus2

jokd (approved):

PR #1455, opened 2022-02-08
Export selection to local file

### LINKED ISSUE #1429: Export selection to file
When a user has selected some features in one or more layers they often would want to make a thematic map to share with other people and organisations. 

A function to export the selection as displayed in infoWindow to a local file that can be imported to other maps. The file format is not crucial as most organisations have means to transform between the most common ones, but as Origo has a _drag and drop_ function, it would of course be of interest to export to a format that can be dragged and dropped into origo.

This could be implemented client side for the text formats supported by OL (GeoJson, Gpx and such). This way it would not require a server side component like the present export which does not support geometries anyway.

Augmenting the server side export with support for geometries, but that would require a server side component as well.

**Additional context**
It could be implemented alongside the existing export as an alternative to the server side export.

steff-o:
Closes #1429

Adds an _export to file_ button to _infoWindow_.

Also exposes a utility function in the `Origo.Utils` "namespace" to export arbitrary features. Usage is documented in code.

Config by adding something like this to the `infoWindowoptions.export` configuration. It has the lowest precedence of the three kinds of export.

```
 "clientExport": {
          "buttonText" : "Exportera fil",
          "layers": [ "SketchMultiPoint", "To2018_Swe99TMWhatever" ],
          "format": "geojson"
        }
```

`format` is the file extension of the file format to export. Currently supported is `'geojson'`, `'gpx'` and `'kml'` . In theory more formats are supported by openlayers, but I picked the subset that can be drag and dropped and has the writeFeatures function implemented.

`layers` specifies the SelectionGroup name for those layers that are allowed to export from. If omitted, all layers are allowed.

In theory the configuration could have been incorporated in the simpleExport and layerSpecificExport using only a `format` option instead of url, but for simplicity and future proofing I made it a new kind of export.

__Known issues__
GPX only exports lines and multi lines. Not my fault, that's the way openlayers does it.

Exported GeoJson files does not contain coordinate system information. This is according to specification, but could be perceived as a flaw as it is common to include a `crs` property.

Exported features contains a property `"state" : "selected"` if they are highlighted. This is the result of how selectionManager handles selected features (which seems a bit dangerous, as 'state' is not unlikely to appear as a field in the data).

### DIFF
--- a/src/infowindow.js
+++ b/src/infowindow.js
@@ -1,4 +1,5 @@
 import { simpleExportHandler, layerSpecificExportHandler } from './infowindow_exporthandler';
+import exportToFile from './utils/exporttofile';
 
 let parentElement;
 let mainContainer;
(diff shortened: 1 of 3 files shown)


jokd:
Works well, to use with the default map:
```
"featureinfoOptions": {
  "infowindow": "infowindow",
  "infowindowOptions": {
    "export": {
      "clientExport": {
        "buttonText" : "Exportera fil",
        "layers": [ "origo-cities" ],
        "format": "geojson"
      }
    }
  }
}
```

jokd (approved):

jokd:
I wouldn't protest if those console.warn were removed though

asemoller:
It's actually possible to export group layers so they should be removed! At least it's possible with simpleExport and layerSpecificExport.

jokd:
> It's actually possible to export group layers so they should be removed! At least it's possible with simpleExport and layerSpecificExport.

Do you mean that there should be a check in the code so the button won't appear for group layers? The easiest way to solve that is to avoid group layers in the export config :)

steff-o:
At least the two last warnings annoy me, as it may very well be the case that is intentional that the layer should not have an export. The first warning, if it is a group layer, is also probably not necessary, as group layer may very well be exported. As the features are exported from the selection group, there is no immediate connection to the underlaying layers so it doesn't add to the complexity. Still there are some caveats that could lead to unexpected results: 
1. The features can have colliding identities
2. In case of clientExport there can be mixed geometry types in the export. Which is fine for many file formats, but GIS applications tend to not import more than one type from each file.

Either way, I never changed any of it in this PR I just added a comment, but I'll be happy to remove the warnings while at it as I think they should go. It is up to the application configurator to test out the application anyway.

PR #1460, opened 2022-02-14
fix: enable print resolution change via compatible loop for htmlColle…

### LINKED ISSUE #1459: Print fails
Load the default map with
`npm run start` , open in a browser, hit print, try and change the resolution

**expected**
changed resolution

**happens instead**
failure has already happened in the console

**Additional context**
The following happens when the print panel is opened:
```
Uncaught (in promise) TypeError: icons.forEach is not a function
    resizeLegendComponent print-resize.js:307
```

a forEach is trying to iterate over an HTMLCollection

(The print panel works well incl legend if the new print option "suppressNewDPIMethod" is set to true)

Please stand by for emergency triage PR

Grammostola:
…ction . Fixes #1459

### DIFF
--- a/src/controls/print/print-resize.js
+++ b/src/controls/print/print-resize.js
@@ -304,11 +304,11 @@ export default function PrintResize(options = {}) {
 
       const icons = document.getElementsByClassName('legend-icon');
       if (icons.length > 0) {
-        icons.forEach(icon => {
-          const style = icon.style;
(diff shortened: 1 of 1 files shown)


Grammostola:
(I follow the linter and I forgot about the corpse)

jokd (approved):

PR #1462, opened 2022-02-15
Ensure that layername is unique

### LINKED ISSUE #1458: Dragging extern files to map - Strange behaviour if adding two layers with the same name
When we drag two or more files with the same name to the map, and remove them from the map then there will be some strange behaviour. 

1. If you take away the layer  from below first, you cannot remove the upper most layer.

externa_filer1

2. If you take away the layer from above first, you can remove all the layers from map. But when you add back one layer with the same name again, all the layers which have been removed will appear again in the layer list (but not the objects).  

externa_filer2

This has been tested in the latest origo code we downloaded on 2022-02-09 and earlier versions. We tested in Chrome and Edge in Window 10 and they give the same behaviour.
.

jokd:
Fixes #1458 
Ensure unique names are used for drag and drop-layers

### DIFF
--- a/src/controls/draganddrop.js
+++ b/src/controls/draganddrop.js
@@ -78,6 +78,17 @@ const DragAndDrop = function DragAndDrop(options = {}) {
       map.addInteraction(dragAndDrop);
 
       dragAndDrop.on('addfeatures', (event) => {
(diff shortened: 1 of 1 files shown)


johnnyblasta (approved):
Tested throwing a bunch of files on it and it handles it without errors!
LGTM

PR #1465, opened 2022-02-15
Readds the drag and drop control after print

### LINKED ISSUE #1457: Dragging extern files to map - sometimes it does not work
We have experienced  for some times (both in the lastest code from 2022-02-09 and earlier versions) the function that adding extern file to map is not stable. Sometimes it does not work and we do not know what the reason is. This is experienced in Chrome (not sure about other webbrowsers)

We have to refresh the map to make it work. 

externa_filer

jokd:
Fixes #1457

### DIFF
--- a/src/controls/draganddrop.js
+++ b/src/controls/draganddrop.js
@@ -10,13 +10,15 @@ import Style from '../style';
 import { Component } from '../ui';
 
 const DragAndDrop = function DragAndDrop(options = {}) {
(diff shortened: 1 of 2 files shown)


johnnyblasta:
Still don't work after open/close of print.

jokd:
Sorry called wrong function, please try again

johnnyblasta (approved):
Nice, now it works!

PR #1473, opened 2022-02-21
Support for changing share map hash

### LINKED ISSUE #1453: Sharemap - respond to parameter change
If the sharemap parameters are changed the map does not reflect the new parameters. This is annoying when embedding a map in an iframe and the source of the frame is changed to display another sharemap state.

The map should identify that the parameters have changed and update itself to reflect the new mapstate.

1. Not embed the map in an iframe and use api-functions to pan and zoom the map. But in some situations it is not feasible and an iframe is the best option and the sharemap mapState can not be used directly as a link.
2. Replacing the entire iframe in order to force the browser to reload the entire map. This would lead to slower response and requires more actions from the calling side
3. Use `window.postMessage()`. This approach would require adding some javascript code to both sides in each application and the sharemap mapState can not be used directly as a link. 

**Additional context**
As the mapstate is read from hash, it could be implemented by adding a `'hashchange'`-eventlistener and do whetever the map does normally does with the hash part of the url.

__Edit:__ Does anyone know why mapstateId uses query param, but stateless uses hash?

steff-o:
Resolves #1453.

This PR makes it possible to update the hash part of the url to change to another sharemap state.

The map is completely "rebooted" to fully reflect the new incoming map state. In order to do so without messing around too much in the viewer a completely new Viewer is created instead of trying to restore all state. As a new viewer is created, external code that uses the `'load'`-event should also reinitialise itself using the new viewer.

### DIFF
--- a/origo.js
+++ b/origo.js
@@ -25,6 +25,7 @@ import { renderSvgIcon } from './src/utils/legendmaker';
 import SelectedItem from './src/models/SelectedItem';
 import 'elm-pep';
 import 'pepjs';
+import permalink from './src/permalink/permalink';
 
 const Origo = function Origo(configPath, options = {}) {
   /** Reference to the returned Component */
@@ -112,9 +113,13 @@ const Origo = function Origo(configPath, options = {}) {
       .catch(error => console.error(error));
   };
   // Add a listener to handle a new sharemap when using hash format.
-  window.addEventListener('hashchange', () => {
-    // "Reboot" the application by creating a new viewer instance using the original configuration and the new sharemap state
-    initViewer();
+  window.addEventListener('hashchange', (ev) => {
+    const newParams = permalink.parsePermalink(ev.newURL);
+
+    if (newParams.map) {
+      // "Reboot" the application by creating a new viewer instance using the original configuration and the new sharemap state
+      initViewer();
+    }
   });
 
   return ui.Component({
--- a/origo.js
+++ b/origo.js
@@ -27,6 +27,8 @@ import 'elm-pep';
 import 'pepjs';
 
 const Origo = function Origo(configPath, options = {}) {
+  /** Reference to the returned Component */
+  let origo;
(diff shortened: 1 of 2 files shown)


jokd:
How about making initViewer take a new config as an argument so it can be used to load a new config as well? Or add a setConfig method to origo.js to do just that ( but then maybe you can't roll back to default config)

steff-o:
It actually kind of already takes a new config as param. It is included in the sharemap hash and the usual config resolving is used. If no mapstate is present it resolves to the default or configured .json file. When a mapstate is present the `map` argument is used. It is not however possible to just call initViewer with another config. I did not want to mess around too much with the config resolving inside the Viewer.

Maybe a good combination of getting exactly what the sharemap state says and a possibility to only change the current state of the map would be checking if the `map` hash argument is present. If it is present a "reboot" using the provided config is performed. If it is not present, all other hash arguments are read and applied to the current state. E.g. it would be possible to only change the extent or select a feature or put a pin or turn on or off layers or a combination of those actions by setting the desired hash arguments.

But to be honest, I don't think I will implement setting individual arguments as I see no strong use case as it is possible to achieve that by using the api and it would require some serious rewriting of the Viewer if the code would not have to be duplicated. The main objective with this PR is to handle when an iframe is reused with different mapstate hashes.

What I will do is however add a check if the `map` argument actually has changed in order to avoid reloading the map if an application has its own _hashchange_ handler dealing with other arguments. When that is in place there will be a suitable place to start implementing all other arguments that can be set in the hash if someone needs it.

steff-o:
I changed the code to only reboot the map if the map parameter is specified in order to allow for application to use its own hashes. All other parameters or hashes are ignored unless in combination with map parameter.

tonnyandersson:
@jokd, any final thought on this? If not, I'll approve and merge this by the end of the week, 'cuz LGTM.

Issue #1329, opened 2021-09-01
Minor legend


asemoller:
**Is your feature request related to a problem? Please describe.**
Today, the legend control functions mainly as a list of layers. When there are many layers and groups in the list, it can be difficult for the user to see which layers are actually visible in the map. A minor legend working as a real legend would be useful.

**Describe the solution you'd like**
Origo version 1 had a separate minor legend that only showed visible layers and background layers in addition to the regular legend. Something similar would have been nice to have in this version as well. But instead of having it separately, the smaller legend could be part of the regular legend.

Thoughts and suggestions for solutions are very welcome!

Grammostola:
Origo version 1's version currently in motion as part of Ödeshög municipality's comprehensive plan https://karta.itsam.se/odeshog/oversiktsplan/

asemoller:
I have made some drafts. I can see pros and cons with all of them. Maybe a mix of the three options is the way forward!
Any thoughts?


1. Two buttons on the map. Minor-legend button next to the close button. Also shows how the minor legend can handle extended styling.

![minor_legend_two_buttons](https://user-images.githubusercontent.com/17123002/135602687-6ce3c88b-f3cb-43a5-84b1-95ac829c67d6.gif)


2. Minimize/maximize next to the close button. If you close when the legend is minimized, it opens minimized.

![minor_legend_minimize](https://user-images.githubusercontent.com/17123002/135602727-6dc70751-5b75-4c99-8802-d3e2b7157eaf.gif)


3. Minnor-legend button next to the turn-off-all-layers button. Always opens the legend.

![minor_legend_row_toal](https://user-images.githubusercontent.com/17123002/135602867-00a2b436-6c86-49e4-8a9f-f2d367eaaccb.gif)

Grammostola:
Nicely visualized! I like number 1 & 3.

filleg:
Also like number 1 and 3. We see a need for this in our version of story maps in Kristianstad. As long as it is possible to configure the minor legend on load or if the state remains within sharemap it would be really useful.

jokd:
Either works for me!

asemoller:
This is what we plan to develop.  
@filleg we'll make this a part of the share map function.


![minor_legend_final](https://user-images.githubusercontent.com/17123002/143934577-64ea64cb-902c-4249-9ffb-0274297ed34a.gif)

sonjahalmstad:
What is happening with this issue? Looks interesting!

[...]

PR #1485, opened 2022-03-03
Origo master feature minor legend

Polleryd:
#1329

New config values for legend: 
visibleLayersControl: boolean - Should visible layers control be active or not.
visibleLayersViewActive: boolean - Sets initial state minor legend.

The above settings, along with the setting "expanded", have also been appended to the share-map functionality.


asemoller:
@tonnyandersson You're absolutely right, it doesn't look good when using the default index.json.
Apparently I haven't tested properly, I'm sorry about that :worried:!

tonnyandersson:
LGTM.

Issue #1484, opened 2022-03-02
Print - export to picture is not working


leafisfalling:
We use WMS in our map (both internally from our own Geoserver and externally generated by ArcGIS from authorities). When we try to export our map with one or more WMS from external authorities to a picture in Print then it does not return any file. But when we turn off the ArcGIS WMS from external authorities then the function save to picture seems work again. 

However, when we export map to pdf, then it works well no matter there is WMS from our Geoserver or WMS generated by ArcGIS from external authorities

1. Turn on a WMS layer generated by ArcGIS (or maybe other map server?)
2. Go to Print
3. Save the map to picture
4. Then it does not return any file back. 

Tested in Window 10, Chrome and Edge, with code for version 2.5

Grammostola:
I tried printing a local ArcGIS WMS layer which worked, I tried https://ext-geodata.lansstyrelsen.se/arcgis/services/hav/P12_Risk_paverkan_specifikt_vatten/ at 300 dpi and that worked as well, as image. What service are you using, how is the layer configured and with what params are you trying to print it?

leafisfalling:
we have the same WMS in the map and it does not work for us to export to picture

![export_ej_bild](https://user-images.githubusercontent.com/36880995/156542274-e1c447a8-5bd8-402c-8e41-1ac4b5b01189.gif)

Grammostola:
Does the web browser console show an error? (f12) 
It looks like that for me too, with the latest master or release, and then Chrome hints at a new download in the lower left or Edge in the top right. It would be interesting to see if Origo's print threw an error for you with the same build and service and layer and if so what that error was.

leafisfalling:
The hints for image file never come up to us. We got this error message:

![image](https://user-images.githubusercontent.com/36880995/156557963-d0128033-d90c-4ae1-9ed3-12fafb77fb28.png)

Grammostola:
With the latest master I defined the layer like so
```json
{
  "name": "P12_Risk_paverkan_specifikt_vatten",
  "title": "P12_Risk_paverkan_specifikt_vatten",
  "group": "WMSer",
  "queryable": true,
  "type": "WMS",
  "visible": false,
  "source": "P12",
  "opacity": 0.5
},
```

from a source like so:
```json
    "P12": {
      "url": "https://ext-geodata.lansstyrelsen.se/arcgis/services/hav/P12_Risk_paverkan_specifikt_vatten/MapServer/WmsServer"
    },
```
and then print just that layer

[...]

PR #1489, opened 2022-03-07
fix: fix geoserver wms theme layer print legend

### LINKED ISSUE #1488: Print: Geoserver theme WMS layers not displayed properly in the print legend
**Describe the bug**
Geoserver WMS theme layers aren't showing up properly in the print legend (they are supposed to be currently supported)

**To Reproduce**
-Have a Geoserver WMS layer with a theme type style, give it a style with an icon source being a getLegendGraphic call and give it an extendedLegend: true property (for it to show up in the Origo legend beyond "show more")
-Open the Print preview

**Expected behavior**
A theme legend for said layer

**Actual behaviour**
The getLegendGraphic theme image is treated as a single symbol
wms_theme

Grammostola:
Aims to fix #1488 

If a layer style definition has the extendedLegend-property (meaning true) the affected function will land at
```javascript
      } else {
        content = await getJSONContent(title);
```

which will for a Geoserver 2.16+ return rules for a pretty theme legend and for others, like an ArcGIS Server with WMS layers, return the title of the layer with no icon, even when printing an Image. #1484

### DIFF
--- a/src/controls/print/print-legend.js
+++ b/src/controls/print/print-legend.js
@@ -122,7 +122,7 @@ const LayerRow = function LayerRow(options) {
       let content = '';
 
       const style = viewer.getStyle(layer.get('styleName'));
-      if (style && style[0]) {
+      if (style && style[0] && (!style[0][0].extendedLegend)) {
         content = getStyleContent(title, style);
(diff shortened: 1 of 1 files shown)


johnnyblasta (approved):
LGTM

PR #1500, opened 2022-03-23
fix: make print attempt to grip the correct styleSheet for rescaling …

### LINKED ISSUE #1483: Print - Scale bar becomes small when change print setting to high resolution (300 dpi)
When we change the resolution for print the scale bar in the preview becomes smaller.

1. Go to "Print" in the menu 
2. Change the resolution to "High (300)"
3. The scale bar becomes much smaller. 

high_low_resolution

Tested in Window 10, Chrome, with the kod for version 2.5

I see there is an issue 1407 about changing print resolution and not sure if this is related to that.

Grammostola:
…purposes. Attempts to fix #1483  . It is uncertain precisely which plugin caused the issue reporter's issue, perhaps the "Help" plugin they employ, it does however seem possible that it works in the same fashion as our version of the layermanager, i.e includes additional styleSheets.

### DIFF
--- a/src/controls/print/print-resize.js
+++ b/src/controls/print/print-resize.js
@@ -48,14 +48,18 @@ export default function PrintResize(options = {}) {
   };
 
   const getCssRule = function getCssRule(selector) {
-    const origoStyleSheet = viewer.getOrigoCSS();
-    const rules = origoStyleSheet.cssRules;
-    for (let i = 0; i < rules.length; i += 1) {
-      if (rules[i].selectorText === selector) {
-        return rules[i];
-      }
+    let soughtRule;
+    try {
+      Array.from(document.styleSheets).some((sheet) => Array.from(sheet.cssRules).some((cssRule) => {
+        if (cssRule.selectorText === selector) {
+          soughtRule = cssRule;
+          return true;
+        } return false;
+      }));
+    } catch (error) {
+      console.warn(error);
     }
-    return undefined;
+    return soughtRule;
   };
 
   const getVisibleLayers = function getVisibleLayers() {
--- a/src/controls/print/print-resize.js
+++ b/src/controls/print/print-resize.js
@@ -48,7 +48,8 @@ export default function PrintResize(options = {}) {
   };
 
   const getCssRule = function getCssRule(selector) {
-    const rules = document.styleSheets[0].cssRules;
+    const origoStyleSheet = Array.from(document.styleSheets).find(sheet => sheet.href && sheet.href.includes('css/style.css'));
+    const rules = origoStyleSheet.cssRules;
     for (let i = 0; i < rules.length; i += 1) {
       if (rules[i].selectorText === selector) {
         return rules[i];
--- a/src/controls/print/print-resize.js
+++ b/src/controls/print/print-resize.js
@@ -48,7 +48,7 @@ export default function PrintResize(options = {}) {
   };
 
   const getCssRule = function getCssRule(selector) {
-    const origoStyleSheet = Array.from(document.styleSheets).find(sheet => sheet.href && sheet.href.includes('css/style.css'));
+    const origoStyleSheet = viewer.getOrigoCSS();
     const rules = origoStyleSheet.cssRules;
     for (let i = 0; i < rules.length; i += 1) {
       if (rules[i].selectorText === selector) {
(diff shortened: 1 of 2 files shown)


Grammostola:
Have recevied feedback from the reporting organization that the fix appears effective for their plugin.

jokd:
Probably works but feels a bit risky to hard code the href of the style. There can be multiple similar stylesheets or it can have another name. 
Another approach would be to hard code the css rules (that rarely changes) or to use window.getComputedStyle and populate it on map load. 
But for now maybe this will work.

tonnyandersson:
Perhaps a stupid question, but why not iterate/filter/map/whatever all available stylesheets in the print-resize `getCssRule` function to find the appropriate CSS rule? That way, you wouldn't have to hardcode anything to single out the main Origo stylesheet.

Grammostola:
That's possible, it can be as little as another for-loop. Since the number of document.styleSheets likely isn't going to change during runtime it seems logical to establish the correct one once and then use that for the multitude of getCssRule calls that are made during print so as not to potentially look in the wrong one/s first over and over.
 Though it may not make a user noticable difference unless there's a slew of styleSheets with a million rules each, I don't know.

It's equally not so pretty to attempt to grip the correct styleSheet via a rule name prefix as it can break, though perhaps not so likely.

tonnyandersson:
All it takes for the rule name gripping scheme to break is the introduction of another stylesheet with the same rule name in it, loaded before the main Origo style. Probably not _extremely_ likely, but not _extremely_ unlikely either. With plugins utilizing their own styling or organisations implementing their own themes, I'd say there's at least a slight risk. And the sneaky nature of this bug kinda makes it even worse.

But you're right, there's an awful lot of calls to `getCssRule` that might have a negative impact on the user experience, so it would be nice to single out the right one once. But I'm not sure it's worth the potential risks.

Not asking you to make any changes at this moment. Just wanted to bring the question to the table.

Grammostola:
Indeed. Since this is now as per prior suggestions a draft I will report that during very limited and subjective testing all those calls with one extra styleSheet with 80 rules (perhaps not an uncommon scenario) doesn't make the print control noticably slower. It is perhaps the slowest component in the system depending on lots of things. Potentially the slowest anyway so slivers of a second are in some likelihood fishfood. Though I haven't seen real beefy styleSheets,

We could do 
```javascript
    let soughtRule;
    Array.from(document.styleSheets).some((sheet) => Array.from(sheet.cssRules).some((cssRule) => {
      if (cssRule.selectorText === selector) {
        soughtRule = cssRule;
        return true;
      } return false;
    }));
    return soughtRule;
```

I think it looks slightly less like the 80s compared to:

```javascript
    for (let y = 0; y < document.styleSheets.length; y += 1) {
      for (let i = 0; i < document.styleSheets[y].cssRules.length; i += 1) {
        if (document.styleSheets[y].cssRules[i].selectorText === selector) {
          return document.styleSheets[y].cssRules[i];
        }
      }
    }
 return false;
```

johnnyblasta:
I have a problem with cssRules when the Origo stylesheet is on another webserver:

`[Exception: DOMException: Failed to read the 'cssRules' property from 'CSSStyleSheet': Cannot access rules at CSSStyleSheet.<anonymous> (<anonymous>:1:29)]`

Grammostola:
Indeed, as expressed by FF: `DOMException: CSSStyleSheet.cssRules getter: Not allowed to access cross-origin stylesheet`. Maybe it has to do with https://drafts.csswg.org/cssom/#concept-css-style-sheet-origin-clean-flag and the creation of the CSSStyleSheet (rather than when there's an attempt to access its cssRules); it applies whether there are multiple stylesheets present in the document or not. It seems a different issue even if this is likely the only function accessing stylesheet rules and as such in need of a fix for that issue. Whether it should be addressed here depends on how quickly we want the proposed fix for the current problem.

Grammostola:
I've changed it as proposed above and with a try/catch that makes the map not crash when the origo stylesheet is remote in order to somewhat handle @johnnyblasta 's scenario. The print control becomes pretty verbose in this instance as this function is called a lot of times but if the dev tools aren't open none's the wiser I think. (The scale bar still becomes small then as described in the related issue)

jokd (approved):

PR #1505, opened 2022-03-24
Auto hide layers and save button in editor

### LINKED ISSUE #1494: Editor - hide tool buttons
If there are only one layer that is editable is it necessary to have the button for selecting between layers visible or should it be hidden then?
If the editor control is set to autoSave is there a point of having the save button is visible?

If only one editable layer no layer select button. If autosave no save button.

A possibility is to have it configurable if the buttons should be hidden

johnnyblasta:
Hides layer selector button when only one layer is editable.
Hides save button when autosave option is true.
Closes #1494

### DIFF
--- a/src/controls/editor/editortoolbar.js
+++ b/src/controls/editor/editortoolbar.js
@@ -174,6 +174,14 @@ function init(options, v) {
   viewer = v;
   editHandler(options, v);
   render();
+  // Hide layers choice button if only 1 layer i editable
+  if (editableLayers.length < 2) {
+    $editLayers.classList.add('o-hidden');
+  }
+  // Hide save button if configured with autoSave
+  if (options.autoSave) {
+    $editSave.classList.add('o-hidden');
+  }
   editorLayers(editableLayers, {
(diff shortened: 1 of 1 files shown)


steff-o (changes requested):
Tested in Chrome on Windows. There is a wider gap between the save button and the button next to it if layer selector is hidden. Probably because the layer button is inside the popover-container. Hide that div instead removes the gap.

image

johnnyblasta:
> Tested in Chrome on Windows. There is a wider gap between the save button and the button next to it if layer selector is hidden. Probably because the layer button is inside the popover-container. Hide that div instead removes the gap.
> 
> image

Made a fix that hides popover-container instead which makes the extra wide gap disappears

steff-o (approved):
LGTM. Still there are some strange stuff going in in the editor toolbar, but that's not because this PR. The add feature button is also wrapped in a popover-div that separates it from the other buttons a bit and the editor layer popover contains a _body_ tag. But both those have been there for ages.

PR #1509, opened 2022-03-30
Default values in editor

### LINKED ISSUE #1495: Editor - default values from session variables
When creating and editing features it is often desirable to add who did it and when.  This could also be applicable to other attributes that the application has knowledge of.

It should be possible to configure a sessionStorage key or a localStorage key or current time as default value for an attribute. I propose that it can be configured using something like below:

```
"attributes": [
    {
      "name": "last_update_user",
      "type": "text",
      "readonly": true,
      "defaultValue": {
        "type": "session",
        "key": "loggedInUser",
        "alwaysUpdate": true
      }
    },
    {
      "name": "last_update",
      "type": "hidden",
      "defaultValue": {
        "type": "timestamp",
        "timeStampFormat": "time",
        "alwaysUpdate": true
      }
    }
  ]
```

The "defaultValue" configuration already exists (undocumented) but only support static values and are only applied when features are created, not updated. I suggest adding a configuration object (which also could be called defaultValue) that further can specify how to handle more complicated defaults. The key ingredients are:
1. The "type"-option. One of "sessionStorage", "localStorage" or "timestamp". Depending on the type additional configuration must be present to specify the storage key or time format. By reading from sessionStorage or localStorage, the application can set any value it wants to any attribute.
2. the "alwaysUpdate"-option. If specified the default value is used to update the edit form with the default value. As the value is only set in the attribute editor it will only actually be saved when user saves. It can be combined with attribute type "hidden" (existing but not documented) and the readonly-flag to make the user see the attribute or not and prevent it from be edited.

Utilise the default date functionality, but that only applies to newly created features (and datetime obviously).
Use a database solution. But as the WFS server usually connects to the database using an application user it would not get the actual user. For timestamp the feature is not re-read after save (which would be another feature request), so it won't be available to origo until layer is refreshed. Also it would not support arbitrary attributes.

**Additional context**
The dialogue plugin has something similar, but it only sets the defaults on create (due to its nature) and does not support editing proposed values.

steff-o:
Closes #1495. 

Adds possibility to update feature variables from localStorage, sessionStorage or setting a timestamp. Configuration basically as in issue #1495, but `alwaysUpdate` is renamed to `updateOnEdit`  and added `useUTC` for timestamps to select between local time and UTC.

If type is timestamp it is possible to set a format depending on what the timestamp is for.
Time formats supported:
- `time` : hh:mm:ss. Used when database field is _time_ or _text_
- `date`: YYYY-MM-dd. Used when database field is _date_ or _text_
- `datetime`: "YYYY-MM-dd hh:mm:ss" Used when database field is text and you want a readable string.
- `timestamp`: Default. Can be parsed by most iso date time parsers. Used when database field is DateTime (or similar) or when using a _date_, _time_ or _datetime_ attribute type in editor.

As a bonus, the `hidden` attribute type is fixed.

### DIFF
--- a/src/controls/editor/editform.js
+++ b/src/controls/editor/editform.js
@@ -146,7 +146,9 @@ const createForm = function createForm(obj) {
       el = `<div class="validate ${cls}"><label>${label}</label><br><input type="number" step="0.01" min="0" name="decimaltal" id="${id}" value="${val}"${readonly}${required}></div>`;
       break;
     case 'hidden':
-      el = `<input type="hidden ${cls}" id="${id}" value="${val}">`;
+      // Note that an input with type="hidden" is not the same as an input with class="o-hidden". Type hidden is to roundtrip values invisible to the user
+      // class o-hidden is to temporarily hide an input as logic in view says it should not be visible right now (batch edit or constraints).
+      el = `<input class="${cls}" type="hidden" id="${id}" value="${val}">`;
       break;
     default:
       break;
--- a/src/controls/editor/edithandler.js
+++ b/src/controls/editor/edithandler.js
@@ -112,11 +112,58 @@ function getFeaturesByIds(type, layer, ids) {
   return features;
 }
 
+/**
+ * Helper that calculates the default value for one attribute
+ * @param {any} attribConf The list entry from "attributes"-configuration that default value should be calculated for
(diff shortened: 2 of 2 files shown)


johnnyblasta (approved):
LGTM

Issue #1020, opened 2020-11-12
Alternative styles for layers


asemoller:
**Is your feature request related to a problem? Please describe.**
Sometimes a user wants to be able to change a style for a layer.

**Describe the solution you'd like**
It would be a nice feature if a user can choose between several different styles on a layer. For example, if a layer have a black style as standard, you should be able to choose a yellow style if you change the background map to orthophoto. Another application could be to show thematic styles from different attributes for a layer in the map.

Grammostola:
We are currently looking at adding functionality like this via one of our existing plugins. It's intended only for WMS but would partially share bits with equal functionality for client-based styles.

asemoller:
We have a solution for this BUT it´s only for client styling (WFS, geojson). I'm not sure it should belong to the core since it isn't working for server based layers (WMS).

Our solution looks like this:

![stylepicker-issue](https://user-images.githubusercontent.com/17123002/156168403-e2e705bb-bbcb-4fe6-845e-04e19a7e3f3f.gif)

Let us know if this is something that should be in the core. If not, I'll close this issue.

jokd:
This is definitely something for the core. WMS support can be added later on.

PR #1510, opened 2022-03-31
Style picker for client stylin

Polleryd:
#1020

In the configuration file, add the field "stylePicker" to the layer. It should be an array with the styles available for the user to choose from on that layer. 

"stylePicker": [
        {"title": "Title_1", "style": "style_1"},
        {"title": "Title_2", "style": "style_2"}
      ],


asemoller:
Just to clarify, this works for vector layer (WFS, geojson) with standard client styling. Doesn't work with group layers or cluster styling.

steff-o:
@Polleryd Please edit your comment to add a magic word before the issue number in order to link the PR to the issue: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

johnnyblasta (commented):
Could be nice to have a label, clarifying what the dropdown does.
image

Don't know what I think about the blue ring around the icon when there is different styles. Is it useful or irritating.
image

tonnyandersson:
I agree with @johnnyblasta, a label would be nice.

Also, the blue ring feels a bit too much. It's not immediately clear what it means and that might be a bit confusing for the average user.

Grammostola:
Nice, at least minus the blue ring. In addition the style picker appears to fail after the Print control has been opened and closed. I don't see an error in the console but for the default origo cities layer switching between styles efter Print control exit doesn't appear to do anything anymore. The better news is that #1514 appears to fix that in my brief and preliminary test.

steff-o:
@asemoller I also object to the blue ring. It would only clutter up the GUI. Soon someone else will want a yellow underscore to indicate that the layer is queryable, a red square to indicate that data is only accurate on thursdays etc...

The users will soon learn which layers must/can be changed when background changes. Or else they can always try the three dot menu and see what is possible to do with each layer. Most likely there will appear more stuff there (like zoom to extent) and the options will differ from layer to layer so the users will soon get a habit of try that menu to see what is possible to do for each layer.

johnnyblasta:
As a did take on the role of accessibility "chief" I have to point out that the dropdown style button is missing aria-label or aria-labelledby (which should point to the missing label if added)

asemoller:
ok, we got it! 😄

We will do this:
- Remove the indicator
- Add "Välj stil" as a label above the dropdown
- Add aria-label and aria-labelledby
- Add changed style to mapstate (if there's time for it)

tonnyandersson:
@steff-o made a good point regarding GUI clutter. And also, blue ring or not, you'd still have to _tell_ the user that there are multiple styles available, either by explaining the meaning of the blue ring or by saying that some layers may have multiple styles, click the _More_ button and find out if you think a layer isn't clearly visible. Just a blue ring is not very intuitive.

That said, I'm not _completely_ against hints like this, @asemoller. I just don't think there's a point having it unless it's somewhat clear what it means. Especially in this case since it "changes" the look of the layer symbol for no _obvious_ reason. So, I'm inclined to say that the blue ring is something you should implement locally instead of adding it to the core. But I also realize that others _might_ find it useful is some cases, so perhaps it could be a configurable feature, off by default. But you'd have to make sure it doesn't eff up the styling of the rest of the legend by pushing pixels around, and that it works well with the tab-index stuff.

asemoller:
@tonnyandersson we'll remove it, no worries!

In fact, some icons are not completely in the middle and that disturbs the symmetry 😮 and that's never good!
I really enjoyed the discussion though 😄 you all have good points!

asemoller:
We have fixed all the points listed above (including mapstate). And it also works well to change one cluster style to another cluster style.

tonnyandersson:
@Grammostola Have you had a chance to test the mapstate thing since the latest update?

Grammostola:
Yes, wfs and geojson and simpler styles at least, worked well.

johnnyblasta:
There are still some wrong use of accessibility tags.
If using aria-labelledby it should point to a id of element that labels the button-element in this case.
[Read more on how to use it](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby)

asemoller:
@johnnyblasta Is there any good example in origo that we can look at?

src/controls/legend/overlayproperties.js
```diff
@@ -43,14 +57,52 @@ const OverlayProperties = function OverlayProperties(options = {}) {
...
     onInit() {
-      this.addComponents([transparencySlider]);
+      styleSelection = Dropdown({
+        direction: 'up',
+        cls: 'o-stylepicker text-black flex',
+        contentCls: 'bg-grey-lighter text-smaller rounded',
+        buttonCls: 'bg-white border text-black box-shadow',
+        buttonTextCls: 'text-smaller',
+        text: getStyleDisplayName(layer.get('styleName')),
+        buttonIconCls: 'black'
```
johnnyblasta, line 93: Add an ariaLabel to the component by adding this row:

```
,
ariaLabel: 'Välj stil'
```

src/controls/legend/overlayproperties.js
```diff
@@ -66,10 +118,23 @@ const OverlayProperties = function OverlayProperties(options = {}) {
           layer.setOpacity(sliderEl.valueAsNumber);
         });
       }
+      if (hasStylePicker()) {
+        styleSelection.setItems(stylePicker.map(altStyle => altStyle.title));
+        const styleSelectionEl = document.getElementById(styleSelection.getId());
+        styleSelectionEl.addEventListener('dropdown:select', (evt) => {
+          onSelectStyle(evt.target.textContent);
+        });
+        styleSelectionEl.setAttribute('aria-label', 'Välj stil');
```
johnnyblasta, line 127: Remove these 2 rows below, since the aria-label is set when dropdown is created above.
The aria-labelledby should have been the id of  the html that is inserted by function renderStyleSelection() but since there is already a aria-label present it is enough.
```
styleSelectionEl.setAttribute('aria-label', 'Välj stil');
styleSelectionEl.setAttribute('aria-labelledby', 'Välj stil');
```

johnnyblasta (approved):
LGTM

PR #1512, opened 2022-04-01
Adds a abort button to editor form if autosave and autoform is true

### LINKED ISSUE #1508: Editor cancel adding new feature
It would be nice to be possible to cancel adding a new feature without leaving a empty feature which happens when using close button when using editor control in autosave/autoform mode.

There should a save and cancel button in edit form. Clicking cancel should reove feature from map. Save button should save the attributes to the feature.
image

johnnyblasta:
Closes #1508

### DIFF
--- a/src/controls/editor/editform.js
+++ b/src/controls/editor/editform.js
@@ -25,14 +25,14 @@ const createForm = function createForm(obj) {
   const maxLengthText = maxLength ? `, max ${obj.maxLength} tecken` : '';
   switch (type) {
     case 'text':
-      el = `<div class="validate ${cls}"><label>${label}</label><br><input type="text" name="text${maxLengthText}" id="${id}" value="${val}"${maxLength}${readonly}${required}></div>`;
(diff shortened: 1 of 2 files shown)


steff-o (commented):
I have tested with related tables and it works when both _autoSave_ and _autoForm_ is true. As the abort function is only added to the main feature it is not necessary to do anything special for related tables. It won't be possible to  abort a child layer creation in this way, but I don't think that it is necessary to support that. It will only confuse the users (and developers). Aborting the main feature will also delete any newly created child features, so that is also OK.

On the other hand, when _autoSave_ is false it crashes on abort if the feature has related layers configured. I'm not sure that it is this PR that is the culprit, it may be an earlier undiscovered bug in related tables. It think the crash occurs because when _deleteFeature_ tries to delete a feature that has not been auto saved, it first tries to read related related objects using the temporary fid as CQL. If the FK-field in the child is an integer, the server won't be able to parse the temporary fid (which is a guid) and crashes instead of returning an empty set. The remedy would be to add a check in _deleteFeature_ to check if the feature is saved to server before trying to delete its children.

Issue #1308, opened 2021-08-10
GetFeatureInfoRequest does not work for tainted canvases


steff-o:
**Describe the bug**
Possibly not really a bug, but getfeatureInfoRequest in getfeatureinfo.js will not behave as expected if the browser does not support crossOrigin for canvases. It could be a bug or just a matter of dead or reduntant code.

The problem is that the _isTainted_ function will always return _false_, so all logic that deals with tainted canvases will never be run, so it will run the last else statement, which does the same thing as when _crossOrigin_ is supported.

In the end it may not make any difference to the functionality as no sane person is using IE 9, but it makes the code unreadable and unnecessary large.

**To Reproduce**
I haven't actually run in to this problem, only discovered by code inspection.

**Expected behavior**
The code should be easy to follow and behave as expected. Right now it appears to handle tainted canvases differently, but it doesn't.

Either the _isTainted_ function and _layerAtPixel_ should be fixed, or removed all together if it is not applicable anymore.

PR #1513, opened 2022-04-04
Ol v6141

### LINKED ISSUE #1423: OL 6.12

jokd:
Fixes #1423 and #1308 
Due to a soon to be deprecated method I made some changes in files modified by #1292 so please @steff-o have a look at the changes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -161,9 +161,9 @@
       "integrity": "sha1-zlblOfg1UrWNENZy6k1vya3HsjQ="
     },
     "@mapbox/mapbox-gl-style-spec": {
-      "version": "13.23.0",
-      "resolved": "https://registry.npmjs.org/@mapbox/mapbox-gl-style-spec/-/mapbox-gl-style-spec-13.23.0.tgz",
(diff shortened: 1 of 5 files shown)


jokd:
No use testing this quite yet, förgot one aspect. Will update this PR ASAP

steff-o (commented):
Tested mainly the cursor hover thing and it fails for image layers that are not tiled.

I have only tested basic featureInfo on the simplest setup with one WMS layer and one WFS layer hosted on my local server, and it seems to work as intended by popping up a info window if there is a feature there and not create a remote request unless there is a symbol there. 

As a consequence this PR will affect the way featureInfo works as the previous implementation very often created a remote request to a layer even if there were no pixels visible at that pixel as several layers shared canvas, which in effect could give you hits in layers with invisible symbology. That is no longer possible if someone exploited that flaw. Also the previous implementation had a pixel tolerance, which this implementation lacks. It may not affect anything as when I last tested i couldn't make the tolerance setting do any difference.

jokd:
Now I've added support for ImageLayers as well. 
Regarding "hit tolerance" it is not supported (unless we want to make request no matter what)

johnnyblasta (commented):
Tested with LMSearch plugin since I was a bit worried that the new OL version might cause a problem, but it seems to work fine so no worries.

PR #1514, opened 2022-04-06
Fix for #1476 and #1482 (correct styling upon exiting print view)

### LINKED ISSUE #1476: Icons rescaled after print
**Describe the bug**
When using scaled symbols in the map, the scale seems to be overlooked when re rendering the map after the print preview.
Eg:
```
		"symbol": [
			[{
				"icon": {
					"size": [64, 64],
					"src": "icon.png",
					"scale": 0.5
				}
			}]
		],
```

**Screenshots**
Before print:
image

After:
image

Ideas on this @Grammostola ?

### LINKED ISSUE #1482: Print - incorrect style for highlighted/selected objects after print
**Describe the bug**
When you return to the map after printing, the style of the selected item is incorrect.

**To Reproduce**
1. Go to print
2. Close print
3. Select an object

**Expected behavior**
The selected object should have the style (or default style) from the json file.

**Screenshots**

print-selected-style-issue

**Additional context**
Related to https://github.com/origo-map/origo/issues/1476?

sweco-sedalh:
This PR contains fixes #1476 and fixes #1482.

It properly "resets" scaling by using the existing `setLayerScale` function using the "original" DPI value (currently hardcoded to 150 as in other parts of the code).

Selection style is corrected by setting the layer style rather than the style of individual features (fix written by @sweco-sefaed).

### DIFF
--- a/src/controls/print/print-resize.js
+++ b/src/controls/print/print-resize.js
@@ -412,39 +412,9 @@ export default function PrintResize(options = {}) {
     }
   };
 
-  // "Resets" layer by changing scale to 1 or removing DPI parameter
+  // "Resets" layer by removing DPI parameter
   const resetLayerScale = function resetLayerScale(layer) {
     const source = layer.getSource();
-
-    if (isVector(layer)) {
-      const features = source.getFeatures();
-      if (features && features.length) {
-        const feature = features[0];
-
-        // Remove styles instead?
-        const styles = feature.getStyle();
-        const scale = 1;
-        if (Array.isArray(styles)) {
-          styles.forEach(style => {
-            const image = style.getImage();
-            if (image) {
-              image.setScale(scale);
-            }
-
-            const stroke = style.getStroke();
-            if (stroke) {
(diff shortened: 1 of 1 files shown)


leafisfalling:
How is it going for this issue? :)

sweco-sedalh:
It's waiting for review by a maintainer.

Grammostola:
We (the sponsoring organization) have of course tested it and it appears to solve the two issues referred to. It would be great if others could check it too, as both issues were found by others and neither issue was reproduced by our normal usage patterns. There is a potential discussion point around 150dpi.

steff-o:
Tested with the simplest of possible configurations. It solves the bugs as described in the linked issues and passes lint. So for me it is OK.

Regarding the potential discussion around 150 dpi: it is unfortunate that this PR keeps building on the antipattern of using constant values scattered in the code, but that was a problem that was already there so I think that is out of the scope for this PR anyway.

johnnyblasta (approved):
Tested and seems to solve the issues, so lets get it approved then.

PR #1515, opened 2022-04-06
Editor searchList support dynamic search

### LINKED ISSUE #1498: searchList not working in editor
**Describe the bug**
The searchList option in the attribute editor is not working.

**To Reproduce**
1. Configure an application to have a searchList with some options.
```
"attributes": [
        {
          "name": "Fritext",
          "title": "Fastighetsbeteckning",
          "type": "searchList",
          "list": [
            {
              "value": "Sel"
            },
            {
              "value": "Stan"
            },
            {
              "value": "Obygden"
            }
          ],
          "config": {
            "minChars": 2,
            "maxItems": 15
          }
        }
      ]
```
3. Edit the attributes of a feature
4. Close the attribute form
5. Edit attributes for the same feature
6. The drop down arrow is in the wrong place and no suggestions appear

Alternatively use autoform = true and create a feature.

**Expected behavior**
Well, it should work

**Screenshots**
image

**Additional context**
Digging into the code it looks like it is the call to `new searchList` in edithandler should be called to init all current searchFields as awesomeplete-inputs, but that 
1. Does not happen for autoform
2. Does not happen for child forms.
3. Otherwise only happens once in a lifetime.
4. Uses same id on several html tags. Which most likely works, but if there are several serachList, it would break the semantic meaning of "id" in html.
5. Roundtrips configuration in custom html attributes, not named data-, which also probably works but it not valid html.

In my opinion the entire handling seems complex beyond reason and the edit form should set up everything without relying on some other functions having to be called afterwards to finish up. That would save us some bytes and clock cycles as well.

### LINKED ISSUE #1499: Editor - Add service URL as list item source for searchList
When using the `searchList` type in the editor one must add add all possible options to the configuration. This is impractical and inefficient if there are many options or values that changes often. Examples of the first case is street names, and the second case is to get suggestions of the already existing values in the database to avoid unnecessary almost-duplicates.

Also the should be a way to limit the user to only enter the values already present in the list, which effectively makes it a searchable drop down. 

It should be possible to configure a source URL from where the possible list values are fetched.
There should be an option to query the service for each keypress or just once.
There should be an option to only allow the user to enter the values already present in the list.

Add the list server side to index.json before the application is loaded, but that is is still impractical for large lists and does not work well with constantly changing data. 

By using `dropdown` instead of `searchList` it is possible to force the user to only enter legal values, but with many options a dropdown is impractical as the search function in a dropdown sucks.

**Additional context**
This is similar to #1295 but with more focus on searchList and dynamic loading.

steff-o:
Closes #1499 and fixes #1498 

Actually I ended up rewriting almost the whole shebang as the previous implementation had some bugs and the easiest way to get rid of them was to do it in a another way instead of try to patch it. Most stuff is backwards compatible. 

**New Features**
- The searchList can be configured to query a server for suggestions after each keypress
- The list of options can be fetched from a remote server when the dialog is opened
- Possibility to only allow values that is in the list (without using required)
- Added help messages when no suggestions are found

**Fixed Bugs**
- The searchList works more than the first time the attribute editor is opened
- The searchList works for related tables
- The searchList works with autoForm
- Some glitches in the help text container (which previously only was empty to denote no matches).

**Breaking Changes**
- The validation is NOT backwards compatible. Previous implementation assumed that if the input was required it also implied that the value must be present in the list of items. I split that in to two different configurations and to make it more intuitive the default behaviour of _required_ does not imply that the value must be present in the list of items. I figured as the usage of searchList is probably not very widespread as it has not been working for a while it wouldn't matter.
- Arbitrary awesome configuration can not be passed. Only documented parameters can be set. Previous implementation used `Object.assign()`, but I removed that to only support selected options.
- The DOM elements have changed some classes and id-field and possibly types so if someone has done some advanced DOM querying that may break;

**Remaining Issues**
- If the drop down does not fit in the modal, it does not look good, but works.

**Usage**
Earlier configuration should work with the exception listed in breaking changes. But as the documentation is a bit vague, here comes a complete example of all possible configurations.

All possible configuration
``` javascript
{
          "name": "fritext",
          "title": "Fastighetsbeteckning",
          "type": "searchList",
          "maxLength" : 123,
          "required": true,
          "list": [
            //{
            //  "value": "Sel",
           //   "src": "http://localhost/static/sel.jpg"
            //},
            //{
            //  "value": "Stan"
            //},
            //"Obygden"
            //,
            {
              "location": "http://localhost/static/",
              "extension" :  "png"
            }
          ],
          "config": {
            "url": "http://localhost:3000/dynamic",
            "queryParameter": "input",
            "dynamic": true,
            "allowOnlyFromList": true,
            "disallowDropDown": true,
            "typeMoreText": "Skriv då",
            "noHitsText": "Go fish",
            "minChars": 2,
            "maxItems": 3
          }
```
Properties set directly on _attribute_ level is supported as in documentation with the following exceptions/explanations:
- `required` Optional. true if the input must contain a value. Previous implementation assumed that the value also must be in the list of suggestions.
- `maxLength`: Optional. Max length of input. defaults to 50. I didn't touch that one, it just was not documented.

The `list` property works like before, I haven't messed at all with that:
It is a list of objects which can either have:
- A `value` and optionally a `src`. The src is an url to an image to display next to the value. The _src_ is relative to the application no matter where configuration comes from, or absolute.
- A `location` and a `extension` property. The location is an url to a HTML document that contains links to image files, typically a file listing page as generated by a webserver. All `a` tags in the document with the _extension_ extension is used as list values. The filename is the value, and the image itself is displayed next to the value in the drop down. _location_ can not be combined with static list items and can not appear in a list returned from a server using the _url_ option. This is exactly as before.
- An array of strings which is used as labels.

The `config` property can contain the following options:
- `url` Optional. An url to a service that will provide a list object. If specified the static `list` configuration is ignored. The server should return a JSON list as described above with the exception that `location` is not supported as that would be ridiculous. It could be a static file or dynamically generated content.
- `dynamic` Optional. if true the `url` is queried for each key press in the input. The server should accept a query parameter which will receive the current value of the input field. The server should return a list of matching search items. The exact algorithm is up to the server, but origo will assume the list is filtered and sorted so the list will be displayed as is as suggestions, but will be truncated to `maxItems`. If the server's algorithm is not based on substrings of the value, the highlighting may not work as expected. 
- `queryParameter` Optional The name of the query parameter to send when `dynamic`is true. Defaults to `input`
- `allowOnlyFromList` Optional. true if the user only can input values from the list. Replaces the _required_ validation for list values.
- `disallowDropDown` Optional. If _true_ the very special function to override minChar when input is empty. If not specified, the user can click the down arrow (or enter) when input is empty and get all possible suggestions. Not sure why, but is was there before and seems to sort of mimic a _select_ input. Should most certainly be set to true when using _dynamic_
- `typeMoreText` Optional. The text to show to the user if the input has less then _minChar_ characters. Defaults to "Skriv fler tecken".
- `noHitsText`  Optional. The text to show to the user if there are no suggestions to show. Defaults to "Inga träffar".
- `minChars` Optional. Minimum number of characters required to get a suggestion. Defaults to 2.
- `maxItems` Optional. Maximum number of suggestions to display. Defaults to 10.

Here's the code to a test server to be used as url. It works both for dynamic mode and non-dynamic:
```
const express = require('express')
const app = express();

const listan = ["bertil", "ansgar", "angus", "bengt", "caesar", "cesar", "the captain"];


app.use(function (req, res, next) {
  res.header("Access-Control-Allow-Origin", "*");
  res.header("Access-Control-Allow-Headers", "Origin, X-Requested-With, Content-Type, Accept");
  next();
});

app.get('/dynamic', function (req, res) {
  const input = req.query.input;
  let result = listan;
  if (input !== undefined) {
    result = listan.filter(item => item.includes(input)).sort().map(item => ({ value: item }));
  }
  console.log(input);
  setTimeout( () => res.json(result), 1000);
});


app.set('port', process.env.PORT || 3000);

var server = app.listen(app.get('port'), function () {
  console.log('Express server listening on port ' + app.get('port'));
});
```


steff-o:
As a bonus I added a utility spinner and moved over infowindow to use it. I had plans to make a CSS spinner, but I took the gif from infowindow to avoid messing with the layout there.

Issue #1025, opened 2020-11-16
WFS layer doesn't respect extent in GetFeature request


jokd:
**Describe the bug**
When limiting a layer with the extent prooperty, this is not respected when making the GetFeature request. 

**To Reproduce**
Set extent on layer level in the config
Pan outside the extent area
Activate layer
Request will not consider the extent bounds

**Expected behavior**
No requests for features outside the extent

jokd:
This may have been solved in OL 6.6

jokd:
I can confirm that this is solved in OL 6.6.1 so I close this issue

PR #1517, opened 2022-04-07
Added Zoom To Layer Extent button

### LINKED ISSUE #1516: Zoom to layer extent button
Now with the nifty lil' _More_ button in place, it actually makes sense to add a Zoom To Layer Extent feature.

I'm thinking that the _More_ menu would be the perfect place to put a Zoom To button. It would probably look quite similar to this:
zoomto

Another (not so refined) approach would be to add the button to the layer row in the legend, next to the on/off switch or something, totally cluttering up the legend. No good.

**Additional context**
This feature will probably not work perfectly with tile based (or similar) layer types since OL doesn't have a clue about the extent of actual _features_ in such layers.

tonnyandersson:
Fixes #1516.

Just add `"zoomToExtent": true` to any non-background layer.

**Known issues:**

- Doesn't work super well with raster type layers like WMS, AGS_MAP etc. since such layers don't contain any _actual_ features. Still works but zooms to the layer extent, whatever that is, rather than source extent.
- For vector layers, this feature will zoom to the full extent of the _source_, even if there's a specified layer extent in the config. It would, however, be quite possible to extract such extent from the config and use that instead of the source extent. Perhaps that would make more sense?


johnnyblasta (approved):
Tested with setting the extent in config for a WMS layer and it works. Perhaps a bit too much work to enter the correct extent manually for every WMS layer.

steff-o:
At least for vector layers I think the most usable extent would be the actual extent where there are features, not a predefined extent.  If using BBOX it would require a roundtrip to the server and will result in a feature download bonanza that totally defeats the idea of BBOX. But maybe that sorts under the "suit yourself" tab if you configure a layer to allow zoomToExtent if it is too heavy for strategy all.

tonnyandersson:
You are absolutely right, @steff-o.

However, if you set an `extent` on a vector layer, _all_ features seems to be loaded, regardless of strategy. The map only _shows_ the features inside the extent, but all the others are there, lurking around invisibly. You can even click on them if you know where they are... Seems like a bug, possibly related to the supposedly fixed #1025. Haven't done any deeper digging, though.

So, right now, if you set an extent on a vector layer, the map will _show_ the features inside the extent only, but the Zoom To feature will zoom to the _full_ layer extent including the invisible features.

PR #1524, opened 2022-04-19
Fix lint errors

### LINKED ISSUE #1451: Fix lint errors
**Describe the bug**
After the webpack update the _lint-run_ and _lint_ tasks produces error on the code base. 

**To Reproduce**
1. `npm run lint-dev` or `npm run lint`
2. There are errors. Funny enough, `lint-dev` produces more errors on practically the same configuration.

**Expected behavior**
No errors.

**Additional context**
The lint check in GitHub still allows these new errors. Probably it would be best to actually fix the errors in the code, and consider updating GitHub rules to match the ones in the tasks. Alternatively revert to the old lint settings.

steff-o:
Fixes #1451 by obeying som rules. Also fixed that all files are linted independent of which npm script is run. (Sort of, `npm run lint` will lint all files in the src tree except those ignored in _.eslintignore_, the other scripts will lint what is actually bundled honoring _.eslintignore_ )

I also added some stuff to the documentation on how to lint.

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -8,7 +8,11 @@ Origo uses OpenLayers as the map component. Development in Origo shall support v
 When developing, progressive enhancement is an encouraged strategy. That means that it is allowed to adopt new technology that may not be widely supported today but enhances the user experience in modern web browsers under the condition that the functionality is not critical.
 
 ## JavaScript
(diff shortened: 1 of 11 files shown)


johnnyblasta (approved):
LGTM

PR #1533, opened 2022-05-02
Editor checkbox value

### LINKED ISSUE #1519: Editor - Checkbox set value
An attribute with type "checkbox" will always update the database with 0 or 1 depending och check status. When displaying the values from database in another place, "0" and "1" may not be intuitive to the user.

There should exist a configuration which value should be sent to the server when the checkbox is checked and one for when it's not checked.

Use dropdown instead, but that requires one more click for the user.

Do some SQL tricks, but I don't want to do that for every table.

Switch the value when presented, but I don't want to do that for every possible place where data is presented.

**Additional context**

### LINKED ISSUE #1520: Editor - Checkbox alignment
Checkboxes have the label to the left of the checkbox. It does not look good when stacking checkbox attributes on top of each other.

This is how it looks like now:
image

Checkbox should be to the left, labels to the right.

This is how it should look like:
image

Requesting for an configuration where to place the label, but I figure that it won't be necessary as everyone would like to have the checkbox to the left.

**Additional context**

steff-o:
Closes #1519 and fixes #1520.

Adds the possibility to add which values a checkbox should use for checked and unchecked. As a bonus the checkbox is moved to the left of the label.

checkedValue defaults to 1 (backwards compatible)
uncheckedValue defaults to 0 (backwards compatible)

It is possible to only specify one of checkedValue and uncheckedValue and make it work, but I wouldn't recommend it as for backwards compatibility the logic interprets any truthy value as checked unless uncheckedValue is specified, but on save a specific value is set.

```json
  {
          "name": "fritext",
          "title": "Check me out",
          "type": "checkbox",
          "config": {
            "uncheckedValue": "okryzz",
            "checkedValue": "kryzza lugnt"
          }
  }
```

### DIFF
--- a/src/controls/editor/editform.js
+++ b/src/controls/editor/editform.js
@@ -20,7 +20,6 @@ const createForm = function createForm(obj) {
   const required = obj.required ? ' required' : '';
   const name = obj.name ? obj.name : '';
   let el;
-  let checked;
   let firstOption;
   const maxLengthText = maxLength ? `, max ${obj.maxLength} tecken` : '';
   switch (type) {
@@ -31,8 +30,9 @@ const createForm = function createForm(obj) {
       el = `<div class="validate ${cls}"><label>${label}<br><textarea name="textarea${maxLengthText}" id="${id}" rows="3" ${maxLength}${readonly}${required}>${val}</textarea></label></div>`;
       break;
     case 'checkbox':
-      checked = val ? ' checked' : '';
-      el = `<div class="o-form-checkbox ${cls}"><label for="${id}">${label}<input type="checkbox" id="${id}" value="${val}"${checked}${disabled}></label></div>`;
+      const isChecked = obj.config && obj.config.uncheckedValue ? obj.config.uncheckedValue !== val : val;
+      const checked = isChecked ? ' checked' : '';
+      el = `<div class="o-form-checkbox ${cls}"><label for="${id}"><input type="checkbox" id="${id}" value="${val}"${checked}${disabled}/>${label}</label></div>`;
       break;
     case 'dropdown':
       if (val) {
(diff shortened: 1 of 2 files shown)


johnnyblasta:
Tested and having some issues:

- When adding a new feature the checkboxes are checked, shouldn't they be unchecked.
- Leaving out the config part results in a error `Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'uncheckedValue')`. Would expect to not configuring specific values give the default 0 and 1 values.

steff-o:
> * When adding a new feature the checkboxes are checked, shouldn't they be unchecked.

They should follow the default setting. If no default is specified they should behave as before (which probably was unchecked)
> * Leaving out the config part results in a error `Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'uncheckedValue')`. Would expect to not configuring specific values give the default 0 and 1 values.

Probably missed that. It defaults to 0 and 1 if uncheckedValue and checkedValue are missing, but apparently it can't handle if the entire config object is missing.

I will soon return with an up-to-expectation-implementation.

steff-o:
@johnnyblasta I can not reproduce the crash. Can you provide a configuration example?

johnnyblasta:
> @johnnyblasta I can not reproduce the crash. Can you provide a configuration example?

I did like this:

```
{
  	"name": "typ",
  	"title": "Grill",
  	"type": "checkbox"
}
```
Checbox

steff-o:
@johnnyblasta, still not able to reproduce. Access to uncheckedValue is guarded by this thing of beauty:
```checked = (obj.config && obj.config.uncheckedValue ? obj.config.uncheckedValue !== val : val) ? ' checked' : '';```, which for me would make it impossible to crash as the lazy evaluation of the && operator would not evaluate ```obj.config.uncheckedValue``` Are you running your own javascript specification???

johnnyblasta (approved):
LGTM

Issue #1139, opened 2021-01-27
Support QGIS Server with the WFS layer filter


MattiasSp:
The `filter` layer option for WFS layers uses the hard coded URL parameter `&CQL_FILTER=` which is specific to GeoServer. The corresponding expression filter parameter in QGIS Server uses the parameter `&EXP_FILTER=`.

It would be great if expression filters could be used also with QGIS Server WFS services.

I'm not sure what the best implementation would look like but here are a few alternatives considered:

- Adding a new layer option in parallell to `filter`, e g `qgsFilter`
- Adding a new layer option to switch between filter types, e g `filterType` which could take `ecql` ([the GeoServer one](https://docs.geoserver.org/stable/en/user/tutorials/cql/cql_tutorial.html)), `qgs` ([QGIS expressions](https://docs.qgis.org/3.16/en/docs/user_manual/working_with_vector/expression.html)) or maybe even `ogc` ([OGC Filter Encoding](https://www.ogc.org/standards/filter), not as useful)
- Somehow checking which server type is used and choose the correct parameter name automatically. Not sure if that is possible though...

MattiasSp:
Any proposed solution should take care to use the appropriate filter type everywhere it is used and not only in the layer config. See also:

- [src/getfeature.js](https://github.com/origo-map/origo/blob/master/src/getfeature.js)
- [src/controls/offline/wfs.js](https://github.com/origo-map/origo/blob/master/src/controls/offline/wfs.js)

Perhaps a more general solution is possible?

steff-o:
#1407 Introduces an automatic detection of server type with a configuration fallback for WMS servers. Maybe a similar approach would keep it consistent when using vendor specific parameters.

Also #1390 is somewhat related as if a feature refresh on a source is implemented it would probably make getfeature.js obsolete by just exposing a getFeature method on the source. _getFeature.js_ is already quite buggy, not handling coordinate systems for WFS so good riddance to that. Maybe it must be kept for backwards compatibility, but in that case it can just call the source getFeature method.

Taking it one step further, it would be nice if the filter could be changed in runtime. Current implementation hides away the filter in a constant url base.  #1398 actually makes it possible to do so, but it won't handle BBox and qis filters gracefully.

MattiasSp:
> #1407 Introduces an automatic detection of server type with a configuration fallback for WMS servers. Maybe a similar approach would keep it consistent when using vendor specific parameters.

I like this idea!

[...]

PR #1535, opened 2022-05-03
Search config option 1 with custom layer projection

### LINKED ISSUE #1534: Allow differing layer projection when displaying search results from layers (option 1)
**Description**
When using "option 1" of the search control config - i e when using a map layer to display feature info on search results - the projection of the named layer is not taken into account.

In the [getFeature](https://github.com/origo-map/origo/blob/3c73d63ec0b7c200e789e7413909c1eba5a1743a/src/getfeature.js#L99) function a layer's default projection is used (no `srsName` parameter is specified) and it is never transformed, regardless of what projection has been specified in the layer configurations.

This becomes an issue when a layer has to use a different projection than the map projection, notably when using WFS layers from QGIS Server which only serves GeoJSON in EPSG:4326 (WGS84).

**Proposed solution**
If a layer's data projection differs from the map projection, transform the coordinates before returning the feature in the getFeature function.

**Alternative solutions considered**
...

**Additional context**
...

MattiasSp:
Fixes #1534.

Lets the [getFeature](https://github.com/origo-map/origo/blob/master/src/getfeature.js) function return features in the map's projection also when the layer projection differs.

Also a step towards #1139 as getFeature called for QGIS Server layers will always need to handle differing projections (due to QGIS Server only serving GeoJSON in EPSG:4326).

### DIFF
--- a/src/getfeature.js
+++ b/src/getfeature.js
@@ -92,7 +92,7 @@ sourceType.WFS = function wfsSourceType(id, layer, serverUrl, extent) {
   const data = ['service=WFS',
     '&version=1.0.0',
     `&request=GetFeature&typeName=${layer.get('name')}`,
-    '&outputFormat=json',
+    '&outputFormat=application/json',
     queryFilter
   ].join('');
 
--- a/src/getfeature.js
+++ b/src/getfeature.js
@@ -96,5 +96,14 @@ sourceType.WFS = function wfsSourceType(id, layer, serverUrl, extent) {
     queryFilter
   ].join('');
 
-  return fetch(url + data, { type: 'GET', dataType: 'json' }).then(res => res.json()).then(json => format.readFeatures(json)).catch(error => console.error(error));
+  return fetch(url + data, { type: 'GET', dataType: 'json' })
+    .then(res => res.json())
+    .then(json => {
+      const sourceOpts = layer.getSource().getOptions();
+      if (sourceOpts.dataProjection !== sourceOpts.projectionCode) {
+        return format.readFeatures(json, { dataProjection: sourceOpts.dataProjection, featureProjection: sourceOpts.projectionCode });
+      }
+      return format.readFeatures(json);
+    })
+    .catch(error => console.error(error));
 };


steff-o:
Sorry, I haven't tested it yet. I have no doubt it is working, but I think that this would probably be a great time to loose a couple of lines of code by using the WfsSource class instead of having double implementation. 

Most of the functionality in _getFeature()_ already exist in the _WfsSource_ class, so a minor change to that class could expose a _getFeature()_ method that does exactly the same thing. Basically the only thing that is needed is to move the line ```super.addFeatures(features);```  to the calling functions and create a _getFeature()_ method that calls ```_loaderHelper()``` and of course make sure all the filter stuff gets handled correctly.

The original _getFeature()_ function is only used by the search functionality, but as it is exposed through the api (At least the _MultiSelect plugin_ uses it to my knowledge), we can not just ditch it, but it can be rewritten as wrapper to WfsSource.getFeature() and taking care of keeping the function prototype intact. The source argument would be obsolete, unless someone actually calls it with another argument than the map configuration source array, which I hold for highly unlikely as it assumes that the features are fetched from a map layer, not any arbitrary WFS endpoint.

steff-o:
Tested but I'm confused. It solves just the exact problem when using a layer configured using a different projection than the map and querying using a FID like search does. But it does not work:

1. When the layer has another default projection than the map but it is not configured in origo. The _WfsSource_ class handles this by always request the features in the map's projection unless the layer has specified its own projection. If the _srsname_ parameter is omitted the server will respond in its native projection, which will be unknown to _getFeature()_ as origo does not use _getCapabilities_ to discover it.
2. When _getFeature()_ is called with an extent. The requested extent is not transformed and no SRS for the bounding box is present in the request. The server will then interpret the bounding box in the layer's native projection. Configuring a projection on the layer does not help here.

Maybe its just me that is lazy and never configure projections for every layer and lets the server handle that, but I think the result is inconsistent. At least case 1 should work. Case 2 is probably another issue as search never calls using extent, but would be nice to have fixed as well.

jokd:
@MattiasSp Any update on this issue?

steff-o:
@MattiasSp Is this PR still relevant now that _getFeature()_ has been completely rewritten? To my belief projections would work as expected in _WfsSource_ when using search.

Please review this PR and update it if possible or remove it.

MattiasSp:
> @MattiasSp Is this PR still relevant now that _getFeature()_ has been completely rewritten? To my belief projections would work as expected in _WfsSource_ when using search.
> 
> Please review this PR and update it if possible or remove it.

Thanks @steff-o! I agree that this can be closed now - great job om rewriting the `getFeature()` 👏

PR #1539, opened 2022-05-06
Fix scaling for drawn features in print-preview when changing DPI

### LINKED ISSUE #1538: Drawn features become smaller and thinner (incl. measure) during print-preview higher DPI
**Describe the bug**
With the current draw-plugin and Origo master drawn lines become thinner and points smaller if print preview then 300 DPI.
The same is true for the measure line.
**Expected behavior**
That they do not shrink / become thinner by a noticable amount.

**Additional context**
Relates to https://github.com/origo-map/draw-plugin.git / issue # 63
The drawn layers and measure act like for instance wfs layers did prior to last falls changes to the print control

sweco-semvtw:
This PR aims to fix #1538

Layers such as measure that do not have a style set directly will have their features styles be rescaled.


Grammostola:
I tested the draw pr + this pr and indeed things look much better in the print preview (measure and drawn features don't become much larger or smaller) however repeated entries and exits of the print preview seems to produce gradually growing drawn features, especially if their size property was adjusted in the print feature property editor.

johnnyblasta:
Please resolve the conflict, so it can be merged.

Issue #1537, opened 2022-05-03
rendermode: image / single tile layer prop for print


Grammostola:
**Is your feature request related to a problem? Please describe.**
If a tiled image layer like WMS or AGS_TILE is used in Origo it can look okay in the map and yield a pleasant experience when zooming and panning (sometimes in contrast to renderMode:image) however if containing labels it can get quite busy when printed because the layer is still tiled in the print preview. (This is in contrast to how Geoserver's mapfish operates by default)

![bild](https://user-images.githubusercontent.com/5138906/166477115-21bb1164-27e3-4c4c-8300-bb7e96044718.png)
(The orange area is most assuredly V Gotlandshavet)

**Describe the solution you'd like**
A layer property indicating whether the print control should preview the layer as a single tile ((AGS_MAP, renderMode: image),(renderMode: image)) 

**Describe alternatives you've considered**
None so far

**Additional context**
- We noticed it just recently on a WMS background map and road labels being repeated quite excessively
- renderMode: image makes layers blink out of and into existence which is a tad jarring but I think it preferrable to map prints with excessive labels

PR #1542, opened 2022-05-20
Separate printRenderMode for better labeling in printed maps

sweco-sedalh:
Introduces a `printRenderMode` layer setting, in parallell to the existing `renderMode`. Setting it to `image` will, during the print, replace the original (usually tiled) layer with an image layer. This has been tested with AGS_TILE/AGS_MAP layers.

Closes: #1537

**Note:** There is an outstanding bug in this that needs to be resolved before merge; after exiting the the print view the temporarily added layer does not get removed from the rendered map. It gets removed from the legend and from the `map.getLayers()` collection though. It is possible that there is some bug in OpenLayers, though I haven't been able to confirm the reason as of yet.


Grammostola:
I've had a look at this via updating the fork to the current master, resolving the component.js conflict in favor of the current master but kept `getEl: () => document.getElementById(id),` required by the print-component in this pr. 

Observed the same thing as noted above, even if the viewer (map) seemingly removes the added rendermode:image layer and thereafter does not seem to have a reference to it, it remains in the map and cannot be turned off (in the case of AGS_TILE layer). WMS layers seem to after the print preview come with the format_option dpi:150 (regardless of where the print preview was left). 

It would seem that if the 'added-for-print' layers are removed at the start of the `close()` method of the print-component, perhaps crucially before the `resetLayers()` and `setResolution()` methods are executed as opposed to last then this PR works a lot better, both for WMS layers and AGS_TILE layers, i.e a WMS layer is not left with a dpi param and there's no AGS_TILE layer ghost.

I am not entirely sure that everything is one hundred percent OK though since things (always) happen between the print preview and the map and labels may not be precisely where you left them once you return to the map.

I can attempt to push some changes? People may have input anyway?
Here's an animated gif showing an AGS_TILE layer with `rendermode: tile` and ``printRenderMode: image`

tile_to_print_singletile


(Yes, the autolegend functionality recently added appears missing. It seems to be the same in the current master, may have to do with the more recently reverted component.js changes, I shall create a separate issue for that)

sweco-sedalh:
Rebased and changed the code based on @Grammostola's findings, successfully tested with both AGS_TILE and WMS using the following configuration:

```json
    "avlopp_wms": {
      "url": "https://ext-geodata-nationella.lansstyrelsen.se/arcgis/services/HaV/lst_hav_sma_avlopp/MapServer/WMSServer"
    },
    "LST": {
      "url": "https://ext-geodata-applikationer.lansstyrelsen.se/arcgis/rest/services/VISS/lst_viss_karttjanst/MapServer"
    }
...
    {
      "name": "Vattenförvaltningens åtgärdsområden",
      "title": "AGS_TILE",
      "id": "26",
      "group": "root",
      "queryable": true,
      "type": "AGS_TILE",
      "renderMode": "tile",
      "printRenderMode": "image",
      "visible": false,
      "source": "LST"
    },
    {
      "name": "P12_Risk_paverkan_specifikt_vatten",
      "title": "P12_Risk_paverkan_specifikt_vatten_WMS",
      "source": "avlopp_wms",
      "group": "root",
      "hasThemeLegend": true,
      "type": "WMS",
      "renderMode": "tile",
      "printRenderMode": "image",
      "infoFormat": "application/geo+json",
      "abstract": "<font style='font-size:0.76rem;font-style: italic'> WMS från Länsstyrelsen.<br><a href='https://ext- geodatakatalog.lansstyrelsen.se/GeodataKatalogen/GetMetaDataById?id=10bae110-c35d-4743-92ea-d2a66178fedf' target='_blank'>Se Länsstyrelsens hemsida för mer infromation här.</a></font>",
      "visible": true
    },
```

There seems to be a bug in that the zoom level is wrongly restored after exiting the print preview, however that bug is also reproducible in current `master`.

Grammostola:
Have tested it with WMS layers from a Geoserver, WMS layers from an AGS and AGS MAP/TILE layers and it seems to work as intended.
There might be an issue with AGS WMS layers disappearing if too large images are requested (printrenderMode: image and 300DPI and a large paper size), that's in the nature of AGS WMS layers' default configurations as far as I'm aware.
 For our purposes this change is primarily for the background maps with labels (and for this it works without a hitch so far, we are currently testing it internally).

It changes a few files. component.js gets a new method, "getEl()" which is used in a few components, map.js' function expression Map becomes a class (which is adapted for in (the) one place where map.js is used),  viewer.js gets a removeLayer method and it introduces an event called "remove:layer" which rhymes with "remove:group" but not "addlayer" so that event has been renamed "add:layer" which is a potentially breaking change for potential listeners to the "addlayer" event. I think this change should ultimately be made to make the events consistent however and the question becomes relevant with this PR as another event is added.

It slightly changes the signature of addLayer() but not in a way that should break current dependencies as far as I can see.

There are some other formatting / documention bits that seem slightly beneficial if anything, at least visually, though not critical for this particular PR perhaps.

It's an "opt-in" bit of functionality but since it touches many files I would appreciate if other maintainers could have a look :)

Grammostola:
The PR is now minimal. It has worked well in our fork for our purpose: not endlessly repeated labels in printed-at-300dpi WMS background maps.

tonnyandersson:
LGTM. Still getting some duplicated labels though, but I think you need to tinker with the service settings to make it perfect.

Withholding an approval in case someone else wants to chime in.

jokd (approved):

PR #1550, opened 2022-06-03
Labels in print bug fix

### LINKED ISSUE #1546: Printing layers with combined styles that includes text
**Describe the bug**
Problem when entering print-mode when a layer that has booth a text style and another style is visible.
For example a point layer styled with both a circle and a text. 
When the print-mode is entered the map freezes. 

**To Reproduce**
Style-example that doesent work. 

```
    "origo-logo": [
        [
            {
                "label": "Origokommuner",
                "circle": {
                    "radius": 10,
                    "stroke": {
                        "color": "rgba(0,0,0,1)",
                        "width": 2.5
                    },
                    "fill": {
                        "color": "rgba(255,255,255,0.9)"
                    }
                }
            },
            {
                "circle": {
                    "radius": 2.5,
                    "stroke": {
                        "color": "rgba(0,0,0,0)",
                        "width": 1
                    },
                    "fill": {
                        "color": "rgba(37,129,196,1)"
                    }
                }
            },
            {
                "text": {
                    "font": "Bold 12px Arial",
                    "textAlign": "center",
                    "offsetX": 10,
                    "offsetY": -10,
                    "text": "My label text",
                    "fill": {
                        "color": "rgba(0,0,0,1.0)"
                    },
                    "stroke": {
                        "color": "rgba(255,255,255,0.7)",
                        "width": 2
                    }
                }
            }
        ]
    ]
```

**Screenshots**
image

 - Windows 10
 - Chrome

jokd:
Fixes #1546 in a quick and dirty way

### DIFF
--- a/src/controls/print/print-resize.js
+++ b/src/controls/print/print-resize.js
@@ -385,29 +385,10 @@ export default function PrintResize(options = {}) {
         }
         const newStyle = Style.createStyle({
           style: styleName,
-          viewer
-        })();
+          viewer,
+          scaleToDpi: resolution
+        });
         if (newStyle) {
-          const styleScale = multiplyByFactor(1.5);
-          newStyle.forEach(style => {
-            const image = style.getImage();
-            if (image) {
-              const imageScale = image.getScale() ? multiplyByFactor(image.getScale()) : styleScale;
-              image.setScale(imageScale);
-            }
-
-            const stroke = style.getStroke();
-            if (stroke) {
-              const strokeWidth = stroke.getWidth() ? multiplyByFactor(stroke.getWidth()) : styleScale;
-              stroke.setWidth(strokeWidth);
-            }
-
-            const text = style.getText();
-            if (text) {
-              const textScale = text.getScale() ? multiplyByFactor(text.getScale()) : styleScale;
-              text.setScale(textScale);
-            }
-          });
           layer.setStyle(newStyle);
         }
       } else if (features) {
--- a/src/controls/print/print-resize.js
+++ b/src/controls/print/print-resize.js
@@ -374,6 +374,15 @@ export default function PrintResize(options = {}) {
       if (styles && styles.length > 1) {
         changeWfsThemeLayer(layer, styleName, styles);
       } else if (styleName && styles && styles.length === 1) {
+        const layerName = layer.get('name');
+        const newStyles = [...styles];
+        if (!(layersOriginalStyles.some(layerInArr => layerInArr.layerName === layerName))) {
+          layersOriginalStyles.push({
+            layerName,
+            styleName,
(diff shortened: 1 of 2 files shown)


Owemyr:
Got this when trying style with text.
`   "stylename": [
      [
        {
          "text": {
            "text": "{{distrnamn}}"
          }
        }
      ]
    ]`

text

No console error but interestingly, when I duplicated text style in two blocks (thematic styling) it worked in print controller.
`  "stylename": [
      [
        {
          "text": {
            "text": "{{distrnamn}}"
          }
        }
      ],
      [
        {
          "text": {
            "text": "{{distrnamn}}"
          }
        }
      ]
    ]`

Owemyr:
Perhaps a separate issue but we got this error in print preview on styles with icon rotation based on attribute.
rotation_svg

Example style
`   "style_name": [
      [
        {
          "icon": {
            "rotation": "{{rotation}}",
            "scale": 1.2,
            "src": "img/svg/pilsvart.svg"
          }
        }
      ]
    ]`

jokd:
The print resizer replaced the style function with a static style, making use of attributes harder. Now I have pushed a change where this hopefully is fixed and that should work both for labels and icon rotation.

Owemyr:
Nice, both labels and rotation styles now work in print-controller.

tonnyandersson (approved):

jokd:
Thanks for the review. It is related so I will fix that as well.

tonnyandersson:
Seems to fix the issues at hand, but I'll let @Grammostola have the final word on this.

However, @jokd, I found another bug - cluster styles get dropped indefinitely when going into print mode.

cluster_print


I would say it's a different issue, but it is related to style and print so maybe it's an easy fix while the hood is popped and the tools are out. Your call.

Grammostola:
Nice work @jokd , makes a certain kind of sense to move some responsibility for rescaling into style.js I guess. Labels don't crash the app anymore which is the main thing, they're merely not rescaled (so become smaller when 300DPI).

As for cluster styles it seems they with this pr go from resizing (and reforming) when changing DPIs in the print pane to not doing that but instead losing their text. As @tonnyandersson mentions they are still missing back in the map though the clustering functionality seems to persist.

I don't see any other side effect introduced, scaled icons and the measure line are still ok in the print pane when changing dpis and after, WMS icons are properly fetched at higher DPIs etc.

(The print control is hard and I won't pretend to understand precisely all of it but I think I can see a need for a smorgasbord map config and a set of use cases for testing the thing. I don't think it can be automated to any significant degree? (not as in selenium or the like)

Grammostola:
Looks good : )

tonnyandersson:
LGTM.

Issue #1551, opened 2022-06-03
Legend groups and subgroups broken


Owemyr:
**Describe the bug**
Legend groups and subgroups broken, see behavior from gif-video below.

**To Reproduce**
Add groups and subgroups to json-config and try to expand and minimize or turn on all layers in a subgroup.

**Screenshots**
![groups_probs2](https://user-images.githubusercontent.com/47249045/171834193-4724746b-b830-4164-aa4e-8c18ca2aa237.gif)

**Desktop (please complete the following information):**
 - OS: Windows10
 - Browser Edge ( version 102.0.1245.30)

jokd:
Thanks for the report. 
Seems to be multiple listeners on the legend items, for some reason.

jokd:
I can backtrack this to changes to 'src/ui/component.js' in #1529 where the component is refactored. This somehow messes up the legend. 
@sweco-sedalh @Grammostola any input on this?

Grammostola:
We shall strive to get back to you sometime next week.

sweco-sedalh:
A fix is now available in #1552.

@Owemyr can you test it to make sure that it fixes the bug for you?

Owemyr:
Tested #1552 with the reverted refactoring, working fine in our fork and configs.
So far we don't see any strange behavior.

PR #1552, opened 2022-06-07
Fix group and subgroup expansion

sweco-sedalh:
This PR reverts the component refactor from #1529, fixing group and subgroup expansion.

~The refactor of the base component in #1529 slightly changed how events where attached, which made it possible for onRender to be called more than once which in turn attached event handlers more than once, resulting in bugs like in #1551.~

~This PR fixes that by somewhat formalizing (and reducing) the handling of especially the "render" event. There is also commented out code to get a warning on similar issues (currently present in the print component, though without any user visible issues that I could find).~

Closes: #1551

### DIFF
--- a/src/controls/print/print-component.js
+++ b/src/controls/print/print-component.js
@@ -18,7 +18,7 @@ import { downloadPNG, downloadPDF, printToScalePDF } from '../../utils/download'
 import { afterRender, beforeRender } from './download-callback';
 import maputils from '../../maputils';
 import PrintResize from './print-resize';
-import { withLoading } from "../../loading";
+import { withLoading } from '../../loading';
 /** Backup of original OL function */
 const original = PluggableMap.prototype.getEventPixel;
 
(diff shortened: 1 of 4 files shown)


tonnyandersson:
This PR may potentially have a negative impact on _evertything_ that uses ui components (controls, plugins, whatevers). We have an upcoming release and we do not find the idea of potentially introducing a myriad of bugs in that release particularly appealing.

The changes made to the base component in #1529 will probably be undone very soon, which will make this PR obsolete. The changes were not necessarily bad, but they should have been made in a separate PR and should not have been merged this close to a release because of the broad impact such fundamental changes might have.

sweco-sedalh:
To bad that no one highlighted this during the 5(!) weeks that the original PR was up for review... I even put the refactor into a separate commit there to make it easy to remove from that PR in case there would have been objections.

Changed this PR to revert the previous commit and created a new PR (#1553) for the component refactor. I hope that one will not be laying around another 5 weeks so that the time can instead be used to fix this "myriad" of bugs your speaking off.

steff-o:
@sweco-sedalh Could you also revert the changes that introduced the lint errors?

steff-o:
@sweco-sedalh Are you running linux? The lint script works on my machine (windows). I recently changed it in #1524 as previously it did not recurse the _src_ directory and _origo.js_ was omitted. I changed it to something that I tested to work on windows. Your approach also seems to work.

However, I strongly oppose to adding the lint Github workflow in this PR. I only asked to get the lint errors that were introduced fixed somehow (reverted or rewritten or whatever). The workflow is a great addition, but according to our developing guideline it should be a separate issue and PR.

sweco-sedalh:
@sweco-sedalh Indeed (WSL to be correct). Recursive glob (`**`) is not enabled by default in bash and thus won't work for most.


I have split out the workflow and lint issues not introduced by #1529 into #1561.

jokd (approved):

tonnyandersson:
I guess this one is good to go. @jokd, merge it if your approval still stands.

PR #1553, opened 2022-06-09
Refactor the core Component

sweco-sedalh:
Refactor the core `Component` into a class and tries to clean up it's usage a bit. Intended to be rebase on top of #1552 once that's been merged to reintroduce the refactoring after the next release.


sweco-sedalh:
Some rationale for this refactor, ordered roughly by the steps taken:

1. **Using class and constructor syntax:** Mostly it's about making things "be what they are". In the previous code the `Component` (or rather its inner function, `createComponent`) function fulfilled the role of a class constructor without actually being one while the `Base` function corresponded to to a class. This refactor makes it as clear as possible that this is the case, by actually implementing them as a class and its constructor.
    * However, because that would have required very wide reaching code changes (changing _all_ "subclasses" of `Component` into actual classes as well) this refactor kept the `Base`-class/`Component`-"constructor". The next step on this path would be to also export `Base` and begin changing the rest of the code base, however I believe that should wait until a decision on using React or not has been made (if so, this would not be needed). While there are "subclasses" not being classes we must keep the `Component` function so that it can call `dispatch("init")`, which later would be part of the subclasses' constructors instead.
    * Doing so opens up a lot of niceties, like getting automatically getting the component name (= class name) in DevTools, being able to use `instanceof`, better auto-completion/IDE-support/IntelliSense, etc.
    * Also, a scaringly huge part of the Origo code base currently violates the naming conventions in the AirBnB style guide as well as the generally accepted convention in JavaScript code of using [`PascalCase` for classes](https://github.com/airbnb/javascript#naming--PascalCase) and [`camelCase` for functions](https://github.com/airbnb/javascript#naming--camelCase), this refactor fixes that at least in one case.
3. **Changing the body of the constructor/`createComponent`:** Previously, while not being a lot of lines of code, it was very specific which event handlers (`on*`) properties where automatically connected as listeners and under which circumstances, with this refactor this gets simplified by connecting _all_ properties with the right name thus making it clearer how `Component` is to be used.
4. **Formalizing the `onRender`/`onAdd` usage:** Further on the path of clearing up how this central part of Origo can be used, there are really only two use cases; either the component is "top level" (for lack of a better description), meaning it renders itself directly somewhere into the DOM and the `render` function returns nothing, or it is not in which case it's rendered into the DOM by one of its ancestor components by returning the HTML from its `render` function. Since there are no other use cases (neither in theory nor in practice in the current code base that I've found) we can safely reduce the amount of manual code that needs to be written for each component, thus reducing the amount of potential mistakes.

Finally, there is one part which isn't as much a refactor but rather a new feature:

5. **Introducing the `html` tagged template literal:** This simplifies the very common use case of "rendering some HTML with the ID of the component and some child components embedded". See [here](https://github.com/origo-map/origo/commit/7e99d2dbd97b6434a152bcedaf8e68ebc394546c#diff-3423bd883103d0c08fa031b3cb9d1813eb8b07901cbca74ac847f8a90ce67f0eR29-R36) for an example.

sweco-sedalh:
No. Something like that _could_ be added (essentially through a large dictionary of components), however I'm quite skeptical to that idea; that's the sort of pattern that leads to spaghetti code and that given a conversion to React or similar will need to be eliminated.

jokd:
OK so then adding useful methods to the specific component makes more sense. Thanks.

Grammostola:
Closing due to lack of interest/activity

PR #1555, opened 2022-06-13
Handle layergroups

### LINKED ISSUE #1554: GetFeatureInfo for WMS layers inside grouplayers
**Describe the bug**
GetFeatureInfo click not working when a WMS layer is inside a grouplayer. 
No warnings or errors in console, the click result in nothing/pin.

**To Reproduce**
Config a WMS-layer with attributes and set it to `"queryable": true` inside a grouplayer `"type": "GROUP"`

 - OS: Windows 10
 - Browser: Edge
 - Version: 102.0.1245.33

**Additional context**
If we revert PR #1513 it works again.

jokd:
Fixes #1554
Layergroups needed to be split up and each layer in the group be handled as individual layers. This will make the title of the popup use the layer title rather than the layergroup title.

### DIFF
--- a/src/featureinfo.js
+++ b/src/featureinfo.js
@@ -1,6 +1,7 @@
 import Overlay from 'ol/Overlay';
 import BaseTileLayer from 'ol/layer/BaseTile';
 import ImageLayer from 'ol/layer/Image';
+import LayerGroup from 'ol/layer/Group';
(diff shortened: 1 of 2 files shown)


Owemyr:
Fix appreciated! Thanks @jokd
Works well, no sideeffects so far.

PR #1557, opened 2022-06-16
Extended legend inline

### LINKED ISSUE #1528: Legend - Display thematic symbology inline
If a layer has a thematic symbology, only an icon is displayed in the legend instead of the actual symbology:
image

To see the actual symbols the user has to click the three-dot menu (or layer name in older versions) to open the layer info, which only displays that layer and hides the rest of the legend.

By clicking the 
image
the symbology should be displayed in the legend in the same way a group is expanded:

image

The actual visual representation does not have to be exactly like in the picture, I just photo shopped some screen shots. Maybe the icon should be replaced with a chevron like groups as well.

As always there has to be a configuration that says if the layer starts expanded...

Put my head in the sand.

steff-o:
Closes #1528 

Extended legends (e.g composite WMS-layers and thematic vector layers) are displayed below the layer name in legend. Default they are collapsed and are displayed by clicking the extended/no symbol icon.

Layers that do not have an extended legend are affected in that way that clicking the symbol now does exactly nothing instead of toggling visibility.

Layers that does not have any symbol at all will display a _legend missing_ text when the list symbol is clicked, which is the same as in layer info.

To start with the extended legend visible add `'expanded ': true` to the layer configuration.

It supports changing symbology using the stylePicker.

__Known issues__
If a thematic symbology has a `'header': true` setting, the thematic symbology will not be available inline. The main reason is that it would not be clear to the user that the symbol in the legend is clickable to see the rest of the symbols. To make it clear, the extended legend would need its own drop down button or be inside the three-dot-menu, which defies the simplicity of displaying them inline.

It may be hard to test with legend groups, as the eventhandling for collapsing and expanding groups is broken until #1551 is resolved.

### DIFF
--- a/scss/ui/_flex.scss
+++ b/scss/ui/_flex.scss
@@ -6,87 +6,96 @@ $flex-grow: 1;
 $flex-grow-all: 2;
 
 .flex {
-  // flex container
-  display: flex;
+    // flex container
+    display: flex;
 
-  &.row {
-    flex-direction: row;
-  }
+    &.row {
+        flex-direction: row;
+    }
 
-  &.row-reverse {
-    flex-direction: row-reverse;
-  }
+    &.row-reverse {
+        flex-direction: row-reverse;
+    }
 
-  &.column {
-    flex-direction: column;
-  }   
+    &.column {
+        flex-direction: column;
+    }
 
-  &.column-reverse {
-    flex-direction: column-reverse;
-  }
-  
-  // flex items
-  .shrink {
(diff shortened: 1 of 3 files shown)


jokd:
Looks great @steff-o 
Nice job!

jokd (approved):

David0099:
Nice function!
Looks great!

Tested and have one thought, noticed that the toggleButton to turn layer on/off and the moreInfoButton jumps down to a new row if the title of the layer is a bit longer. 
before
after

PR #1562, opened 2022-06-17
Adds export option for layers

### LINKED ISSUE #1559: Export/import of layer
Possibility to export layer in the legend (primarily draw layers), which can be imported (drag and drop) with retained styling into another origo instance.

jokd:
Fixes #1559 

Should be used with caution and works only for vector layers. It exports the features that has been requested from the server, thus works best with geojson-layers or wfs with strategy:all but is primarily intended for draw layers.

Configured like this:
   
>  {
      "name": "origo-cities",
      "title": "Origokommuner",
      "group": "root",
      "style": "origo-logo",
      "type": "GEOJSON",
      "attributes": [
        {
          "name": "name"
        }
      ],
      "visible": true,
      "exportable": true,
      "exportFormat": "geojson" // ["geojson", "kml"]
    }

exportFormat can be a string (geojson, gpx or kml) or an array of formats. Defaults to geojson in case om omitted or invalid.

### DIFF
--- a/src/controls/legend/overlay.js
+++ b/src/controls/legend/overlay.js
@@ -1,6 +1,7 @@
 import { Component, Button, dom } from '../../ui';
 import { HeaderIcon } from '../../utils/legendmaker';
 import PopupMenu from '../../ui/popupmenu';
+import exportToFile from '../../utils/exporttofile';
 
 const OverlayLayer = function OverlayLayer(options) {
   const {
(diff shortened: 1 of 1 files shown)


johnnyblasta:
It would be nice it exportFormat was case insensitive.

jokd:
Right on Sloppy J

johnnyblasta:
> Right on Sloppy J

I was mainly thinking of the formats them self to be case insensitive, so you could write geojson or GEOJSON, kml or KML.

jokd:
Have a go again. Now its very forgiving. 
Could have done this transform in the export util instead but I put it in the same function as before.

PR #1566, opened 2022-06-30
Fix blurry editor form

### LINKED ISSUE #1563: Blurry text in editor form
**Describe the bug**
When editor form gets longer so scroll kicks in the text get blurry, a minor problem but still annoying.

**To Reproduce**
1. For example copy some more textarea attributes in examples/editor.json 
2. Scroll down or click at an input field and see the difference.

If you don't see the problem, try changing browser window height when looking at the editor form.

**Expected behavior**
Sharp text that not change appearance to blurry.

**Screenshots**
editor_blurry_text

**Desktop:**
 - OS: Windows 10
 - Browser: Edge (102.0.1245.44) and Chrome (102.0.5005.115)

**Additional context**
Not tested on mobile/tablets.

tonnyandersson:
Fixes #1563, quick and dirty.

Can't see that `transform-style: preserve-3d` has been used for anything in particular in the past and I haven't noticed any negative side effect after removing it. But please, throw your weirdest modal configs at it and see if it breaks!

### DIFF
--- a/scss/_modal.scss
+++ b/scss/_modal.scss
@@ -18,18 +18,12 @@
   border-radius: 6px;
   bottom: auto;
   box-shadow: 0 2px 8px 1px rgba(0, 0, 0, 0.45);
-  box-shadow: 0 2px 8px 1px rgba(0, 0, 0, 0.45);
   font-family: Arial,sans-serif;
   left: 50%;
   position: absolute;
   right: auto;
   top: 50%;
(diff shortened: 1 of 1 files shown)


Owemyr:
Thanks, tried with a few long editor forms and some other modals.
No problems so far.

johnnyblasta (approved):
LGTM
Thanks, fixed a blurry splash control.

PR #1569, opened 2022-07-11
feature: measure elevation profile

### LINKED ISSUE #1567: See elevation profile on measures
When using measure control it would be nice to get a elevation profile of the measured object. The measured geometry is sent to a service which sets the height on every coordinates like this PR origo-map/origo-server#120

ElevationProfile

johnnyblasta:
Closes #1567 

Add this config option to measure control to point to a Origoserver that adds height to the geometry:
```
"options": {
        "elevationProfileURL": "http://localhost:3001/origoserver/lm/elevation"
}
```
Use this PR origo-map/origo-server#120 for setting height.

ElevationProfileNew


tonnyandersson:
I think there's [a plugin for this](https://github.com/origo-map/elevation-profile-plugin).

johnnyblasta:
> I think there's [a plugin for this](https://github.com/origo-map/elevation-profile-plugin). Well, not quite, but kinda.

Yes, I'm aware of the plugin, but there has to be a configured layer with heights on that one and since this uses the Lantmäteriet Höjd Direkt service to add the heights we thought it were appropiate to have it in measure control.

tonnyandersson:
I'm sure the elevation profile plugin can be modified to make use of the LM service. There's even an [open issue on this very subject](https://github.com/origo-map/elevation-profile-plugin/issues/2).

The thing is, this feature uses the ol-ext library, which is why the elevation profile plugin was created as a plugin and not a built-in feature. Same thing with the multiselect plugin, which uses the turf.js library.

johnnyblasta:
> I'm sure the elevation profile plugin can be modified to make use of the LM service. There's even an [open issue on this very subject](https://github.com/origo-map/elevation-profile-plugin/issues/2).

Then it has to be hooked on too some draw function also to get an arbitrarily drawn objects elevation profile, which I guess could be done somehow.

> 
> The thing is, this feature uses the ol-ext library, which is why the elevation profile plugin was created as a plugin and not a built-in feature. Same thing with the multiselect plugin, which uses the turf.js library.

Hm, is there a way to build it without including the external library and those that want to use the function includes it from the html file?

tonnyandersson:
> Hm, is there a way to build it without including the external library and those that want to use the function includes it from the html file?

Well, I guess there is, technically. But that's bad design. Like, _really_ bad.

I guess this raises a bunch of philosophical questions regarding plugins and non-essential external libraries that needs to be discussed in a larger forum. To me, it's a no-brainer. But this is not a dictatorship. Sadly. :)

johnnyblasta:
Moved showing of elevation profile to plugin to lose dependency on ol-ext in Origo.
New version of plugin is on the way.

johnnyblasta:
Closing this since Karlstad has been working on the plugin which solves this.

PR #1571, opened 2022-07-18
feature: turn on all layers button

### LINKED ISSUE #1511: Turn on all layers button
There is a turn off all layers button in legend, should it also be a turn on all layers button?

A button in legend control which turns on all layers. Default not visible.

If there are many layers turning them all on may be a heavy action and possible not very useful.

johnnyblasta:
Closes #1511 

Config:
```
{
  "name": "legend",
  "options": {
    "turnOnLayersControl": true
  }
}
```


tonnyandersson:
Why remove the `layerControl` wrapper element? Just asking.

Also, `addDiveder` should be spelled `addDivider`.

Also, the whole `addDivider` flag thing wouldn't be necessary if you had kept the `layerControl` wrapper.

Also, things become a bit awkward if you add more stuff to the legend toolbar:
image

Seems to work fine, though!

tonnyandersson:
Better!

But I kinda liked having the layer search control outside the `layerControl` wrapper element, with the divider between them and all (like in the screenshots above). Not quite sure what's more logical, though.

johnnyblasta:
> Better!
> 
> But I kinda liked having the layer search control outside the `layerControl` wrapper element, with the divider between them and all (like in the screenshots above). Not quite sure what's more logical, though.

I've no strong opinion of if it should be within or without, so we can wait to here what other thinks.

johnnyblasta:
Which is preferred?
Search outside of layer control component.
image

Search inside of layer control component. (This one is missing some space to the right of the search)
image

jokd:
I like the version with the divider to indicate that the toggle buttons and search field are not related.

Grammostola:
I like the divider between the search field and the buttons too.

A minor note is that this turns on all layers in a group with the exclusive property. The alternative I see would be to not do anything with layers in such groups but presently several or all such layers can be visible when the map loads and if that is the case and the user proceeds to click/tap on of those layers to activate the exclusive functionality and there's only one visible layer and then hits this button said user would likely expect all layers in this group to become visible.

Though I can't really say what the user of a map with a group with the exclusive property that loads with only one of those layers visible would expect when clicking tapping this button. "Not all background layers" is more obvious.

jokd:
> I like the divider between the search field and the buttons too.
> 
> A minor note is that this turns on all layers in a group with the exclusive property. The alternative I see would be to not do anything with layers in such groups but presently several or all such layers can be visible when the map loads and if that is the case and the user proceeds to click/tap on of those layers to activate the exclusive functionality and there's only one visible layer and then hits this button said user would likely expect all layers in this group to become visible.
> 
> Though I can't really say what the user of a map with a group with the exclusive property that loads with only one of those layers visible would expect when clicking tapping this button. "Not all background layers" is more obvious.

I think exclusive groups should be excluded in this functionality

johnnyblasta:
Put in the divider after search input and made the turn on all skip exclusive groups.

tonnyandersson:
Seems to work fine. Still getting this empty void to the right of the layers list, though. Probably because the layers list has got a fixed width set to `220px` in overlays.js. Setting the width to `100%` _seems_ to do the trick, but I'm not `100%` sure it won't break anything else.

johnnyblasta:
> Seems to work fine. Still getting this empty void to the right of the layers list, though. Probably because the layers list has got a fixed width set to `220px` in overlays.js. Setting the width to `100%` _seems_ to do the trick, but I'm not `100%` sure it won't break anything else.
> 
> image

The problem with the width of the layers list seems to be a pre existing problem not caused by this PR, but perhaps amplified by adding more width to the control components. Setting the width to 100% seems to cause the legend to take up a large portion of the screen.  I usually set contentStyle to a specific width for the legend to avoid empty space after layers list, but that is something that should be unnessary, since it should work right away.

tonnyandersson:
Setting the layers list width to `100%` doesn't seem to make the legend take up any larger portion of the screen than before. All it does is extending the layers list all the way to the right. Or am I missing something?

image

jokd:
@johnnyblasta Can you fix the css issues so we can merge this?

johnnyblasta:
> 

One thing this changes is that there will be no row breaks on long layer titles. So if there is long titles the legend can take up a large portion of the screen, this can be avoided by setting the width of the legend see examples below.

With 100%
image

With 100% and `"contentStyle": { "width": "310px" }`
image

With 220px
image

I think using 100% is the better option than having  a set width of 220px, so will add this if no one objects.

jokd:
LGTM

jokd (approved):

PR #1576, opened 2022-08-03
Fix for popup menu

### LINKED ISSUE #1558: Scrolling the legend with popup menu expanded
It would be sweet if the popup menu followed scrolling of the legend. 

chrome-capture-2022-5-17

And if you open the menu, close it, scroll a bit in the legend and then open it again it will be in the same spot it was before, quite confusing.

jokd:
Fixes #1558 
Cleaned up a bit, hopefully didn't mess up anything

### DIFF
--- a/scss/ui/_flex.scss
+++ b/scss/ui/_flex.scss
@@ -50,6 +50,10 @@ $flex-grow-all: 2;
     flex-wrap: nowrap;
   }
 
+  .wrap {
+    flex-wrap: wrap;
+  }
+
   .justify-start {
     justify-content: flex-start;
   }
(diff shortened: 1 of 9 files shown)


Grammostola:
Hi, it seems to work as intended now, i,e sticks with the legend when the legend is scrolled and doesn't re-open in the same place as last place but rather always somewhat above the three dots, very much an improvement. Would it be possible to slightly reduce the distance above the three dots that it opens at?

jokd:
Tested now and I can see that the first layer has less margin between the three dots and the menu. Is that distance better?

jokd:
This is ok to test now. Tried different approaches to slide down the menu as well as to position the menu on top of the legend but stuff got messed up no matter what so I think this for now is the least lousy solution without rewriting the legend quite a bit.

tonnyandersson:
LGTM.

Grammostola:
Looks good to me

PR #1578, opened 2022-08-05
feat: enable automatic style creation for wms layers' getLegendGraphic

### LINKED ISSUE #1560: Use style from WMS server
When you have a WMS layer and the style in the map comes from the WMS server and the Origo style in the config is basically a URL to the GetLegendGraphic call from the server. It would simplify things if it was possible in the layer config to say just to get the legend graphics from the server. Then there is no need to setup a style for the layer in the config, which reduces administration of styles and less places to make edits if for say the adress to the server is changed f.e. switching between test and production servers.

On the layer config there should be a option to say that the style should be fetched from the server.

I guess the option should include more settings such as the possibility to use an alternative style other than the default. Also extendedLegend should be able to set. 
In print legend there is a similar function with GetLegendGraphic call one could look at.

Grammostola:
Proposes to fix #1560 . Introduces two properties of a WMS layer:
```json
      "hasThemeLegend": true,
      "legendParams" : {
        "scale" : 5000,
        "legend_options" : "dpi:300"
      }
```
Where `hasThemeLegend` is necessary for "auto" style creation for WMS layers with a theme legend (to extendedLegend) and
`legendParams` is more or less what @steff-o suggested in the related issue. 

Both properties are ignored if a `style` property also exists for the given WMS layer.

The default "scale" for the getLegendGraphic call (there needs to be a scale for scale dependant styles) is the largest scale of the configured map (because typically theme layers have more visible classes at a larger scale than at a smaller scale if scale dependant) however that value is overriden if `scale` is given as a `legendParam`.

The possibility exists to make `hasThemeLegend` part of the `legendParams` object but it's slightly easier to merely define the `hasThemeLegend` property rather than it as part of an object if no special legendParams as per getLegendGraphic() are to be employed.

### DIFF
--- a/src/layer/wms.js
+++ b/src/layer/wms.js
@@ -37,6 +37,20 @@ function createImageSource(options) {
   }));
 }
 
+function createDefaultStyle(wmsOptions, source, viewer) {
+  const maxResolution = viewer.getResolutions()[viewer.getResolutions().length - 1];
+  const styleName = `${wmsOptions.name}_WMSDefault`;
+  const getLegendString = source.getLegendUrl(maxResolution, wmsOptions.legendParams);
+  const style = [[{
+    icon: {
+      src: `${getLegendString}`
+    },
+    extendedLegend: wmsOptions.hasThemeLegend || false
+  }]];
+  viewer.addStyle(styleName, style);
+  return styleName;
+}
+
 const wms = function wms(layerOptions, viewer) {
   const wmsDefault = {
     featureinfoLayer: null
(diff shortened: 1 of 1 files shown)


jokd:
Tested and LGTM

Grammostola:
Currently this pr sets the styleName property of a layer while its style property remains 'default'. In comparison the same layer with a style given in index.json will have both the style and styleName property as the name of that style. (layer.js does 
`layerOptions.styleName = layerOptions.style;`)
At a glance I don't see any `layer.get('style')` references in the origo client, merely `layer.get('styleName')` so I don't know whether this makes a difference however I can see the logic in making these auto-default-WMS-layers identical to entering a style manually for them. Any thoughts on this?

tonnyandersson:
@Grammostola Quagmire.

jokd:
I guess we can live with that

Grammostola:
_Answers clear as glass
Understanding rolling in
So shall it be now_

:otter:

PR #1579, opened 2022-08-08
Make vector sources aware of custom layer extent

### LINKED ISSUE #1523: Configured vector layer extent not fully respected
**Describe the bug**
Even though it seems to be working well at a glance, setting an extent (a _bbox_, essentially) on a vector layer actually fails in spectacular silence. _All_ features in the layer are loaded, but only the ones inside the extent are shown on the map. The ones outside the configured extent are still there, though. Invisible. Lurking. Scheming.

This bug affects at least WFS and GEOJSON type layers, possibly others.

**To Reproduce**
1. Configure a WFS layer with an extent that excludes some features
2. Load map and click _outside_ the configured extent, in a spot where you know an "excluded" feature is situated
3. Hear the audience gasp in disbelief as the previously invisible feature is revealed

extentfail

**Expected behavior**
One would expect that setting an extent on a vector layer would actually exclude all features outside the configured extent, not just hide them.

**Additional context**
Probably not related to #1025.

tonnyandersson:
Fixes #1523.

Had to copy and reset `extent` parameter to ensure backwards compatibility and to make sure it doesn't affect layer rendering since `extent` is also a native vector layer parameter.

### DIFF
--- a/src/layer/geojson.js
+++ b/src/layer/geojson.js
@@ -12,6 +12,9 @@ function createSource(options) {
         const numFeatures = vectorSource.getFeatures().length;
         for (let i = 0; i < numFeatures; i += 1) {
           vectorSource.forEachFeature((feature) => {
+            if (!feature.getGeometry().intersectsExtent(options.customExtent)) {
+              vectorSource.removeFeature(feature);
+            }
             if (!feature.getId()) {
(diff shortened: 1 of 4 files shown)


jokd (approved):
Looks solid

PR #1581, opened 2022-08-11
fix: make sure all AGS layer ids compared in print-legend are numbers

### LINKED ISSUE #1580: auto print legend for AGS_MAP/TILE layers depending on id doesn't work
If an AGS_MAP or AGS_TILE type layer is configured in index.json without a style but with the correct id and some name the layer will not automatically receive a print legend (there will simply be the title of the layer and no legend icon/s)

The culprit appears to be the following line in the "getAGSJSONContent()" of print-legend.js:
```javascript
const legendLayer = json.layers.find((l) => l.layerId === id || l.layerName === id)
```
where the id of the received json can be a Number and the id argument is a string.

Grammostola:
Aims to fix #1580

### DIFF
--- a/src/controls/print/print-legend.js
+++ b/src/controls/print/print-legend.js
@@ -215,7 +215,7 @@ const LayerRow = function LayerRow(options) {
     if (!json) {
       return getTitleWithIcon(title, '');
     }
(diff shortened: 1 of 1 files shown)


jokd:
Unable to test this but I suppose you have tested this thoroughly and can merge it unless @tonnyandersson oppose it

PR #1597, opened 2022-09-19
fix: add method for plugins to regenerate visible layers

### LINKED ISSUE #1585: Regenerate show-visible-layers at the very least-method
Our Layermanager plugin onAdd() checks if there's a mapState with Layermanager-added-layers defined, if so it does
`viewer.addLayer(layer)`. Since this happens after 
`origo.on('load', function(viewer) {..` 
the visible layers control has already established if active and what layers should be visible by the time Layermanager adds layers, which makes a mapState featuring added layers shared with visible / active layers and the visible layers control active upon load not show the Layermanager layers. Toggling the control twice makes them show ok.

There should be a method available to plugins to refresh the visible layers of the visible layers control.

This might be a part of a broader function that "resets" other things in the legend, for a sort of universal "legend reset", if there are other similar needs

**Additional context**
A function of legend.js that looks like so
```javascript
    redoShowVisibleLayers() {
      toggleShowVisibleLayers();
      toggleShowVisibleLayers();
    },
```
seems to do the trick however it isn't particularly elegant and it is limited to the visible layers control (which we may and may not like).
There seemed to be no annoying visual artifact from the visible layers list being reset, not with a limited amount of layers anyways.

Grammostola:
Aims to fix #1585 

It could be seen as potentially incomplete (setting setVisibleLayersViewActive(true) again as opposed to toggling as suggested in the issue. I've not encountered any ill effects in preliminary testing however. With this our Layermanager-csw can as part of its onAdd() do:
```javascript
      if(sharedLayers) {
        ReadAddedLayersFromMapState(sharedLayers, viewer);
        if (viewer.getControlByName('legend').getState().visibleLayersViewActive) {
          viewer.getControlByName('legend').resetVisibleLayersView();
        }
      }
```

### DIFF
--- a/src/controls/legend.js
+++ b/src/controls/legend.js
@@ -424,6 +424,7 @@ const Legend = function Legend(options = {}) {
     },
     getuseGroupIndication() { return useGroupIndication; },
     getOverlaysCollapse() { return overlaysCmp.overlaysCollapse; },
+    resetVisibleLayersView() { setVisibleLayersViewActive(true); },
     addButtonToTools(button) {
       const toolsEl = document.getElementById(toolsCmp.getId());
       toolsEl.classList.remove('hidden');
--- a/src/controls/legend.js
+++ b/src/controls/legend.js
@@ -424,7 +424,7 @@ const Legend = function Legend(options = {}) {
     },
     getuseGroupIndication() { return useGroupIndication; },
     getOverlaysCollapse() { return overlaysCmp.overlaysCollapse; },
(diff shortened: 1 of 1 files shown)


jokd:
Why not add setVisibleLayersViewActive to the component instead? Then it can be used both ways.
(Haven't tested, just looked at the code)

Grammostola:
This should be pretty ok to merge, I can't see how it would do any harm : )

tonnyandersson (approved):

PR #1621, opened 2022-10-20
fix: exclusive layers in legend search

### LINKED ISSUE #1599: Background layers in legend search
**Describe the bug**
When searching the legend, background layers are presented in the result list. Maybe they should be excluded, but otherwise when clicking the layer name and turn the layer on active background layers should be turned off. The same goes for layers in an "exclusive" group.

**To Reproduce**
1. Search for a background layer
2. Turn it on
3. You have two active background layers

**Expected behavior**
Max one layer in an exclusive group should be active.

 - OS: Windows
 - Browser Edge

jokd:
Fixes #1599 
Adds a check whether a layer is in an exclusive group or not when activating layers in the layer search result list.


johnnyblasta (approved):
LGTM

PR #1622, opened 2022-10-20
fix: splash control with multiple maps

### LINKED ISSUE #1591: Splash control doesn't consider what map it is
**Describe the bug**
Splash control doesn't consider what map it is. If you have mutiple maps, the content of splashContent item in local storage may be overwritten.

**Expected behavior**
The splash control should somehow distinguish what url/map is writing to local storage and consider that whether to splash or not.

jokd:
Fixes #1591 
Changes the key for splash control to consider the path of the map as well as map config.
Example key: `splashContent;/index.html;index` instead of just `splashContent`

This will make previous settings for splash visibility invalid so the splash will prompt no matter if the user has previously turned it off.

### DIFF
--- a/src/controls/splash.js
+++ b/src/controls/splash.js
@@ -9,6 +9,8 @@ const Splash = function Splash(options = {}) {
   let hideButton;
   let modal;
   let component;
+  let contentKey;
+  let visibilityKey;
 
   let {
     title,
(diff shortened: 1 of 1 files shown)


johnnyblasta (approved):
Tested and LGTM

PR #1623, opened 2022-10-27
feature: support for DMS in position control

### LINKED ISSUE #1602: Display WGS coordinates in DMS
It is common for various sub-cultures of GIS users, especially GPS users, aviation and naval applications, to use coordinates in WGS84 presented as as degrees, hours and minutes (DMS). Origo only supports decimal degrees, which for most people is not trivial to convert by mental arithmetic.

Possibility to display WGS84 coordinates using DMS as well as decimal degrees in the _position control_. The configuration should support adding the same coordinate system several times using different labels and different presentation formats.

Fishing up a reference to OL _MousePosition_ control and set a new coordinate format function and look in the DOM for the change coordinate system button and add a listener to change format function when coordinate system changes. But it feels kind of hacky as none of this is exposed by the _PositionControl_.

__Edit:__ The reverse function would also be desired, i.e. the go to position functionality in _PositionControl_ (Yes, there is one. Just click the hair cross and enter some coordinates in the textbox next to it) should accept DMS.

steff-o:
Closes #1602. This PR contains code to format coordinates in the position control  in DMS-format and also enter coordinates in DMS-format in the go to coordinate function.

As the existing configuration made it impossible to add the same coordinate system twice, I added a new configuration format. The old one is still valid for backwards compatibility, but lacks the new functionality.

In the new configuration format it is possible to specify dms as format and also an optional precision parameter that controls how many decimals a coordinate should have.

Example configuration:
```json
{
      "name": "position",
      "options": {
        "title": "Web Mercator",
        "projections": [
          {
            "projectionCode": "EPSG:4326",
            "projectionLabel": "Wgs84"
          },
          {
            "projectionCode": "EPSG:4326",
            "projectionLabel": "Wgs84 DMS",
            "dms": true,
            "precision": 2
          }
          ]
      }
    }
```

The big difference is that `projections` now is an array instead of an object in order to add the same coordinate system several times with different format.

The `projectionCode` property is mandatory and corresponds to the old configuration format's object key

The `projectionLabel` property is mandatory and specifies the text on the coordinate system selector button. It corresponds to the old configuration format's object value.

The `dms` property is optional and if set to true it is formatted as DMS.

The `precision` property is optional and if specified it sets the number of decimals. It works for both DMS and non-DMS. If not specified it defaults to 5 for EPSG:4326 in nonDMS format and 0 for everything else (same as before)

### DIFF
--- a/src/controls/position.js
+++ b/src/controls/position.js
@@ -138,7 +138,7 @@ const Position = function Position(options = {}) {
   }
 
   /**
-   * Update coords in input, transformin if necessary.
+   * Update coords in input, transforming if necessary.
    * @param {any} sourceCoords
    */
   function updateCoords(sourceCoords) {
@@ -254,7 +254,7 @@ const Position = function Position(options = {}) {
       } else if (e.which === 13) {
         findCoordinate();
       }
-      // For DMS, we handle everything ourselves. Ignore all keypresses (including current key) i order to keep browser not interfering
+      // For DMS, we handle everything ourselves. Ignore all keypresses (including current key) in order to keep browser from interfering
       e.preventDefault();
     } else if (e.which === 13) {
       findCoordinate();
--- a/src/controls/position.js
+++ b/src/controls/position.js
@@ -168,8 +168,8 @@ const Position = function Position(options = {}) {
     if (currentConfig.dms) {
       // Assume that input logic enforces a correct format
       const coordArray = strCoords.match(/\d*\.?\d+/g);
-      const lat = parseInt(coordArray[0], 10) + parseInt(coordArray[1], 10) / 60 + parseFloat(coordArray[2]) * 3600;
-      const lon = parseInt(coordArray[3], 10) + parseInt(coordArray[4], 10) / 60 + parseFloat(coordArray[5]) * 3600;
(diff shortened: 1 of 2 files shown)


johnnyblasta:
I'm trying to change position in DMS, but are getting an alert that I'm outside the map extent although it shouldn't be outside.

steff-o:
Strange. Never changed that stuff. It seems to check against both map extent and coordinate system extent, so if you have a local coordinate system (like sweref 17 15) it may throw on that.

Other possibility is that I messed up in transforming coordinates to map projection or parsing is just plain wrong.

I will look in to it!

steff-o:
My bad, I simplified the calculation of seconds * 1/3600 and kept the * instead of / after lint complained and obviously didn't test after. Fixed now.

src/controls/position.js
```diff
@@ -112,47 +137,58 @@ const Position = function Position(options = {}) {
...
 
-  function round(coords) {
-    if (precision) {
-      return coords.map(coord => coord.toFixed(precision));
-    }
-    return coords.map(coord => Math.round(coord));
-  }
-
+  /**
+   * Update coords in input, transformin if necessary.
```
johnnyblasta, line 141: Missing g i transforming

johnnyblasta (approved):
LGTM, except some misspellings in comments

steff-o:
Tough crowd... Fixed comments.

PR #1624, opened 2022-10-28
feature: support for value and label in searchList

### LINKED ISSUE #1608: Add support for Label in SearchList attribute editor
SearchList can only work with _Value_, not _Label_ and _Value_, which means that _Value_ is used as both _Label_ and _Value_ in the drop down.

The list should support both _Label_ and _Value_ so that _Label_ is displayed when searching, but _Value_ is sent to database when saving. Sort of like a normal HTML dropdown.

Also, the reverse mapping should also work, i.e. when opening the dialog the _Label_ for the current _Value_ should be displayed.

As usual, do some server side magic, but as it should support WFS access it quickly becomes nasty.

**Additional context**

steff-o:
Closes #1608  
SearchList now offers a value-label dropdown similar to HTML select.

To access this new feature a label is added to each value in the list object like this:

```json
"list": [
            {
              "label": "Sel",
              "value": "1"
            },
            {
              "label": "Stan",
              "value": "2"
            },
            {
              "label": "Obygden",
              "value": "3"
            }
          ],
```

And in the _config_ property a `"useBackingValue": true,` setting must be added. Configuration is fully backwards compatible. This setting implies `allowOnlyfromList`, so it does not have to be set implicitly.

What happens is that the searchList displays and searches on _label_ but _value_ is stored in database. When opening the searchList, a reverse lookup is performed. When using a dynamic remote source, a new server endpoint must be implemented in order for the reverse lookup to work. The reverse lookup uses the same `url`-setting as the usual remote endpoint but uses another query parameter.

The remote endpoint should support a GET request looking like this: <url>?<valueQueryParameter>=<value>, where <url> is the configured url,  _<valueQueryParameter>_ is the configured valueQueryParameter or defults to "value" if not specified. The <value> is the current value of the attribute. The server should then respond  as the usual non-reverse lookup endpoint, i.e. a list of label-value pairs, but only the first entry is used. The list of label-value pairs looks like the static configuration above.

Here is a complete example configuration for a dynamic searchList displaying the new features. Not present settings will default as described in the documentation:
```json
"attributes": [
        {
          "name": "text",
          "title": "Some Text",
          "type": "searchList",
          "config": {
            "url": "http://localhost:3000/domain/personer",
            "valueQueryParameter": "value",
            "dynamic": true,
            "useBackingValue": true,
          }
        }
      ]
```

### DIFF
--- a/src/controls/editor/addons/searchList/searchList.js
+++ b/src/controls/editor/addons/searchList/searchList.js
@@ -72,9 +72,10 @@ function searchList(input, conf) {
   // useBackingValue implies allowOnlyFromList, otherwise it would be rather silly.
   const allowOnlyFromList = searchListConfig.allowOnlyFromList || searchListConfig.useBackingValue;
   // Set up default config for awesome
+  searchListConfig.minChars = searchListConfig.minChars === undefined ? 2 : searchListConfig.minChars;
   const awesomeConfig = {
     minChars: searchListConfig.minChars,
-    maxItems: searchListConfig.maxItems,
+    maxItems: searchListConfig.maxItems || 10,
     filter,
     item: imageItemFormatter
   };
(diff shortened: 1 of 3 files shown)


jokd:
LGTM except for a couple of lint errors in edithandler.js

PR #1625, opened 2022-10-28
feature: editor api

### LINKED ISSUE #1615: Expose methods on Editor Component to add, edit and remove features
There is no (official) way to communicate with the _edithandler_ from other Components, which makes it hard to programmatically control edit flow.

Expose methods on the Edit Component to:
1. Add a new feature to a layer
2. Open attribute editor for a feature
3. Delete a feature

Throwing some obscure DOM events at document level to affect the editor and grab the _SelectInteraction_ from the map and modify the selection in order to edit attributes. It might work, but would be very ugly.

It would also be possible to find the editor toolbar in the DOM and emulate clicks on the buttons, but that would be even uglier.

**Additional context**
This is a concretization of the ideas in #1613

### LINKED ISSUE #1616: Support features without geometries in editor
The _edithandler_ assumes that a feature and the layer has a geometry and a geometry type for some operations. Currently it is almost impossible to get into a situation where that is a problem, but if #1615 is implemented or someone is adding features fr

Make all code in the edit flow honor the _isTable_ setting on a layer and not require a geometry. (Or by other means allow features without geometry where it makes sense)

Adding a dummy geometry to layers without geometry. But it seems rather silly when we already have the _isTable_ setting.

**Additional context**
This is a concretization of the ideas in #1613

steff-o:
Closes #1615 and closes #1616. (Sorry had to implement them together)

Adds api functions to the Edit-component in order to make it possible to implement plug-in controls that can access the editor without having to make the user interact with the map or the editor toolbar.

# Implemented api-functions
All functions can be called without the editor actually being active (toolbar visible) and are accessed by getting a reference to the Edit-control: `viewer.getControlByName('editor');` and executing the methods on the component. All functions honor autoSave-setting.
## Change active layer
_changeActiveLayer:(layerName)_ : Changes the editor's active layer. The layer must be a layer that is not configured as a table, i.e. it should have a geometry and configured as editable as usual. The selected layer is indicated in the toolbar.

## Create new feature
_async function createFeature(layerName, geometry = null)_: Creates a new empty feature in the indicated layer. The layer does not have to be the same as the editor's active layer and the editor's active layer is NOT changed. The layer can be configured as `isTable`. If layer has geometry, a geometry should be passed as parameter. When the feature is created it is returned (in the promise). If autoSave is set, it will not return until it is saved. If autoForm is set the attribute editor will open after this function has returned.
## Delete feature
_async function deleteFeature(featureId, layerName)_: deletes a feature, honoring cascading delete setting for related tables. The layer does not have to be the same as the editor's active layer and the editor's active layer is NOT changed. The layer can be configured as `isTable`.  If autoSave is set it returns (in promise) when deleted from database.
## Open Attribute Editor
_function editFeatureAttributes(featureId, layerName)_: Opens the edit attributes dialog for the feature and returns immediately. The layer does not have to be the same as the editor's active layer and the editor's active layer is NOT changed. The layer can be configured as `isTable`. If autoSave is set the feature is saved when the dialog closes as usual.

### DIFF
--- a/src/controls/editor.js
+++ b/src/controls/editor.js
@@ -1,5 +1,6 @@
 import { Component, Button, dom } from '../ui';
 import editorToolbar from './editor/editortoolbar';
+import EditHandler from './editor/edithandler';
 
 const Editor = function Editor(options = {}) {
   const {
@@ -10,12 +11,19 @@ const Editor = function Editor(options = {}) {
   let editorButton;
   let target;
   let viewer;
+  let isVisible = isActive;
+
+  /** The handler were all state is kept */
+  let editHandler;
 
   const toggleState = function toggleState() {
     const detail = {
       name: 'editor',
       active: editorButton.getState() === 'initial'
     };
+    // There are some serious event dependencies between viewer, editor, edithandler, editortoolbar, editorlayers, dropdown and editorbutton,
+    // which makes it almost impossible to do things in correct order.
+    isVisible = detail.active;
     viewer.dispatch('toggleClickInteraction', detail);
   };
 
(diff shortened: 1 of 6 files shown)


jokd:
When switching layer with `changeActiveLayer`, the drawTools (the submenu that can appear when clicking the +-button) menu isn't updated.

steff-o:
Yep. Missed that. The Drawtool selector listens to the event sent by layer selector and that is not sent when called from api. Not emitting the event is deliberately as I want to keep as far away from that eventhandling.

Also saw that I forgot to update the available tools when active layer is changed. 

Fix coming up!

steff-o:
Fixed allowedTools and drawTools.

jokd (approved):

PR #1626, opened 2022-10-28
fix: reset features style after print control

### LINKED ISSUE #1605: Print messes with LMSearch
**Describe the bug**

**To Reproduce**
1. Search for a estate
2. Go to print
3. Close print
4. See that the style has been changed

**Expected behavior**
The style should be the same before and after using print.

**Screenshots**
Before
image
After
image

johnnyblasta:
Fixes #1605 
Solves issue that arises when style is set on individual features for example in measure control, drag and drop control and LMSearch plugin.

### DIFF
--- a/src/controls/print/print-resize.js
+++ b/src/controls/print/print-resize.js
@@ -46,44 +46,48 @@ export default function PrintResize(options = {}) {
 
   // Resize features when DPI changes
   const resizeFeature = function resizeFeature(style, feature, styleScale) {
-    const image = style.getImage();
-    if (image) {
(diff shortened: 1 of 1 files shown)


jokd:
When changing dpi in the print layout the style is not reset.
E.g measure a line, -> print, change to high res, close print window and the line will be wider than before.

johnnyblasta:
> When changing dpi in the print layout the style is not reset. E.g measure a line, -> print, change to high res, close print window and the line will be wider than before.

Made a fix for this.

jokd (approved):

PR #1630, opened 2022-11-09
Only header draggable in infowindow expandable list.

### LINKED ISSUE #1628: Can not select text in infowindow when using listlayout
**Describe the bug**
It is not possible to select text for copy-paste when using the `listLayout` option on infowindow, instead the window is dragged.

**To Reproduce**
1. Configure app to use
 ```
"featureinfoOptions": {
    "infowindow": "infowindow",  
    "infowindowOptions": {
      "listLayout": true
    }
  }
```
2. Click on a feature to open infowindow
3. Try to select text by pressing down the mouse and while keeping it pressed move the mouse
4. The entire info window is dragged

**Expected behavior**
The text should be selected. This is what happens when using the "old" infowindow. Also the mouse pointer is only _move cursor_ when hovering over the title field, which indicates that this is the intended behavior (or a copy paste error from old infowindow).

steff-o:
Fixes #1628 by making only the header draggable in infowindow when using expandable list.

Before: Entire window was draggable

Now: Only header is draggable, just like when using "old" infowindow.

### DIFF
--- a/src/infowindow_expandableList.js
+++ b/src/infowindow_expandableList.js
@@ -135,10 +135,12 @@ function render(viewerId) {
   mainContainer = document.createElement('div');
   setInfowindowStyle();
   mainContainer.classList.add('sidebarcontainer', 'expandable_list');
-  mainContainer.id = 'sidebarcontainer-draggable';
+  mainContainer.id = 'sidebarcontainer';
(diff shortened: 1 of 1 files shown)


johnnyblasta (approved):
LGTM

PR #1633, opened 2022-11-21
feature: enable WMS-layers in stylepicker

### LINKED ISSUE #1583: Alternative styles for WMS layers
This relates to #1020 

-> That alternative styles for a WMS layer can be shown in the same way in a layer's properties in the legend as is possible today via the stylepicker for vector layers

That getCapabilities/ a capabilities doc is used to find alternative styles.

The names of the on the WMS server available styles for a layer are defined in the map config  for that layer 

**Additional context**
It should work with added layers (we are using the Layermanager plugin for this). This can be made to work via the preferred as well as alt solution in that getCapabilities could be fired at the workspace of every added layer every time or less often; the names of alternative styles could also be part of a layer's external metadata (accessed via our Layermanager)

Edit. I discussed this with a few colleagues and they expressed a preferance for the alternative solution. It would, well look more like alternative styles defined for vector layers.

sweco-semvtw:
Proposes to fix #1583 

Allows users to declare alternate styles for WMS-layers in the stylepicker. It is possible to declare whether a style is has theme. A title for the default style is declared by setting the `default` flag. Chosen style will be saved when sharing the map. 

Example of how to format the settings in json:
es-85

es-85-legend

Will also take style-url for WMS-layers declared in json document into consideration before the url from layer source in the print legend.

### DIFF
--- a/src/controls/legend/overlayproperties.js
+++ b/src/controls/legend/overlayproperties.js
@@ -63,7 +63,10 @@ const OverlayProperties = function OverlayProperties(options = {}) {
   }
 
   function getStyleDisplayName(styleName) {
-    const altStyle = stylePicker.find(s => s.style === styleName);
+    let altStyle = stylePicker.find(s => s.style === styleName);
+    if (!altStyle) {
+      altStyle = stylePicker.find(s => s.default);
+    }
     return (altStyle && altStyle.title) || styleName;
   }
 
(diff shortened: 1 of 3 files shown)


johnnyblasta (commented):
Had a problem with getting it to work, so did some investagating. Seems the problem was that we used workspace in layer name and therefore the stylepicker config wasn't found.
My config that don't work:
image
Config that works:
image

Grammostola:
@johnnyblasta Yes, I noticed the same. It seems the same for the current master for WFS-layers too, uncertain if a fix is needed as part of this PR. It could be nice I guess.

I've tested it with `renderMode: image` WMS layers and tiled, shared as mapstate and url hash, printed with legend and it seems to work as intended. Haven't noticed ill effects for the print legend or sharing with non-alt-style WMS or WFS layers or with alt-style WFS layers.

johnnyblasta:
> @johnnyblasta Yes, I noticed the same. It seems the same for the current master for WFS-layers too, uncertain if a fix is needed as part of this PR. It could be nice I guess.

@Grammostola well if it is the same for wfs it's a separate preexisting issue not caused by this and can be solved in a standalone bug issue.

Grammostola:
@johnnyblasta indeed, I'll create an issue for it tomorrow.

johnnyblasta (approved):
Tested and works as it should except when configuring as in #1634 but since that is a pre existing problem it should not hinder approving this PR.

PR #1635, opened 2022-12-06
feature: hide controls instead of not adding them on "hideWhenEmbedded"

### LINKED ISSUE #1632: Hide mapmenu when embedded results in error
**Describe the bug**
When setting the mapmenu `"hideWhenEmbedded": true` results in a error.

**To Reproduce**
1. Config the mapmenu to `"hideWhenEmbedded": true`
2. Open the map embedded
3. Results in a error for example when adding sharemap or print control to the menu

**Expected behavior**
The map shouldn't cause errors when adding object to hidden menu.

johnnyblasta:
Closes #1632 

Hides controls with visible buttons instead of simply not adding them when using `hideWhenEmbedded`. Some special cases still don't adds the control for example those that is added as a menuitem and also drag and drop control which hasn't a visible component on the map.


johnnyblasta:
Changed so that controls that hasn't have a hide function isn't added when embedded, so that the fallback that was there previously might have caused unintended problems later on.

jokd:
Shouldn't hide-method simply add `hidden` class to the element? Otherwise maybe the method should be named toggle

johnnyblasta:
> Shouldn't hide-method simply add `hidden` class to the element? Otherwise maybe the method should be named toggle

Did it like so to make it possible to unhide it programmatically. Could have done both hide and unhide method, but probably felt lazy that day.

jokd:
But if you don't know that it is hidden and you `hide()` it then it will be visible. Thinking from an API point of view.

johnnyblasta:
> But if you don't know that it is hidden and you `hide()` it then it will be visible. Thinking from an API point of view.

Made a switch from hide "toggle" to separate hide/unhide methods.

Grammostola:
Looks good as far as I can tell. Calling "hide()" when hidden has no effect would make something communicating via the api less dependant upon checking whether hidden first. I was thinking about print and seem to recall a discussion regarding a need for it to be added to an embedded map without showing. (print can also be configured with undocumented `placement` to be a westerly button)

Issue #1636, opened 2022-12-09
Update plugins.md


filleg:
**Is your feature request related to a problem? Please describe.**
Plugind.md needs a small information bump. 

**Describe the solution you'd like**
Proposed changes incoming in PR.

PR #1637, opened 2022-12-09
docs: Plugins doc

filleg:
Proposed elaboration on plugins-documentation in plugins.md. 
Additional improvements are more than welcome. 
Fixes: #1636


tonnyandersson:
Just drop the dash in "origo-core" (occurs twice) and we're good to go.

tonnyandersson:
I think it looks better with a space in there, like "Origo core". Otherwise it sounds like a brand name or something.

PR #1648, opened 2023-01-09
dep: OL v7.2.2

### LINKED ISSUE #1647: OL v7.2.2
New version of OpenLayers out!

jokd:
Fixes #1647 
New version of OpenLayers and removed monkey patch in print component since OL now have native support for transform: scale


johnnyblasta (commented):
Tested with some local configs and seems to still work, so LGTM.
A conflict needs to be resolved.

PR #1652, opened 2023-01-10
feat: add imagefeatureinfomode featureinfo option to control featurei…

### LINKED ISSUE #1651: WMS: There ought to be a featureInfoOption for always sending getFeatureInfo-requests
**Issue**

Sometimes WMS services whose style cannot easily be modified are employed in a map. If these contain polygon layers with totally transparent fillings then currently if infoclicked on the transparency nothing appears to happen (and no getFeatureInfo request is sent). This was not the case in Origo 2.4 as it seems because Openlayers and Origo worked differently then and always sent a getFeatureInfoRequest for each visible (WMS) layer in the map.

An featureInfoOption in the settings json for always sending a getFeatureInfo request to all visible WMS layers in the map. Option because lots of requests might be less desirable than the potential for totally transparent polygon styles.

**Additional context**
https://github.com/origo-map/origo/issues/1587
https://github.com/origo-map/origo/issues/513
https://github.com/openlayers/openlayers/releases/tag/v6.13.0

Grammostola:
…nfo requests for image type layers. 'location' is the current and default behaviour. It seems to work the same for AGS_TILE and WMS layers. There are poly layers with transparency inside at https://ext-geodata-nationella.lansstyrelsen.se/arcgis/services/VM/lst_vm_wms_miljokvalitetsnorm/MapServer/WMSServer for instance.

Aims to fix #1651 

Used like so:
```json
  "featureinfoOptions": {
    "imageFeatureInfoMode": "visible",
    "infowindow": "overlay"
  },
```
where `ìmageFeatureInfoMode` can be `always` (a click will produce a featureinfo if there's a feature at the location regardless of whether there's a layer turned on), `visible` (a click will produce a featureinfo if there's a feature of a layer that is turned on at the location) and `location` which is default and the current behaviour, ie a click will produce a featureinfo if there's a feature of a layer that is turned on at the location and that feature isn't wholly transparent at the location.

### DIFF
--- a/src/getfeatureinfo.js
+++ b/src/getfeatureinfo.js
@@ -185,9 +185,20 @@ function getFeatureInfoRequests({
   const requests = [];
   const queryableLayers = viewer.getLayersByProperty('queryable', true);
   const layerGroups = queryableLayers.filter(layer => layer instanceof LayerGroup);
-  if (layerGroups) { layerGroups.forEach(item => item.getLayersArray().forEach(element => queryableLayers.push(element))); }
-  const imageLayers = queryableLayers.map(layer => layer instanceof BaseTileLayer || layer instanceof ImageLayer);
+  if (layerGroups) {
+    const visibleGroups = layerGroups.filter(group => group.get('visible') === true);
+    if (visibleGroups.length) {
+      visibleGroups.forEach(visibleGroup => visibleGroup.getLayersArray().forEach(layer => queryableLayers.push(layer)));
+    } else {
+      layerGroups.forEach(layerGroup => layerGroup.getLayersArray().forEach(layer => {
+        if ((layer.get('imageFeatureInfoMode')) && (layer.get('imageFeatureInfoMode') === 'always')) {
+          queryableLayers.push(layer);
+        }
+      }));
+    }
+  }
 
+  const imageLayers = queryableLayers.filter(layer => layer instanceof BaseTileLayer || layer instanceof ImageLayer);
   imageLayers.forEach(layer => {
     let item;
     let imageInfoMode;
@@ -201,7 +212,7 @@ function getFeatureInfoRequests({
       if (pixelVal instanceof Uint8ClampedArray && pixelVal[3] > 0) {
         item = getGetFeatureInfoRequest({ layer, coordinate }, viewer);
       }
-    } else if (imageInfoMode === 'visible' && layer.get('visible') === true) {
+    } else if ((imageInfoMode === 'visible') && (layer.get('visible') === true)) {
       item = getGetFeatureInfoRequest({ layer, coordinate }, viewer);
     } else if (imageInfoMode === 'always') {
       item = getGetFeatureInfoRequest({ layer, coordinate }, viewer);
--- a/src/getfeatureinfo.js
+++ b/src/getfeatureinfo.js
@@ -181,19 +181,27 @@ function getFeatureInfoRequests({
   coordinate,
   pixel
 }, viewer) {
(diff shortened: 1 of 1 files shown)


jokd:
Wouldn't it be better if this was a per-layer setting?

Grammostola:
It's an interesting suggestion. It wouldn't be for our municipality's application, we have quite a lot of layers in several maps and we always want the apps to behave as in ~Origo 2.4, "if a feature is clicked on, regardless of its style, show feature info". A per-layer configuration would make that possible, albeit with more configuration.

jokd:
I think that if the 'always' option shall be useful it is necessary. Can we do both on map and layer level?

Grammostola:
Would it be reasonable to see the `visible` and `location` currenty suggested as a map setting (`imageFeatureInfoMode` ) two standard forms of a common feature whereas `always` is rather a specialized creature (that would be a layer property, perhaps `searchable: always` instead of `searchable:true` and if such a layer was turned on would respond as whatever `imageFeatureInfoMode` was set to)?

jokd:
Yeah that sounds reasonable. However if you have like one layer with a problematic styling in your config you would like to set 'visible' on just that layer and keep 'location' on map level to avoid numerous unnecessary getfeatureinfo requests.

Grammostola:
Is this somewhat understandable:

`imageFeatureInfoMode`: `location`(default) or `visible` 


per layer (takes precedence where contradicts the above):
`queryable`: `visible` or `location` or `always` or `true` or `false`

jokd:
Where true is the only layer setting actually letting imageFeatureInfoMode kick in?
Works for me, maybe someone else has an opinion?
location maybe should be renamed pixel?
and imageFeatureInfoMode renamed getFeatureInfoMode?

Grammostola:
> Where true is the only layer setting actually letting imageFeatureInfoMode kick in?
yes
> location maybe should be renamed pixel?
agreed
> and imageFeatureInfoMode renamed getFeatureInfoMode? 
for WMS services the request is getFeatureInfo, for AGS_* services the request is called differently

tonnyandersson:
I just skimmed through the conversation quickly so I might have missed something, but... Is it really necessary to have all those different settings, like `always`, `visible`, `sometimes`, `never`, `true`, `false` etc.? Isn't the current `queryable: true | false` option enough if we just make it work for transparent layers as well?

Grammostola:
>  Isn't the current queryable: true | false option enough if we just make it work for transparent layers as well?

It is, if we accept one getFeatureInfo / AGS_TILE/MAP request for every turned on layer in the map. As Origo functions now it avoids those by checking the pixel value of the clicked location which I perceive to be desired.

Anyway I've made the changes for the last suggested version and can push unless this discussion goes elsewhere : )

tonnyandersson:
In the issue you were talking about fully transparent polygons so I thought that was the actual issue here. We can skip the transparency discussion for now, because that's not actually a part of my point.

I just don't see the need for additional options. If a layer is `queryable: true` AND `visible: true`, a getFeatureInfo request should be sent. If either is set to `false`, a getFeatureInfo request should _not_ be sent.

In what situation is that not enough?

Grammostola:
A wholly transparent part of a polygon style is the point because the above quoted piece of code decides not to send a featureinfo request for a specific layer if you missed clicking on one of its features or if you did click on a transparent bit of one of its features. If the check were to be skipped it would send one featureinfo request for every turned on layer regardless of where you click.

tonnyandersson:
OK, I think I get the "why" now. Sorry for being a bit slow.

IMO, a few extra "unwanted" requests is better than adding a bunch of confusing config options on different levels, unless there's a risk for false positives with such requests.

jokd (approved):
LGTM

PR #1658, opened 2023-01-25
fix: print fix ios

### LINKED ISSUE #1619: Print - PDF export not working properly in IOS Safari
**Describe the bug**
PDF export is a bit sketchy in IOS Safari. The map is sometimes blank in the export and sometimes present but with the top-left corner anchored to the center of the layout.

**To Reproduce**
1. Open an Origo map (master version) in IOS Safari
2. Open print control
3. Choose PDF export

 - Device: Iphone 7
 - OS: IOS 15.7
 - Browser: Safari

**Additional context**
Works fine on Android device.

jokd:
Fixes #1619 
On iOS devices it recalculates and sets the resolution when print size changes to stay below threshold for canvas size o iOS.
Changes the library used for create the map image.
Print preview is now better scaled to the browser window.


tonnyandersson:
Seems to work quite well.

Two things, though:

- Print layout is no longer centered on desktop, it's left aligned. This seems to work well on mobile devices, but...
- Print margins are no longer toggleable.

jokd:
> Seems to work quite well.
> 
> Two things, though:
> 
> * Print layout is no longer centered on desktop, it's left aligned. This seems to work well on mobile devices, but...
> * Print margins are no longer toggleable.

Its left aligned to look good on mobile but maybe I can make fix that in the css.
I am able to toggle the margins in my build but maybe I messed up while cleaning the code, will have another look.

jokd:
I added a simple check if the screen is above 1000px width or not and placed the preview accordingly. Perhaps we can look into a more robust solution later on.

tonnyandersson:
Now the print margin size changes when changing print resolution, and not in a good way. And toggling the margins off and on again changes the margin size back to the original size.

jokd:
Yeah that was a bit... brutal. I moved the function so it'll recalculate when switching eg resolution too. Went for the same padding all way round as well.

tonnyandersson:
Works better now!

What was the reason for changing the way the print margins are handled in the first place? Just curious because I never noticed any issues with the margins before.

jokd:
html2image wasn't handling it very well

tonnyandersson:
OK.

LGTM, then. Well done!

tonnyandersson (approved):

jokd:
@Grammostola Any thoughts on this since you have developed alot of the printing functionality?

Grammostola:
> since you have developed alot of 

🐎 

In simulation mode on FF iPhone SE things look good in the preview, resolution change and then margins on off, -> pdf, but not -> png. The scale bar isn't positioned right and the legend gets truncated. On desktop the legend gets a little truncated.

jokd:
@Grammostola Have another go

PR #1663, opened 2023-02-01
fix: really disable print buttons while print preview map is loading

### LINKED ISSUE #1661: Print button not really disabled
**Describe the bug**
The buttons in print for saving image and create pdf isn't really disabled while loading the map. The buttons looks diasbled while loading the map , but they can stil be clicked on and will create a map which isn't fully loaded.

**To Reproduce**
See animation of problem:
Print-button-not-working

**Expected behavior**
The buttons should not be clickable while the map is being loaded.

Grammostola:
Aims to fix #1661 
(and prevent unnecessary event listening when the print panel isn't active)

### DIFF
--- a/scss/_print.scss
+++ b/scss/_print.scss
@@ -16,6 +16,11 @@ $print-scale-line-font-size: $font-size-smallest;
   overflow: hidden;
 }
 
+// for print buttons while print preview map is loading
+.print-button-disable {
+  pointer-events: none;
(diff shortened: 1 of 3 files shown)


jokd:
Go for number 2

Grammostola:
Second gif is implemented.

jokd (approved):

PR #1664, opened 2023-02-01
fix: search result not hidden in legend search

### LINKED ISSUE #1641: Legend search - the list of results is only displayed within the legend
The list of results from the legend search is only displayed within the legend window. If you search in a legend with, for example just one group, you may not see all search results.

The result list should display regardless of the size of the legend.

legend-search-result-list

johnnyblasta:
Fixes #1641 
Had to set a fixed width to input and result list to not have it runaway outside page. Perhaps some css guru has a fix for this?

### DIFF
--- a/scss/externs/_awesomplete.base.scss
+++ b/scss/externs/_awesomplete.base.scss
@@ -20,7 +20,7 @@
   flex-direction: row;
   flex-grow: 1;
   position: fixed;
-  min-width: 180px;
+  min-width: 100%;
 }
 
 .awesomplete > input {
@@ -47,7 +47,7 @@
   left: 0px;
   list-style: none;
   margin: 0;
-  min-width: 180px;
+  min-width: 100%;
   padding: 0;
   position: absolute;
   top: auto;
--- a/scss/externs/_awesomplete.base.scss
+++ b/scss/externs/_awesomplete.base.scss
@@ -49,6 +49,7 @@
   margin: 0;
   min-width: 180px;
   max-width: 220px;
+  overflow-wrap: break-word;
   padding: 0;
   position: absolute;
   top: auto;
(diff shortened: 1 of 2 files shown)


johnnyblasta:
Damn, I found a bug with this myself. Will try to rework it.

johnnyblasta:
> Damn, I found a bug with this myself. Will try to rework it.

Bug handled.

johnnyblasta:
Demo
Legend search

Grammostola:
Hi, it seems to fix the reported issue, nicely done.

I noted that if the legend search is specified to be able to search layer names (that won't auto rewidth the legend) the hit list can run off the page. If it could detect the end or perhaps the legend width and newline there that could be good.

johnnyblasta:
> Hi, it seems to fix the reported issue, nicely done.
> 
> I noted that if the legend search is specified to be able to search layer names (that won't auto rewidth the legend) the hit list can run off the page. If it could detect the end or perhaps the legend width and newline there that could be good.

Put a max-width on search result, so that a new line is forced on long results and it don't falls outside screen.

Grammostola:
It seems the black search hit area does respect the max width though long names don't. An arbitrary newline in a layername might not be the best way to handle the situation.
bild

johnnyblasta:
> It seems the black search hit area does respect the max width though long names don't. An arbitrary newline in a layername might not be the best way to handle the situation. bild

Can solve it with `overflow-wrap: break-word;` and it will break the long word at arbitrary place:
image

Is this the preferred way to handle it? @Grammostola

Grammostola:
Hahaha, good example. Yes I definitely think it is the way.

Grammostola:
I quite like it. When the legend is very wide it can look a bit funny but I think that is preferrable to a layer name or perhaps abstract running off the page. So while it does look good to me, it being a cosmetics issue the more eyes the merrier

PR #1666, opened 2023-02-03
fix: crash when printing on iOS

### LINKED ISSUE #1665: Print crashes on iOS
**Describe the bug**
When loading the print preview the map sometimes crashes when using an iOS device. 

 - Device: iPhone 12
 - OS: iOS 15
 - Browser Safari, Edge, Chrome

jokd:
Fixes #1665 
Lowered the pixelratio when printing if on iOS device. Also added touch.action: none for the print container since pinch zooming in the print preview also could make the print function crash.


tonnyandersson:
Works better now. Still possible to crash the app in A3 mode with certain configs though, but it's definitely a step in the right direction. And I think it actually gets a little bit better with the changes coming in via #1663 as well!

tonnyandersson (approved):

PR #1675, opened 2023-02-08
feature: confirm removal of layer

### LINKED ISSUE #1674: Option to confirm removal of layer
It is sometime useful to require the user to confirm removal of a layer eg in an application where you actually delete a layer.

Popup a confirm window.

jokd:
Fixes #1674
Setting "confirmRemoval": true pops up a confirm dialog window This commit also includes also removes redundant calls to removeOverlay-method and relies on the event instead.

### DIFF
--- a/src/controls/legend/overlay.js
+++ b/src/controls/legend/overlay.js
@@ -222,14 +222,10 @@ const OverlayLayer = function OverlayLayer(options) {
       onRender() {
         const labelEl = document.getElementById(this.getId());
         labelEl.addEventListener('click', (e) => {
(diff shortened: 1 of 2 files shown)


johnnyblasta (approved):
Works fine, perhaps something else than the ugly window.confirm could have been used but it does the trick.

jokd:
Yeah we could have an Origomodal for that

johnnyblasta:
> Yeah we could have an Origomodal for that

I made createRadiusModal for measure control but didn't make it a generic modal.

PR #1677, opened 2023-02-08
feat: add feature styling

### LINKED ISSUE #1676: Style feature based on attribute
If a feature has a style object as an attribute this can be used when setting the style of a feature instead of a static layer styling

Config option for the layer and it should be optional whether to use feature styling or layer styling.

jokd:
Fixes #1676
Adds the possibility to set the drag and drop control as well as a geojson source to style features based on the "style" attribute of the feature rather than the layer style. Configured as `"styleByAttribute":true` as options for the drag and drop control or property of a geojson-layer. Set to false by default. This functionlity is most useful with export/import/sharing functionality.

Example usage:
```
    {
      "name": "draganddrop",
      "options": {
        "styleByAttribute":true
      }
    }
```
OR
```
    {
      "name": "origo-mask",
      "opacity": 0.25,
      "title": "origo-mask",
      "group": "none",
      "queryable": false,
      "source": "data/origo-mask-3857.geojson",
      "style": "mask",
      "type": "GEOJSON",
      "visible": true,
      "styleByAttribute": true
    }
```
Origo mask geojson with style object:
```
............."features": [
{ "type": "Feature", "properties": {"style":{"fillColor":"rgba(255,120,20,0.8)","fillOpacity":0.8,"strokeColor":"rgba(255, 0, 0, 1)","strokeOpacity":1,"strokeWidth":"6","strokeType":"dash"}}, "geometry": { "type": "MultiPolygon", "coordinates": [ [ [ [ -19679343, -19747663 ],....................
```
And it will override the layer setting:
image

### DIFF
--- a/src/controls/draganddrop.js
+++ b/src/controls/draganddrop.js
@@ -8,6 +8,7 @@ import VectorSource from 'ol/source/Vector';
 import VectorLayer from 'ol/layer/Vector';
 import Style from '../style';
 import { Component } from '../ui';
+import { getStylewindowStyle } from './editor/stylewindow';
 
 const DragAndDrop = function DragAndDrop(options = {}) {
   let dragAndDrop;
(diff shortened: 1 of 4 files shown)


johnnyblasta (approved):
Great stuff!
LGTM

PR #1682, opened 2023-02-10
feature: featureinfo templates with alias and API functionality

### LINKED ISSUE #1681: Use alias in featureinfo templates
It is sometimes useful to have aliases instead of attribute names presented when using the feature info template function. For instance whn adding layers with layer manager that aren't preconfigured.

A config setting.

jokd:
Fixes #1681 and #1690
Configured like:
```
  "attributeAlias": {
    "name":"Namn"
  }
```
in the config. Can be tested by removing the attribute setting for the origo cities layer in the standard map.

### DIFF
--- a/src/map.js
+++ b/src/map.js
@@ -4,10 +4,22 @@ import mapInteractions from './mapinteractions';
 
 const Map = (options = {}) => {
   const interactions = mapInteractions({ target: options.target, mapInteractions: options.pageSettings && options.pageSettings.mapInteractions ? options.pageSettings.mapInteractions : {} });
-  const mapConfig = Object.assign({}, options);
-  delete mapConfig.layers;
-  delete mapConfig.styles;
-  delete mapConfig.source;
+  const mapConfig = {};
+  const keys = Object.keys(options);
+  keys.forEach((key) => {
+    switch (key) {
+      case 'controls':
+      case 'defaultControls':
+      case 'groups':
+      case 'layers':
+      case 'source':
+      case 'styles':
+        break;
+      default:
+        mapConfig[key] = options[key];
+    }
+  });
+
   const mapOptions = Object.assign(options, { interactions });
   delete mapOptions.layers;
   mapOptions.controls = [];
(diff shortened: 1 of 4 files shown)


jokd:
Added API functionality to add feature info templates without the need for a custom build (#1690). Can be used like this to alter the default template and using the templateHelper function:
```
const featureinfotemplates = origo.api().getFeatureinfo().featureinfotemplates;
featureinfotemplates.addFeatureinfotemplate('default', function(properties) {
  const els = `${featureinfotemplates.templateHelpers.each(properties, obj => `<li><h5>${obj.prop}:</h5><b>${obj.value}</b></li>`)}`;
  return els;
});
```

jokd:
image
image

src/map.js
```diff
@@ -4,13 +4,17 @@ import mapInteractions from './mapinteractions';
 
 const Map = (options = {}) => {
   const interactions = mapInteractions({ target: options.target, mapInteractions: options.pageSettings && options.pageSettings.mapInteractions ? options.pageSettings.mapInteractions : {} });
+  const mapConfig = Object.assign({}, options);
+  delete mapConfig.layers;
+  delete mapConfig.styles;
+  delete mapConfig.source;
```
johnnyblasta, line 10: Can this be done without first assign and then delete? Layers and styles can be quite large so wasting memory perhaps.

jokd: I suppose a bit more verbose solution is to loop through an array of the keys and skip some of them.

jokd: Rewrote this a bit

johnnyblasta:
@jokd is it supposed to be a space between colon and attribute alias?
image

jokd:
@johnnyblasta That's the way the default template looks (haven't touched it). But I wouldn't object is anyone was to change it.

johnnyblasta (approved):
LGTM

johnnyblasta:
> @johnnyblasta That's the way the default template looks (haven't touched it). But I wouldn't object is anyone was to change it.

OK, thought it was something you had introduced. It looked a bit strange, so perhaps a change would be good.

PR #1685, opened 2023-02-10
feature: import layer button

### LINKED ISSUE #1678: Add layer with button instead of drag and drop
Drag and drop is super cool and all but not the always most intuitive way of adding a layer to the map. Therefore it would be nice to also have the possibility to add layer by clicking a button. 

A "Add layer"-button in the drag and drop legend group that mimic a file that is dropped.

A stand alone solution but then we won't be able to use the bells and whistles that comes with the drag and drop interaction.

jokd:
Fixes #1678
Adds a button in the legend for import of layer as file. Future improvement can be to put this along with the layer manager plugin in a context menu. Defaults to false.
```
    {
      "name": "draganddrop",
      "options": {
        "showLegendButton": true
      }
    },
```


johnnyblasta:
Tested, but got an error:

image

johnnyblasta:
> Tested, but got an error:
> 
> image

Oops, I put the draganddrop control before the legend control in the config.

johnnyblasta:
When open/close the popup menu it's no longer possible to add more layers.
DADAddButton

jokd:
Now it looks like this instead: 
image

asemoller:
Oh, the new icon looks like a file/document icon from the 90's, doesn't it?? 😨 😃 
Personally, I liked the plus better but in origo blue instead of red.

johnnyblasta:
Tested thoose where it failed before and now it works as expected.
Waiting to approve, so more can weigh in on the appearance of the button image.

jokd:
> The reason for not going with the blue plus is because that is used by the layermanager plugin, so using the tools side by side will cause confusion. But I do agree it looks not totally origoish.

The reason for not going with the blue plus is because that is used by the layermanager plugin, so using the tools side by side will cause confusion. But I do agree it looks not totally origoish.

Grammostola:
Seems to work as described. Could the tooltip perhaps be an option? (I was thinking "Add layer from file" or something along those lines).

Not really relevant for this PR:

I tried this control in conjunction with the layermanager:
bild
The different icon makes it clear it's a different action. This can be good. The alternative would be layermanager's blue plus icon but before a combined plus with a context menu there'd be two identical icons (with different tooltips on desktop) if both controls are employed. (If the layers are added to the same group the drag-and-drop of layermanager added layers does not magically transfer to the file-added-layers).

asemoller:
From a user perspective I guess it's the same, a user wants to add a layer to the legend, whether it's from a file or a library doesn't matter to the user. It may be different from a technical perspective, but having two different buttons is not preferable.

How about changing the plugin instead? I guess there'll be more users using a feature in origo core than a plugin... but that doesn't solve the problem with two buttons though...

tonnyandersson:
Yep, two buttons makes it a bit cluttered and confusing, especially since the main goal (add a layer to the map) is the same for both functions.

I think it makes most sense to either change the Layermanager button (like @asemoller suggested) to make it look more "managey", or to add some sort of context menu with options to add layer from file or open Layermanager. I vote for the latter. And it doesn't necessarily have to be a context menu like the "More" menu for layers - I think a modal with two buttons would work just fine.

With a context menu type of thing it would make sense to move the blue + button to the core and handle all the context menu logic there.

Grammostola:
Some stray thoughts while awaiting input from the pr author:

If modal or context menu because it's the same to the user and becomes UI-cluttered elsewise then it makes a certain sense that the added layers would end up in the same group, in which case I the user would expect draggability (what I ambiguously referred to as 'the-drag-and-drop' above) on all of those layers. I think it's a good idea for layers added no matter how added. That might be somewhat beyond the scope of the PR though.

In addition I'm only 74% convinced it's the same to the user as one way requires only an idea of what kind of data to add and the other an actual spatial file.

jokd:
I've done it like this in case of more than one "addLayerButtons", else a click will get instant reward:
image

It will hopefully just require an extra argument from the layermanager plugin.

jokd:
And I suggest making layers in a group draggable is a separate issue.

johnnyblasta:
Tested now with `"searchLayersControl": true` and the add layer button was hidden behind the search input which wasn't the case before.
image

jokd:
> Tested now with `"searchLayersControl": true` and the add layer button was hidden behind the search input which wasn't the case before. image
I'd say it's because of the awesomeplete-black class of the input element. It shouldn't be position: fixed and min-width:100%

johnnyblasta:
> > Tested now with `"searchLayersControl": true` and the add layer button was hidden behind the search input which wasn't the case before. image
> > I'd say it's because of the awesomeplete-black class of the input element. It shouldn't be position: fixed and min-width:100%

That was added so that awesomplete result list wasn't hidden when larger than the legend.

jokd:
> > > Tested now with `"searchLayersControl": true` and the add layer button was hidden behind the search input which wasn't the case before. image
> > > I'd say it's because of the awesomeplete-black class of the input element. It shouldn't be position: fixed and min-width:100%
> 
> That was added so that awesomplete result list wasn't hidden when larger than the legend.

Gotcha. Tried to solve it in another way. Can you have another look?

johnnyblasta (approved):
Excellent, now I can't find any fault with this.

jokd:
@Grammostola Can you test this with your layermanager implementation?

Grammostola:
@jokd Will do so on the morrow!

jokd:
> Seems to function well : ) If it isn't a major hassle maybe they could be aligned, visually, somehow, else I'm happy with it as is.
> 
> kombo_läggtill

Rearranged a bit so there's some default styling for the popup. Should work for both text and buttons.
image

Grammostola:
Lgtm : )

PR #1686, opened 2023-02-10
feature: center marker for externalurl

### LINKED ISSUE #1672: Center marker for external url
When using the external url control it is useful if the crosshair is active to get a more precise position.

When clicking the external url button, bneside activate the sub menu it should add a crosshair in the center of the map, just like the position tool. This obviously won't work when there is just one link since the sub menu won't appear. One idea is to move the center marker handling from the position control and make it global and able to use from any control such as draw and measure for better touch functionality.

It is possible to activate the crosshair in the position control on the bottom left, but that is not as intuitive.

jokd:
Fixes #1672
Adds a centerMarker component to the viewer, easy to use with show/hide methods. Future improvements are to use this along with eg the position control, but they somehow has to respect each other.

### DIFF
--- a/src/components/centermarker.js
+++ b/src/components/centermarker.js
@@ -1,0 +1,30 @@
+import { Component, Icon, dom } from '../ui';
+
+export default function CenterMarker() {
+  let viewer;
+  let markerEl;
+
+  function createMarker() {
+    const markerIcon = Icon({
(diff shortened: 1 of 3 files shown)


johnnyblasta (approved):
LGTM

PR #1689, opened 2023-02-10
fix: drawtool other than default

### LINKED ISSUE #1687: Editor failed to select drawtool
**Describe the bug**
When configured the editor to use other tool than the standard tool for the geometrytype it's not able to select it.

**To Reproduce**
EditorBugLineDraw

**Expected behavior**
It should be able to have the possibility to add lines when the default geometrytype is polygon or vice versa. It works as expected when box is in the drawtools.

johnnyblasta:
Fixes #1687

### DIFF
--- a/src/controls/editor/shapes.js
+++ b/src/controls/editor/shapes.js
@@ -5,6 +5,15 @@ export default (drawType) => {
     box: {
       type: 'Circle',
       geometryFunction: createBox()
+    },
+    Line: {
+      type: 'LineString'
+    },
+    Polygon: {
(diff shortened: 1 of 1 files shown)


jokd (approved):

johnnyblasta:
Tested saving linestring to a polygon layer and it worked fine. The setup was a PostGIS with Geoserver and the geometry was saved as ST_LineString in the database and Geoserver doesn't seem to have a problem serving both the linestring and polygons and they are both displayed in Origo map.

PR #1694, opened 2023-02-24
feature: add draggable option for groups

### LINKED ISSUE #1692: Make layers in a group draggable
I suggest a group setting for making layers in the group draggable, meaning they can be rearranged inside that group to control their draw order. Perhaps most useful for layers added by the user.

jokd:
Fixes #1692 
Set `draggable: true` to make a group sortable by dragging layers. Useful for groups with user added layers.

### DIFF
--- a/origo.js
+++ b/origo.js
@@ -25,6 +25,7 @@ import { renderSvgIcon } from './src/utils/legendmaker';
 import SelectedItem from './src/models/SelectedItem';
 import 'elm-pep';
 import 'pepjs';
+import 'drag-drop-touch';
 import permalink from './src/permalink/permalink';
 
 const Origo = function Origo(configPath, options = {}) {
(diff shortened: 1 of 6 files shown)


Grammostola:
Well done, I think I like the visual effect (lines move)  slightly more than layermanager's (visible dropzones) but honestly they're both pretty great. A (rearranged) layer (draw) order seems to apply to print and post-print. 
There's an error (that doesn't appear to hinder any bit of functionality) in the console when dragging a layer unto the legend area:
```console
group.js?e8dc:201 
        
       Uncaught TypeError: Cannot read properties of null (reading 'parentNode')
    at HTMLLIElement.handleDragEnter (group.js?e8dc:201:1)
handleDragEnter @ group.js?e8dc:201
``` 
(that does not happen with the current master)

jokd:
Should be fixed now (but drag and drop onto the legend have never worked, right?)

Grammostola:
Seems fixed now (no crash when hovering over). Re ever worked to actually drop on the legend: don't know, isn't working in the current master anyway.

jokd:
OK to merge?

Grammostola:
Yes, I think so

PR #1706, opened 2023-03-06
feature: change the type of layertype.js to be layerType instead of t…

### LINKED ISSUE #1705: VSCode syntax highlighting has issues with one file
 VSCode has issues with src/layer/layertype.js in terms of syntax highlighting. This makes it harder than it needs to be to edit this file. (The linter has no issues). See image:

nu_type

efter_type

Use notepad++ or any other editor that doesn't have this syntax highlight issue. (I haven't tried starting Visual Studio)

**Additional context**
(I don't think it's a config issue since it happens to a new empty installation of VSCode though I could be wrong. The proposed solution is a change to layertype.js interface but hasn't produced any artifacts in limited testing and I think it's doable: there seems to be a limited amout of places where layertype.js is used.

Grammostola:
…ype for VSCode syntax highlight compatibility. Aims to fix #1705

### DIFF
--- a/src/layer.js
+++ b/src/layer.js
@@ -1,6 +1,6 @@
 import mapUtils from './maputils';
 import group from './layer/group';
-import type from './layer/layertype';
+import layerType from './layer/layertype';
 
 function onChangeVisible(e) {
   const layer = e.target;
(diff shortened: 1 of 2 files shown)


johnnyblasta (approved):
LGTM

PR #1708, opened 2023-03-06
feature: Add group aggregations and length attribute helper.

### LINKED ISSUE #1662: Display total area and length for a selection
"My users" complains about having to use the measure tool in order to get length or area of an object of set of objects.

_Infowindow_ should display the area and length for all selected objects individually and also as a total for each layer and also a grand total for all layers. It would of course be configurable to display and optional, possibly on layer level.

Use the undocumented {{@area}} notation in an _html_ type attribute on feature level. Does not solve length and sum though.  I have not delved into in detail, but I have not found a way to inject custom functions without rebuilding origo to solve the length issue.

Some databases (arc sde) would provide area and length as proper attributes. Other databases would probably also be capable of adding an area and length column, either by a view or triggers. Still does not solve the aggregation and requires some server side shenannigans. 

**Additional context**
The sum function could possibly be implemented as a more generic aggregation function that aggregates arbitrary attributes of a selection if the measurement is exposed as an attribute.

steff-o:
Fixes #1662.

This PR consists of two parts. The first is adding a length HTML attribute helper function. It works like the undocumented `area` helper function. this is __not__ needed in order to aggregate length or area, it is just to display length and area for individual features. See below for how to configure:

```
"attributes": [
        {
          "name": "fritext"
        },
        {
          "html": "Areal: {{@area(3)}}"
        }
      ],

"attributes": [
        {
          "name": "fritext"
        },
        {
          "html": "Length: {{@length(3)}}"
        }
      ],
```

The second part is the actual aggregation of attributes. It only works when using _SelectionManager (InfoWindow, InfoWindow list view, MultiSelelect plugin)_.  Any (numeric) attribute can be aggregated in addition to area and length.
Aggregation of the configured attributes are displayed at the bottom of the InfoWindow for each layer. It will only display the aggregation if the layer has the attribute or attribute function. 

There is no aggregation between layers. I left that out for now.

See below configuration example for all possible options.
                         
```
"featureinfoOptions": {
    "infowindow": "infowindow",   
    "infowindowOptions": {
      "listLayout": true,
      "groupAggregations": [
        {
          "layer": "linjelager", // Optional. Name of layer to apply to. If omitted or empty applies to all layers
          "function": "sum", // Required. Aggregation function. Only "sum" is supported
          "scalefactor": 0.01, // Optional. Factor to multiply result with. Defaults to 1
          "decimals": 2, // Optional. Number of decimals in result. Defaults to 2
          // Required. Name of attribute to aggregate. If it starts with @ it is a function that is applied to each feature.
          // Supported functions are @area and @length. 
          // If no features in a layer has the attribute or the function can not be applied to the feature type (e.g. using @area on line features)
          // the aggregation is ignored for that layer
          "attribute": "@length",
          "label": "Total length ", // Optional. Text to put before aggregation result. Defaults to function(attribute):
          "suffix": " hektometer" // Optional. Text to put after aggregation result. Default to empty
        },
        {
          "layer": "", // If empty apply to all
          "function": "sum",
          "attribute": "@area",
          "label": "Total yta ", 
          "suffix": " kvadrat" 
        },
        {
          "layer": "", // If empty apply to all
          "function": "sum",
          "attribute": "inten",
          "decimals": 0,
          "scalefactor": 100
        }
      ]
    }
  },
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1062,7 +1062,7 @@
     "arrify": {
       "version": "1.0.1",
       "resolved": "https://registry.npmjs.org/arrify/-/arrify-1.0.1.tgz",
-      "integrity": "sha512-3CYzex9M9FGQjCGMGyi6/31c8GJbgb0qGyrx5HWxPd0aCwh4cB2YjMb2Xf9UuoogrMrlO9cTqnB5rI5GHZTcUA==",
+      "integrity": "sha1-iYUI2iIm84DfkEcoRWhJwVAaSw0=",
       "dev": true
     },
     "asn1": {
@@ -1077,7 +1077,7 @@
     "assert-plus": {
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/assert-plus/-/assert-plus-1.0.0.tgz",
-      "integrity": "sha512-NfJ4UzBCcQGLDlQq7nHxH+tv3kyZ0hHQqF5BO6J7tNJeP5do1llPr8dZ8zHonfhAu0PHAdMkSo+8o0wxg9lZWw==",
+      "integrity": "sha1-8S4PPF13sLHN2RRpQuTpbB5N1SU=",
       "dev": true
     },
     "astral-regex": {
@@ -1098,13 +1098,13 @@
     "async-foreach": {
       "version": "0.1.3",
       "resolved": "https://registry.npmjs.org/async-foreach/-/async-foreach-0.1.3.tgz",
-      "integrity": "sha512-VUeSMD8nEGBWaZK4lizI1sf3yEC7pnAQ/mrI7pC2fBz2s/tq5jWWEngTwaf0Gruu/OoXRGLGg1XFqpYBiGTYJA==",
+      "integrity": "sha1-NhIfhFwFeBct5Bmpfb6x0W7DRUI=",
       "dev": true
     },
     "asynckit": {
       "version": "0.4.0",
       "resolved": "https://registry.npmjs.org/asynckit/-/asynckit-0.4.0.tgz",
-      "integrity": "sha512-Oei9OH4tRh0YqU3GxhX79dM/mwVgvbZJaSNaRk+bshkj0S5cfHcgYakreBjrHwatXKbz+IoIdYLxrKim2MjW0Q==",
+      "integrity": "sha1-x57Zf380y48robyXkLzDZkdLS3k=",
       "dev": true
     },
     "atob": {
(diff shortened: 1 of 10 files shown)


asemoller:
Is it possible to get a picture or a gif of this new feature? Just for a quick look 😄.

steff-o:
By popular demand, here are some piccies:

image
Example showing old infowindow with multiple lines selected.

image
Example with old infowindow showing selections in two layers. (Both layers are named Multipoly, hence the strange looking header)

image
Example showing listview selections in two layers.

tonnyandersson:
Doesn't seem to work with MultiLineString geometries, probably because MultiLineString geometries don't have a getLength function.

steff-o:
@tonnyandersson you are right. Maybe I was a bit too lazy and just ripped off the old attribute helper code that used getArea. 

Good thing you discovered this as getLength and getArea only calculates plane coordinates anyway. Probably should rewrite it to use ol/sphere like measure do.

tonnyandersson:
Lazy for not testing all kinds of geometry types, maybe, but weird that OL has implemented getArea for MultiPolygon but not getLength for MultiLineString.

steff-o:
Updated to calculate area and length the same way as measure tool by using ol/shpere. Result is support for MultiLineString as well as non-projected coordinate systems.  This change applies to both the html attribute helpers and the aggregation.

tonnyandersson:
LGMT. Approval pending, holding off a day or two in case anyone else wants to express their opinion.

jokd:
Looks good!

asemoller:
Nice with pictures! Thanks @steff-o !

Has anyone tested together with the export buttons as they share the same space?

asemoller:
I've tested the function and it's a really nice new feature. I actually have a user case for this already 👍 .

However, I have a few thoughts:

1. Is it possible to use the same units of measurement with predefined breakpoints for @area and @length in the same way that the measurement tool has by default (eg breakpoints for square, hectare, square kilometer for areas and meters, kilometers for length)? I think it would be more useful, especially for the attribute function which currently has no scale factor (or has it?).

2. The placement for the text for summarizing the objects is, in my opinion, too far to the left. I prefer the text to line up either with the layer title or with the object text. Same goes for listLayout which apparently doesn't have anything aligned and that's really annoying but I guess that's another issue.

I'm adding some pictures to illustrate the placements and units.

length_area_function_infowindow_issue


length_area_function_listLayout_issue

steff-o:
Pushed a change that automatically formats area and length to appropriate units. Applies to both attribute helpers and aggregations. 

For aggregations it can be overridden by setting a unit i config:
```javascript
 {
          "layer": "", // If empty apply to all
          "function": "sum",
          "attribute": "@area",
          "label": "Total yta ", 
          "unit": "kvadrat" 
 },
```

This means that the setting `suffix` has been removed (or renamed to unit).

The attribute helpers are always automatic.

Also aligned the aggregations with the attributes.

jokd:
Looks like we can merge this, @tonnyandersson

tonnyandersson:
Yep!

tonnyandersson (approved):

Issue #1617, opened 2022-10-17
Measure tool distance problem


David0099:
**Describe the bug**
Problem with measure tool, not getting the correct distance.

I used the measure tool to measure buildings (using snap) and noticed that I did not receive the same distance-result as in other software we use (QGIS and Topocad)
For example the result on a 20 meter building differed about 5 cm. 

**Screenshots**
![Origo](https://user-images.githubusercontent.com/52694843/196150142-cc7139b8-a6f7-43d5-aae9-cceaaf4e119f.JPG)
Origo

![QGIS](https://user-images.githubusercontent.com/52694843/196150164-920fdb5d-6849-482d-9fd1-b994455d56bc.JPG)
QGIS

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser Chrome

jokd:
I can verify this and here are discussions on the subject:
https://github.com/openlayers/openlayers/issues/9141
https://github.com/openlayers/openlayers/pull/13572

jokd:
> Is there anything we can do in Origo to achieve higher accuracy? If not, perhaps this issue can be closed.

We can implement another library but obviously adds size to the core. Maybe a plugin is the way to go?

tonnyandersson:
Closing this since it's mostly an OL issue. Based on the information found in the links provided by @jokd above, there are various ways to solve this issue but all of them seem to require third party libraries to do the heavy lifting.

For now, if anyone feel the need of desktop GIS levels of accuracy in Origo we suggest you develop a plugin utilizing such libraries.

PR #1714, opened 2023-03-16
feature: draw as core control

### LINKED ISSUE #1610: Adding draw to core
The draw plugin is widely used and should be in the origo core. Adding it won't be many kB extra to the build and it will be the first step in solving #1609.

jokd:
Fixes #1610 
Major rewrite of the draw plugin to make it a core control. Also the measure control has some new functionality.
Will update this post with the changes made and how to configure the control soon.


jokd:
It is now possible to add measures to the drawed objects:
image

And the measure control objects can be modified:
image

Grammostola:
With the given configuration, except `queryable: true` and `showSaveButton: false` the drawn features and their given popup texts transferred via a saved mapstate and were printable. When changing the DPI in the print pane however the shapes change form/size. related to it needing to be a vector-layer as per pr 64 of the draw plugin (that you previously included; it isn't immediately apparent where the layer is coming from here). 

I also experienced troubles with the two line types,
``` 
 function stringToRgba(colorString, opacity) {
    if (colorString.toLowerCase().startsWith('rgba(')) { return colorString; }
```
Thinks that colorString is undefined

Anyway I think the info popup definition of drawn objects is a great idea and I like where this PR is headed

johnnyblasta:
The font-weight is a bit much in my opinion:
image

jokd:
Now I have hopfully fixed the printing issues. The measure labels won't scale though since they are hard coded. But maybe we can live with that? 

Changed the label class:
image

johnnyblasta:
When a buffer is used in the measure control the print throws a error.

jokd:
> When a buffer is used in the measure control the print throws a error.

Pushed a fix for this as well as label scaling

Grammostola:
My prior complaints are now void. I can't seem to make measurements saveStateToServer however.

jokd:
> My prior complaints are now void. I can't seem to make measurements saveStateToServer however.

Forgot about measure in mapstate, and I haven't been able to test against old mapstates but seems to work with newly created.

filleg:
Some really nice improvements. Tested some measurements both in measure and in draw. The labels of measurements seems to be off when changing to a different projection code than web mercator. Verified by exporting the drawing and measuered it in QGIS.

jokd:
> Some really nice improvements. Tested some measurements both in measure and in draw. The labels of measurements seems to be off when changing to a different projection code than web mercator. Verified by exporting the drawing and measuered it in QGIS.

Is it the same as this issue: https://github.com/origo-map/origo/issues/1617 ?

filleg:
No, the difference is a bit bigger. Otherwise I probably wouldn't have noticed :)

image

johnnyblasta:
Looks like it measures correct now when I tested.
Perhaps ready to merge now?

jokd:
@Grammostola @filleg 
Can you have another look?

filleg:
Looks good now.

steff-o:
@johnnyblasta, I don't think it's ready to merge quite yet. I have some issues that I will post soon.

steff-o:
As promised, here are my observations. I have not found anything that actually is blocking a merge, but some things that would be nice to have fixed at some point.

First of all, great job. It has a great look and feel and some really nice features.

__Here are some stuff that I think should be fixed as soon as possible:__
- The exported filename should default to the name of the draw layer. As on now, the name is "export" when saved from menu and layer dialog. When saved from TOC it is "Ritlager".
- Buttons for attributes and delete should be greyed out when no feature is selected
- It is a bit unclear what the style selector changes. The idea to switch between currect object and "default" is good, but I would like a change in title when no feature is selected to indicate that it will be the new "default".
- I would prefer if there was an actual default symbol. Right now the next drawn object will get the same symbol as the last selected.
- I think its a bit unclear that the export button in the menu only exports the active layer. Not sure what to do about it. Export all? Remove it? Maybe its clearer if the export gets the name of the layer?
- The attribute name in the geojson file is "style". Not really anything crucial about it, but it could possibly collide with a user attribute name, especially if it will become possible to import arbitrary files. Maybe it should be named something more unlikely like "__origo_style"
- I don't see the logic in having all other tools in the left menu when draw is the same type of tool as measure and edit where only one can be active at a time. If moving to the left menu the close button can be removed. BTW, the measure tools should also be in the bottom toolbar.
- It is not obvious that the only way to abort a draw tool (get back to select) is to click the same tool again (or draw an object and delete it). Maybe should have a select tool.

__Code review__
As the diff is a couple of thousand of lines of code I have not code reviewed much. But when browsing through the code if saw that the drawhandler dispatches events on behalf of the draw component. Looks a bit backwards to me when the purpose is to communicate with the draw component and possibly makes it harder to follow the flow. In my opinion it would be handled by a direct call to the draw component and let the component dispatch its own events, or having draw component listen to events from the drawhandler.

__Nice to haves__
Here are some features that I would like to have, but is probably best suited as later additions:
- More custom attributes
- Automatically store drawings in localstorage
- Import of arbitrary geojson files for editing. If they have the style-attribute it will be used.

jokd:
Pushed a commit where:
Rewrote drawHandler as a component
Changed export names
Grayed out icons

As for the stylewindow, a reset button can be useful. The idea of not having a default style is to be able to reuse a certain (not default) style without resetting every setting, therefor you inherit last objects style. But a button to reset it makes sense.

The export button maybe is redundant since you also can export from the legend?

I plan to make it configurable whether to have the button i the legend and/or in the left menu. Just like the print control.

steff-o:
@jokd I have to break it to you but the fix for greyed out attributes button does not work the first time draw is activated (it works for delete). All other changes looks good.

johnnyblasta:
@jokd might be nice to have a way in config to set that the draw tool should be active on start.
`"initialState": "active"`
Default value should be not active.

jokd:
Now there's an isActive-setting that defaults to false if you want to activate the control at start.
Also a button to import geojson layer:
image
Supports edits and sharemap as well.

johnnyblasta:
I'm not getting the specified exra draw tools. @jokd 

My config:
```
 {
    "name": "draw",
    "options" : {
      "buttonText": "Rita", 
      "drawTools": { 
         "Polygon": ["freehand", "box"],
         "LineString": ["freehand"]
      },
```

jokd:
> I'm not getting the specified exra draw tools. @jokd
> 
> My config:
> 
> ```
>  {
>     "name": "draw",
>     "options" : {
>       "buttonText": "Rita", 
>       "drawTools": { 
>          "Polygon": ["freehand", "box"],
>          "LineString": ["freehand"]
>       },
> ```

Now it should work. I moved the event handling and forgot about the extra tools.

jokd:
Anyone dare to approve this or are there more stuff to fix?

asemoller:
I've some inputs!

1. I'm not sure if the icon for attributes should be the same as for the mapmenu, it's a bit confusing. What about [textsms](https://fonts.google.com/icons?selected=Material%20Icons%3Atextsms%3A), [textfield](https://fonts.google.com/icons?selected=Material%20Icons%3Atext_fields%3A) or something similar?
2. When you add a new layer, it becomes active and gets the default name. It's not obvious that the layer on top is the new one. What about adding a number suffix to the new layer (Ritlager (1)) or just not making the new layer active?

I really like this new feature!

asemoller:
> Auto activating a new layer makes sense to me but I guess it could be done the other way as well.

I agree, it makes sense! A suffix makes it more obvious that a new layer has been created and I think that's enough. Well done! (haven't tried it yet but it looks good.) 😃

Grammostola:
Drawn features incl measurement labels get thinner anew when -> 300dpi in the print preview (measure lines from the measure tool and their labels are ok)

jokd:
@Grammostola Scaling should work now

Grammostola:
Cheers for the quick attention, drawn features now rescale properly when 300dpi in the print preview, however measure lines do not and upon exiting the print preview they turn into red non-dashed lines with no label.

jokd:
> Cheers for the quick attention, drawn features now rescale properly when 300dpi in the print preview, however measure lines do not and upon exiting the print preview they turn into red non-dashed lines with no label.

Yeah there was problem with default style as well. The resize function can't handle style functions so the function itself must do the scaling.

jokd:
...thinking a bit more, perhaps the default style function should be the same as for draw. But that's another issue.

jokd:
> Looks fixed : ) Depending on how things progress I'll endeavour to test some more tomorrow but currently I've no further things to note.

Please do, would be great to merge before release

jokd:
@Grammostola OK to merge this? Keen to make the release and this is the last piece in the puzzle

Grammostola:
I detected no other anomalies when I tested yesterday but haven't had the time to check today (we are in the process of a software update). As far as I can tell at the moment it looks good but further testing wouldn't hurt. I don't expect to be able to do that before lunch however. (I still have some documentation stuff to do too which would rhyme well with the release)

johnnyblasta:
@jokd I'll have a wack at testing it now to see if I find anything more.

johnnyblasta (approved):
This gets a big thumbs up from me, good work!
Happy easter!

Grammostola:
Indeed, I've tested some more (primarily around printing and sharing) which seemed to work : ) Well done.
(One peripheral thing I noted is that drawn features from the current plugin version do not seem to retain their style when opened with this version. There are various ways around that, including communication, as it may not feasibly be within the scope of this pr). Happy easter holidays!

MattiasSp:
What a fantastic job on this PR @jokd and big applause to everyone giving constructive feedback 👏👏👏

This feature alone warrants a swift upgrade to 2.7! Cheers! 😄

PR #1715, opened 2023-03-17
fix: decouple default WMS server style and preselected style for the …

### LINKED ISSUE #1642: wmsStyle not fully compatible with stylePicker for WMS layers
The stylePicker is currently available for vector layers and WMS layers (documentation pending).

Specifying that a WMS layer should initiate with an alternative style can be done via the `wmsStyle` property of a `style` object as per the documentation. It should be possible to also specify a stylePicker for such a WMS layer without issues. This is not the case currently. 

The `default` style is always pre-selected in the stylePicker and combining `default` with a styleName does not work quite as expected.

**To Reproduce**
Specify a WMS layer to start with an alternative style and also a stylePicker which includes the start style

```json
    "style": "papperskorgaraltstil",
    "stylePicker": [
      { "title": "Papperskorgar standardvisning", "default": true  }, 
      { "title": "Tömningsintervall", "style": "papperskorgar_tomning", "hasThemeLegend": true }    
    ]

///
    "papperskorgaraltstil":
    [[{
      "icon": {
        "src": "https://mapserver/geoserver/workspace/wms?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetLegendGraphic&FORMAT=image%2Fpng&LAYER=sokvy_park_papperskorgar&SCALE=2132.8571428571427&STYLE=papperskorgar_tomning"
       },
       "extendedLegend": true,
      "wmsStyle": "etuna_park_papperskorgar_tomning"
    }]],
```

**Expected behavior**
The given `default` style is still treated as the default layer style on the WMS server and in Origo but the stylePicker is preset to the alternative style specified by `wmsStyle`

**Current behavior**
The given `default` style is  treated as the default layer style on the WMS server and in Origo and is also the preselected style in the stylePicker. The stylePicker appears partially non-functional (the legend graphic does not change however the layer style in the map does)

**Additional context**
```json
    "stylePicker": [
      { "title": "Papperskorgar standardvisning", "style": "etuna_park_papperskorgar"  }, 
      { "title": "Tömningsintervall", "style": "papperskorgar_tomning", "hasThemeLegend": true, "default": true }
```
appears to work until the map is shared and opened where the layerPicker does not work as expected and the configuration is perhaps a bit unintuitive.

Grammostola:
…stylepicker in order to accomodate preselected alternative WMS server styles when a stylepicker is employed for WMS layers.
Aims to fix #1642 by removing the `default` property from the stylePicker for WMS layers and adding the `defaultWMSServerStyle` property as well as the `initialStyle` property for what style is the WMS server's default style (that need not be named in WMS requests) and what style to employ when loading the map (which is then preselected in the stylePicker dropdown).

It isn't meant to alter the function of vector layer stylePickers. It seeks establish the following priority between a `style` prop, the `hasThemeLegend` and `legendParams` props and the `stylePicker` prop:
1. `stylePicker` (then ignore the other three if present)
2. `style` (then ignore the final two)
3. WMS autostyle which considers `hasThemeLegend` and `legendParams`

This means that when a stylePicker is defined for a WMS layer everything regarding that layer's styles can and must be defined in the stylePicker. 

A sample stylePicker:
```json
      "stylePicker": [
        { "title": "CSS", "style": "Bakgrund_bke_fk_mo_0484_css" },
        { "title": "Mörk", "style": "Bakgrund_bkm_fk_mo_0484", "hasThemeLegend": true,
          "legendParams": {
            "legend_options": "dpi:600"
          }
       },
        { "title": "Standard", "defaultWMSServerStyle": true, "hasThemeLegend": false, "initialStyle": true,
          "legendParams": {
            "scale": 100
            } 
        }
      ]
```

Currently there's no error handling, the schema needs to be followed otherwise expect errors. I can make it more robust if there's a preference for that but there are a multitude of config possibilities here and what is expected behaviour if incorrect input is not always easy to ascertain.
If the given schema is adhered to it should never break, though since the config options are of said multitude there's the saying of forest roads 
> Hinder och fara kan förekomma

 so please do test a few creative scenarios, including without a stylePicker to make sure WMS autostyle never goes astray, as well as a `style` prop that both employs an alternative `wmsStyle` and does not, including sharing and printing.

(See the reasoning in the issue about breaking changes, i.e WMS wasn't in the stylePicker in the last release)

### DIFF
--- a/src/controls/legend/overlayproperties.js
+++ b/src/controls/legend/overlayproperties.js
@@ -65,7 +65,7 @@ const OverlayProperties = function OverlayProperties(options = {}) {
   function getStyleDisplayName(styleName) {
     let altStyle = stylePicker.find(s => s.style === styleName);
     if (!altStyle) {
-      altStyle = stylePicker.find(s => s.default);
+      altStyle = stylePicker.find(s => s.defaultWMSServerStyle);
     }
     return (altStyle && altStyle.title) || styleName;
   }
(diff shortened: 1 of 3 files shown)


Grammostola:
A style can be both `initialStyle` and `defaultWMSServerStyle`, one of them, or neither. I was thinking of employin the internal ref "defaultStyle" for WMS styles too but `defaultStyle` and `defaultWMSServerStyle` do not go too well together.

jokd:
Works well and I just have one idea. In case initialStyle is omitted, pick one (the first?) to use as initialStyle instead of crashing the map.

Grammostola:
That's a good suggestion and I've tried to make it so. (It chafes a little to repeat the logic of patching an incomplete stylePicker in viewer.js and layer/wms.js. I could alter wms.js some to expose the method. Or we can leave it as it is, it's not one hundred repeated lines)

jokd:
LGTM

jokd (approved):

PR #1717, opened 2023-03-20
feature: mapState as map parameter

### LINKED ISSUE #1716: Init Origo with a mapState
It is useful to be able to add a mapState-parameter in the init params instead of adding a hash to the URL.

```
Configured like this:
  var origo = Origo('index.json', {
    mapState: {
    "layers": [
        "osm/v/1/s/0/o/100",
        "origo-mask/v/1/s/0/o/25",
        "origo-cities/v/1/s/0/o/100"
    ],
    "center": "1810000,8390000",
    "zoom": "8.7",
    "legend": "expanded",
    "map": "index"
    }
  });
```

Use the available functionlaity with hash or mapState but they both have their drawbacks for this application

jokd:
Fixes #1716 
Makes it possible to add a mapState when initializing a map. A way to alter the default config eg when customizing a map based on the user

### DIFF
--- a/src/loadresources.js
+++ b/src/loadresources.js
@@ -1,4 +1,5 @@
 import permalink from './permalink/permalink';
+import permalinkParser from './permalink/permalinkparser';
 import getUrl from './utils/geturl';
 import isUrl from './utils/isurl';
 import trimUrl from './utils/trimurl';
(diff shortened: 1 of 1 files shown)


johnnyblasta (approved):
LGTM

PR #1727, opened 2023-03-24
feature: defaultLayerParams in config

### LINKED ISSUE #1725: Default layer params
When setting up a map with lots of layers theres a lot of layer settings that has to be repeated for every layer. Moving them to a defaultParams object could shrink the config size as well as make the config easier to read.

Something like this:
```
defaultLayerParams:{
  visible: false,
  queryable: true,
  searchable: true,
  source: 'geoserver'
}
```

jokd:
Fixes #1725

### DIFF
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -43,6 +43,7 @@ const Viewer = function Viewer(targetOption, options = {}) {
     zoom: zoomOption = 0,
     resolutions = null,
     layers: layerOptions = [],
+    defaultLayerParams = {},
     map: mapName,
     params: urlParams = {},
     proj4Defs,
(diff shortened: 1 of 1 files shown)


jokd:
Changed behaviour a bit so that there's no default setting, you must explicitly add a param to each layer. On the other hand you can have multiple layerParams to choose from (though not on the same layer)

Use it like this:
```
  "layerParams": {
    "default": {
      "group": "root",
      "visible": true
    }
  },
  "layers": [
    {
      "name": "origo-cities",
      "title": "Origokommuner",
      "style": "origo-logo",
      "type": "GEOJSON",
      "attributes": [
        {
          "name": "name"
        }
      ],
      "visible": false,
      "layerParam": "default"
    },
```

Layer setting will override the layerParam so in this case the layer will be visible: false

johnnyblasta (approved):
LGTM

Issue #1726, opened 2023-03-24
Add CQL filter on WMS layer


johnnyblasta:
**Is your feature request related to a problem? Please describe.**
When having a WMS layer with for example multiple owners and maintainers it is cumbersome to setup a lot of styles to handle all of the possible variations. So it would be nice to add a CQL filter directly to a layer and use one style for all of the layers and let the filter handle all the possible different combinations.

**Describe the solution you'd like**
Add the filter parameter, which is possible to use on WFS layers,  to also be used on WMS layers.

PR #1728, opened 2023-03-24
feature: filter on wms layer

johnnyblasta:
Example add on layer in config `"filter": "owner = 'me'"`

Closes: #1726

### DIFF
--- a/src/layer/wms.js
+++ b/src/layer/wms.js
@@ -17,7 +17,8 @@ function createTileSource(options) {
       TILED: true,
       VERSION: options.version,
       FORMAT: options.format,
-      STYLES: options.style
+      STYLES: options.style,
+      CQL_FILTER: options.filter
     }
   }));
 }
@@ -32,7 +33,8 @@ function createImageSource(options) {
       LAYERS: options.id,
       VERSION: options.version,
       FORMAT: options.format,
-      STYLES: options.style
+      STYLES: options.style,
+      CQL_FILTER: options.filter
     }
   }));
 }
@@ -103,6 +105,7 @@ const wms = function wms(layerOptions, viewer) {
   sourceOptions.crossOrigin = wmsOptions.crossOrigin ? wmsOptions.crossOrigin : sourceOptions.crossOrigin;
   sourceOptions.projection = viewer.getProjection();
   sourceOptions.id = wmsOptions.id;
+  sourceOptions.filter = wmsOptions.filter;
   sourceOptions.format = wmsOptions.format ? wmsOptions.format : sourceOptions.format;
   const styleSettings = viewer.getStyle(wmsOptions.styleName);
   const wmsStyleObject = styleSettings ? styleSettings[0].find(s => s.wmsStyle) : undefined;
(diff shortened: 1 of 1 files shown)


jokd (approved):

jokd:
Looks good. I suppose sending in an undefined CQL_FILTER-param won't mess anything up.

johnnyblasta:
> Looks good. I suppose sending in an undefined CQL_FILTER-param won't mess anything up.

If you don't specify a filter in config and sends a undefined CQL_FILTER Openlayers will handle it and not set a CQL_FILTER in the call to WMS server thereby it works as unfiltered.

steff-o:
I have not had the time to check this thoroughly yet, but I think there can be some caveats here. AFAIK the cql_filter is a vendor specific parameter for Geoserver. If Qgisserver also has a CQL_filter parameter it should also be supported if it has a different name. The format of the filter may also differ, but that is no problem as the configuration will have to be adapted to the actual server anyway.

Also care must be taken when calling fetureinfo. I think the code uses the OL function _getFeatureInfoUrl_, which should copy all params from the WMS source. For geoserver this is not a problem as CQL_FILTER is also a valid parameter for the getFeatureInfo endpoint. But I think this must be verified.

johnnyblasta:
Added support for QGIS Server which uses EXP_FILTER param instead of CQL_FILTER.
Haven't tested on QGIS Server.

steff-o:
Another radical solution would be to just assign all incoming properties. In that way we support any vendor specific stuff, past, present and future but also loses all control.

johnnyblasta:
Been reading more on QGISServer and it seems EXP_FILTER is only for WFS, might work with FILTER for WMS though.
Perhaps should test on QGISServer before merging.

johnnyblasta:
> Another radical solution would be to just assign all incoming properties. In that way we support any vendor specific stuff, past, present and future but also loses all control.

QGIS Server seems to support quite a few extra params and counting the other systems extra params, so perhaps something to think about.

filleg:
FILTER should work. The difference is that QGIS server wants layername included inside the FILTER parameter.

johnnyblasta:
Changed to setting a params object which will be added to the calls.
More flexible, but demands more of the user knowing what to set.

Example filtering on Geoserver:
```
"sourceParams": {
            "CQL_FILTER": "pg_detaljtyp = 'parkering kvartersmark'"
}
```

Example filtering on QGIS Server (Works on Geoserver also):
```
"sourceParams": {
            "FILTER": "mylayer1:'col1';mylayer1,mylayer2:'col2' = 'blabla''"
}
```

Example setting multiple params on QGIS Server:
```
"sourceParams": {
            "DPI": "300",
            "BGCOLOR": "0x00FF00"
}
```

filleg:
Tested the changes on qgis server and works with a kind of workaround I guess. The column value needs to be enclosed by double quotation marks which is a bit tricky in json. Solved it by using \" instead. Might not be obvious to users/administrators.

jokd:
Then there isn't much to do but to merge this I'd say

PR #1730, opened 2023-03-31
fix: update zoom level when scroll zooming

### LINKED ISSUE #1719: Bug when wheel zooming the print preview
**Describe the bug**
When using the mouse wheel zoom in the print preview and changing the dpi or rotation, the zoom level will reset to the inital value.
Probably due to the scale value not updating when wheel zooming

jokd:
Fixes #1719

### DIFF
--- a/src/controls/print/print-component.js
+++ b/src/controls/print/print-component.js
@@ -79,12 +79,14 @@ const PrintComponent = function PrintComponent(options = {}) {
   let descriptionAlign = `text-align-${descriptionAlignment}`;
   let viewerMapTarget;
   let today = new Date(Date.now());
+  let initZoom;
   let printScale = 0;
   let widthImage = 0;
   let heightImage = 0;
   const originalResolutions = viewer.getResolutions().map(item => item);
   const originalGrids = new Map();
   const deviceOnIos = isOnIos();
+  const view = map.getView();
 
   // eslint-disable-next-line no-underscore-dangle
   const olPixelRatio = map.pixelRatio_;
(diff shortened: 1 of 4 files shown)


Grammostola:
Looks good overall, one smaller thing is when used with `constrainResolution` on the map 1: 50 000 becomes 1:50 010 in the scale selector in the print module. Might
```js
    if (differens !== 0) {
      scaleValue = Math.round(scaleValue / 100) * 100;
```
work (to attempt to round the scale to the nearest hundred)?

jokd:
I bet it is the roundScale-function that does it but I didn't want to alter it. I saw this weird behaviour as well 1:2501 will be rounded to 1:2510
But I can push a change for you to test.

Grammostola:
Yes indeed, that function. Please push and I shall test

Grammostola:
Works well here (with our local gridset with epsg:3010) with constrainResolution off or on (and no specific print scales given); sensible scales in the dropdown when mousezooming and no sudden view change when changing the dpi after.

jokd:
That's strange but I think we can live with that

Grammostola:
I agree, could round to nearest fifty but I've never seen 1:250 in a webmap and am not precisely sure how large a difference can become so lgtm.

jokd:
> I agree, could round to nearest fifty but I've never seen 1:250 in a webmap and am not precisely sure how large a difference can become so lgtm.

Should be closest ten so it's just odd. 
OK to merge?

Grammostola (approved):

PR #1735, opened 2023-04-05
fix: only activate buttons on first level when expanding

### LINKED ISSUE #1734: Not visible subgroups in legend take focus
**Describe the bug**
When tabbing through a map with several levels of subgroups in the legend. A subgroup which isn't visible/expanded will take focus 

**Expected behavior**
Not visible groups and layers shouldn't take focus.

johnnyblasta:
Fixes #1734


jokd (approved):
Looks good

PR #1743, opened 2023-04-19
Print WMS fix

### LINKED ISSUE #1740: Problem with WFS in print
In Origo map I have WMS and WFS layers added. All layers are set in EPSG:3010 which is defined in index.json.
All layers are fetched from Geoserver.

image

When printing from the Origo menu, the WFS layers are pushed aside, while the WMS stays put.

image

Lundqan:
In addition to #1741 to fix #1740 .
Refreshes source on Print close for WMS layers.


steff-o:
@Lundqan: Is this still relevant? According to the original issue there has been reports that the problem can not longer be reproduced.

tonnyandersson:
@Lundqan Is this still an issue?

PR #1748, opened 2023-04-24
feature: map marker

### LINKED ISSUE #1747: Marker as point symbol
As a complement to the regular shapes when styling points it's useful to also have an ordinary map marker. 

image

jokd:
Fixes #1747


johnnyblasta (commented):
Works good, but I think the popup covers to much of the marker when open.
Could it be offset a bit or is the out of scope for this issue since it's how the popup works?

jokd:
Haven't actually tested with the popup but offsetting it would require some adjustments that perhaps can be addressed in another issue.

johnnyblasta (approved):
LGTM

Issue #1228, opened 2021-04-14
Search hits could have a choice of infowindow: type


Grammostola:
**Is your feature request related to a problem? Please describe.**
Search hits* today display an info popup 'overlay' 

**Describe the solution you'd like**
Search hit infos could be displayed in the infowindow: that the map is set to

**Describe alternatives you've considered**
Search hit infos could be displayed according to a parameter for the search control in index.json

**Additional context**
We are currently working with internal maps employing infowindow:infowindow and have received feedback noting the inconsistency between overlay when clicking search hits and iw:iw when infoclicking/selecting

*employing the standard search control

jokd:
@Grammostola Will this be fixed this week or the next?

Grammostola:
@jokd Last sunday?
Is it something that has seen recent interest?

jokd:
Sunday’s fine
Jokes aside, I aim to have a go at it if you won’t. Makes sense to have this feature.

PR #1751, opened 2023-04-25
feature: show search results in infowindow

jokd:
Fixes https://github.com/origo-map/origo/issues/1413
Replaces #1436 
Adds a infowindow-component which can be used to create a floating infowindow or aligned to the left.
Search control is configured ike this:
```
{
"name": "search",
"options": {
...
"searchlistOptions":{
  "placement": "floating", //can be search, left or floating, activated by hitting enter (defaults to nothing)
  "export": true,
  "exportUrl": "origo-server-url/excelcreator",
  "exportButtontext": "Exportera lista",
  "exportFilename": "export.xlsx"
},
"autocompletePlacement": "search", //can be search, left or floating (defaults to search)
}
}
```


tonnyandersson:
@jokd Umm, yeah... If you could just resolve the current conflicts, that'd be great.

jokd:
I fixed the conflicts. But testing it I have a couple of comments myself so I will try to:
Make the export button a real button, and
Put it in a footer so you don't have to scroll down to the bottom to make an export

jokd:
Now the export buttons should lok similar to the infowindow export buttons:

```
  "placement": "left", 
  "roundButton": false
```
image

```
  "placement": "floating",
  "roundButton": true
```
image

Example config:
```
 {
 "name": "search",
 "options": {
  "url": "//origoserver/search",
  "searchAttribute": "NAMN",
  "layerNameAttribute": "TYPE",
  "idAttribute": "GID",
  "geometryAttribute": "GEOM",
  "hintText": "Sök adress eller platser...",
  "includeSearchableLayers": true,
  "autocompletePlacement": "left" //can be search, left or floating (defaults to search),
  "searchlistOptions":{
    "placement": "left", //can be left or floating, activated by hitting enter (defaults to nothing)
    "export": true, // default is false
    "exportUrl": "//origoserver/excelcreator",
    "exportButtonText": "Exportera lista", // default is "Export"
    "exportFilename": "export.xlsx", // default is export.xlsx
    "roundButton": false, // default is false,
    "title": "Sökresultat" // Title for the searchlist, defaults to 'Sökresultat för "{{value}}"', {{value}} will be replaced with the search input
  }
 }
}
```

jokd:
I will also add an option to make a selection of the search result for a layer.

filleg:
Nice additions. Not really a bug, but found something that can be a bit confusing for the user. If you activate the searchlist-component after doing a search you can still access the previous search in the searchbar which won't work.

Probably outside the scope of this PR but if you use the regular search for autocomplePlacement your hits will be with the regular popup. While the searchlist will use infowindow. Would be more coherent to be able to set the regular search to infowindow as well. Covered here I guess #1228

jokd:
Added these options:

```
      "makeSelectionButton": true, //default is false
      "makeSelectionButtonText": "Använd som urval"
```
Which will turn a search result into a selection

jokd:
> Nice additions. Not really a bug, but found something that can be a bit confusing for the user. If you activate the searchlist-component after doing a search you can still access the previous search in the searchbar which won't work.
> 
> Probably outside the scope of this PR but if you use the regular search for autocomplePlacement your hits will be with the regular popup. While the searchlist will use infowindow. Would be more coherent to be able to set the regular search to infowindow as well. Covered here I guess #1228

Thats the way it has always been right?

jokd:
I think I'm done with this, anyone wanna have a look?

jokd:
@tonnyandersson Wanna have another look on this? Would be nice to merge in case it's OK.

tonnyandersson:
@jokd Never had a look at this in the first place. My earlier comment was just about the conflicts.

I don't have a lot of configs including the Search control set up, so it might take some time for me to configure a decent test case. If anyone else has got something ready to roll. please feel free to take this beauty out for a ride around the block.

jokd:
> Nice additions. Not really a bug, but found something that can be a bit confusing for the user. If you activate the searchlist-component after doing a search you can still access the previous search in the searchbar which won't work.
> 

@filleg I addressed this issue in #1817, hopefully it will work well along with this PR (I haven't tested it)

filleg:
I tried the changes and seems to work fine. Can't confirm all parts since it is not fully compatible with QGIS server. But that is more related to getfeature.js and #1139 
Might be good if a geoserver user tests as well.

steff-o:
Tested using GeoServer. Most stuff seems to work and no errors are given in the console. However, I have some comments:
1. If the _placement_ parameter is missing, no searchList is displayed. Maybe intentional but must be documented
2. The relation between _autocompletePlacement_ and _placement_ is not obvious. Maybe a really good documentation will clearify that. In the end it is apparently possible to configure search to use the searchList as one and only display of results or use the old awesomeplete as usual and searchList when hitting enter. 
3. It only works for Option 1 and 2 in the search documentation, which means that _idAttribute_ and _geometryAttribute_ must be specified and also _layerName_ or _layerNameAttribute_. Possibly intentional, but it is possible to make the searchList work without _idAttribute_, but when clicking a feature or pressing the select button all features in the layer are selected.
4. I don't see the point with the export button when infowindow has better export possibilities. But it's there and seems to work (haven't actually tested getting a response, as I don't have a report server) so it might as well stay.
5. If you have typed more than minimum characters and has a result and backspaces to fewer than the minimum search characters the searchlist remains  at the last search. The awesomplete is emptied as expected.
6. The _showFeatureInfo_ method in _featureinfo.js_ has no documentation for new newly added argument _opts_. Not the end of the world, but since it is an api function and the other argument is documented opts should also be documented.
7. After pressing enter the search input field is emptied. Clicking the field to type something new, the old search result appears in the awesomplete list. Clicking one of the old results throws an error. Most likely this is the same error as when manually emptying the search input and clicking it.

Other than that there are some other issues already mentioned by others or outside the scope of this PR:
1. getFeature has issues with Qgis server which affects zoom to.
2. Clicking a search result in the awesomeplete list does not obey the infoWindow setting and always displays the result in overlay. searchList on the other hand obeys infoWindow setting which makes the mixed mode with both awesomplete and searchList a bit confusing.

src/utils.js
```diff
@@ -61,6 +61,18 @@ export default {
   makeElementDraggable(el) {
     const touchMode = 'ontouchstart' in document.documentElement;
     const elmnt = el;
+    let draggableEl;
+    if (document.getElementById(`${elmnt.id}-draggable`)) {
+      /* if present, the header is where you move the DIV from: */
+      draggableEl = document.getElementById(`${elmnt}-draggable`);
```
steff-o, line 67: Test is `elmnt.id`, assignment is `elmnt`. Is that correct

steff-o: By the way, I don't think that the `if (document.getElementById(`${elmnt.id}-draggable`))` clause is ever used. It possibly could be from infowindow and infowindow_list to remoge redundant code, but it is not.

jokd:
> Tested using GeoServer. Most stuff seems to work and no errors are given in the console. However, I have some comments:
> 
> 1. If the _placement_ parameter is missing, no searchList is displayed. Maybe intentional but must be documented

We could use searchList: true as an option to activate the list, would that be better?

> 2. The relation between _autocompletePlacement_ and _placement_ is not obvious. Maybe a really good documentation will clearify that. In the end it is apparently possible to configure search to use the searchList as one and only display of results or use the old awesomeplete as usual and searchList when hitting enter.

placement is a searchlistOptions-option while autocompletePlacement is a search-option. Will make it clear in the documentation

> 3. It only works for Option 1 and 2 in the search documentation, which means that _idAttribute_ and _geometryAttribute_ must be specified and also _layerName_ or _layerNameAttribute_. Possibly intentional, but it is possible to make the searchList work without _idAttribute_, but when clicking a feature or pressing the select button all features in the layer are selected.

This can be looked into when issue #1420 is solved

> 4. I don't see the point with the export button when infowindow has better export possibilities. But it's there and seems to work (haven't actually tested getting a response, as I don't have a report server) so it might as well stay.

They do export different stuff and I think that export all of the search results can be useful instead of making a selection and export it layer by layer

> 5. If you have typed more than minimum characters and has a result and backspaces to fewer than the minimum search characters the searchlist remains  at the last search. The awesomplete is emptied as expected.

Will push a fix for this

> 6. The _showFeatureInfo_ method in _featureinfo.js_ has no documentation for new newly added argument _opts_. Not the end of the world, but since it is an api function and the other argument is documented opts should also be documented.

Will push a fix for this

> 7. After pressing enter the search input field is emptied. Clicking the field to type something new, the old search result appears in the awesomplete list. Clicking one of the old results throws an error. Most likely this is the same error as when manually emptying the search input and clicking it.

This is addressed in #1816

src/utils.js
jokd: Yeah, that part is unnecessary and probably not functioning the way it is written.

steff-o:
LGTM.

PR #1756, opened 2023-04-27
feature: post method for wms and wfs

### LINKED ISSUE #1754: getFeature as POST instead of GET
I have an application where the user is filtering the layer with cql-filter and sometimes hit the roof for GET request length. This can be solved by using POST instead so an option to set the request type can be useful.

Set the request method for a source/layer

Filter the features in the client. Won't work for WMS though.

jokd:
Fixes #1754 
Set requestMethod: 'post' for either layer or source to make POST requests instead of GET


jokd:
Found an issue with wms services so this is set to draft

johnnyblasta (approved):
Tested with both WMS and WFS, seems to work fine.
LGTM

PR #1759, opened 2023-04-28
feature: make prompt configurable for removable layers

### LINKED ISSUE #1758: Promptless removal for added layers
**Issue at hand**
We have user-tested 2.7+ and  #1675 received the reaction that the prompt isn't strictly necessary for us since added layers can just be re-added. We use the Layermanager and the same goes for layers added via the drag-and-drop-control. (We don't use the removable property on predefined layers)

That Origo upon remove-action on a removable layer looks for a `promptlessRemoval` layer prop that is `true` and if that isn't found then show the prompt.
The Drag-and-drop control could receive the `promptlessRemoval` option that would default to `true` (because if a layer is drag and dropped to a map that map cannot be shared with the added layer so it should reasonably be re-addable. If for some reason drag-and-drop layers won't be then opt-out via `promptlessRemoval: false`)

Layermanager could act in a similar way as drag-and-drop

Grammostola:
Seeks to resolve #1758 . Drag-and-drop receives the `promptlessRemoval` option and makes it true by default, so it's opt-out.
Layers can have this property and then not cause a prompt to verify when being removed.

### DIFF
--- a/src/controls/draganddrop.js
+++ b/src/controls/draganddrop.js
@@ -71,6 +71,7 @@ const DragAndDrop = function DragAndDrop(options = {}) {
       const groupName = options.groupName || 'egna-lager';
       const groupTitle = options.groupTitle || 'Egna lager';
       const draggable = options.draggable || true;
+      const promptlessRemoval = options.promptlessRemoval !== false;
       const styleByAttribute = options.styleByAttribute || false;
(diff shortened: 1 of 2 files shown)


jokd (approved):

jokd:
LGTM

PR #1761, opened 2023-05-02
feature: add overlayOptions from the origostyle-object

### LINKED ISSUE #1760: Offseting popup
In some cases you want to offset the popup, eg as discussed here #1748 

A setting on style level or as a featureinfoOption to move the popup from its anchor point.
chrome-capture-2023-4-2

jokd:
Fixes #1760
When setting origostyle.overlayOptions.offset for a feature it will offset the popup overlay.
Use with caution because when having multiple features in the carousel it will only check the property of the first feature.

Eg:
image

Can be used like this:
```
const viewer = origo.api();

const layerOptions = {
group: 'root',
name: 'markerLayer',
title: 'Markers',
zIndex: 6,
styleByAttribute: true,
queryable: true,
removable: true,
visible: true,
source: 'none',
type: 'FEATURE',
attributes: [
{
html: '{{content}}'
}
],
style: viewer.getStylewindow().getStyleFunction
};

const layer = viewer.addLayer(layerOptions);
const coords = [1575114, 8396023];
const newFeature = viewer.getMapUtils().createPointFeature(coords, null);
newFeature.set('origostyle', {pointType: 'marker', overlayOptions: { offset:[0, -35] }});
newFeature.set('content', 'This markers content!');
layer.getSource().addFeature(newFeature);
```

### DIFF
--- a/scss/_popup.scss
+++ b/scss/_popup.scss
@@ -13,11 +13,9 @@
   border: 1px solid #ccc;
   border: 1px solid rgba(0, 0, 0, 0.2);
   border-radius: 6px;
-  bottom: 12px;
   box-shadow: 0 5px 10px rgba(0, 0, 0, 0.2);
   box-shadow: 0 5px 10px rgba(0, 0, 0, 0.2);
   max-height: 70vh;
-  padding: 1px;
   position: absolute;
   text-align: left;
   white-space: normal;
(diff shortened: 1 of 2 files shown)


johnnyblasta:
Tested and works with the API, but wouldn't it be cool to be able to set it in config on the style also?

jokd:
> Tested and works with the API, but wouldn't it be cool to be able to set it in config on the style also?

Added possibility to make this in the style object as well:
```
    "origo-logo": [
      [
        {
          "overlayOptions": {
            "offset":[15, 15],
            "autoPan": false,
            "positioning": "center-right"
          },
          "label": "Origokommuner",
.....

All Openlayers overlay options (https://openlayers.org/en/latest/apidoc/module-ol_Overlay-Overlay.html) are available however probably only positioning, autoPan and offset are useful.
Added styling to support bottom-center, center-left, top-center and center-right positioning
```

johnnyblasta (approved):
LGTM

PR #1765, opened 2023-05-09
fix: make featureStyles apply to dragdrop layers by partially mimicin…

### LINKED ISSUE #1764: featureStyles for drag-and-drop layers have no visible effect
If a set of Origo style styles are defined under the `featureStyles` prop for the drag and drop control these styles are not adhered to when matching geometry type layers (without their own style) are dragged and dropped unto the map
**To Reproduce**
1.  Have a drag and drop control defined something like
```json
    {
      "name": "draganddrop",
      "options": {
        "groupName": "drag-and-drop-layers",
        "groupTitle": "Insläppta lager",
        "featureStyles":  {
          "Point": [{
            "circle": {
              "radius": 8,
              "stroke": {
                "color": [0, 0, 90, 1],
                "width": 2
              },
              "fill": {
                "color": [0, 0, 90, 1]
              }
            }
          }],
          "LineString": [{
            "stroke": {
              "color": [0, 0, 0, 1],
              "width": 5
            }
          },
          {
            "stroke": {
              "color": [0, 0, 0, 1],
              "width": 3
            }
          }],
          "MultiLineString": [{
            "stroke": {
              "color": [0, 0, 0, 1],
              "width": 5
            }
          },
          {
            "stroke": {
              "color": [0, 0, 0, 1],
              "width": 3
            }
          }],
          "Polygon": [{
            "stroke": {
              "color": [0, 0, 0, 1],
                "width": 5
              }
            },
            {
              "fill": {
                "color": [0, 0, 0, 1]
              }
            }
          ],
          "MultiPolygon": [{
            "stroke": {
              "color": [0, 0, 0, 1],
                "width": 5
              }
            },
            {
              "fill": {
                "color": [0, 0, 0, 1]
              }
            }
          ]
        }
      }
    },

```
2. Drag a polygon or multipolygon geojson into the map 

**Expected behavior**
A polygon as black as midnight in January in the far reaches of the northern hemisphere

**Current behaviour**
An Origo-blue polygon 

**Additional context**
- The drag and drop control sends a style param with an array of two style functions to `viewer.addLayer()`. (These appear to describe something black) . Somewhere after that the only thing applied visibly is what is in style/stylefunctions/default.js
- The same is the result for the default featureStyles defined in draganddrop.js

Grammostola:
…g the past. Aims to resolve #1764 . The functionality became unavailable after https://github.com/origo-map/origo/blob/1451535bf749236ab4e54034492fce4d472d89b6/src/controls/draganddrop.js it seems.  This PR "fixes it" by making drag and drop layers definitely recognizable and then just letting the style machinery behind the viewer do its somewhat complicated thing and then more or less behave like at the mentioned point in the history of draganddrop.js (add a layer with a style to the map no matter if it isn't registered with the viewer). It shouldn't interfere with anything else.
We can hopefully discuss how appropriate this is tomorrow.

### DIFF
--- a/src/controls/draganddrop.js
+++ b/src/controls/draganddrop.js
@@ -155,6 +155,7 @@ const DragAndDrop = function DragAndDrop(options = {}) {
           queryable: true,
           removable: true,
           promptlessRemoval,
+          isDragDroppedLayer: true,
           visible: true,
           source: 'none',
           type: 'GEOJSON',
--- a/src/viewer.js
+++ b/src/viewer.js
@@ -424,6 +424,11 @@ const Viewer = function Viewer(targetOption, options = {}) {
     if (insertBefore) {
       map.getLayers().insertAt(map.getLayers().getArray().indexOf(insertBefore), layer);
     } else {
+      if (layerProps.isDragDroppedLayer === true) {
+        if (layerProps.styleByAttribute === false) {
+          layer.setStyle(thisProps.style);
+        }
+      }
       map.addLayer(layer);
     }
     this.dispatch('addlayer', {


Grammostola:
The proposed fix has flaws - the style goes back to default after print exit. The style-things that follow using `viewer.addLayer()` need to be altered so that styling works after print exit, and so that styles aren't resized when changing the DPI in the print pane (which appears to have been the case at the state linked to above)

jokd:
Look into it a bit and the easiest way to solve this is probably to register the layer style when adding the layer and then it will work as usual when printing.
I'll post a PR for this to test.

Grammostola:
Closing as superseded by #1770

PR #1768, opened 2023-05-12
Fix: Image in related table

### LINKED ISSUE #1757: cannot save image in the child table for a related layer
**Describe the bug**
Hello, 
We have a related-table layer in Origo 2 and we tried to save images into a column in our Postgres database childtable but it seems it does not work (while other attributes work perfectly in the same table). 
There is no problem to upload the picture in the editing formula, it looks like this: 

image

But when we click “OK” to save the picture then it disappears, and nothing is saved in the database table.

steff-o:
Fixes #1757

Problem was that there was an asynchronous resize when the form is saved which messed up the current layer state variable.

Resolved by moving the resize to when the file is loaded (which in my opinion is more logical anyway). Previous implementation resized the image on save, which meant that the preview image was full resolution and the image was read twice.

The result should be the same which includes that the image is always resized to a max size of 600 pixels in the largest dimension. That can be overridden by setting a new max size by the undocumented maxSize option.

```JSON
"attributes": [
        {
          "name": "name",
          "type": "text"
        },
        {
          "name": "imgsrc",
          "type": "image",
          "maxSize":  50
        }
      ]
```

### DIFF
--- a/src/controls/editor/editform.js
+++ b/src/controls/editor/editform.js
@@ -78,7 +78,7 @@ const createForm = function createForm(obj) {
       const imageClass = val ? '' : 'o-hidden';
       el = `<div class="validate ${cls}"><label>${label}<br>`;
       el += `<img src="${val}" id="image-upload" class="${imageClass}"/>`;
-      el += `<input type="file" name="bildfil" id="${id}" value="${val}" accept="image/*"${disabled}></label>`;
+      el += `<input type="file" name="bildfil" id="${id}" accept="image/*"${disabled}></label>`;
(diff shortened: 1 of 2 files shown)


steff-o:
@leafisfalling, can you test this?

Owemyr:
We have tested the image in related table fix, working fine, thanks.

johnnyblasta (approved):
Tested without related table and this fixes #1787 also. 
Since it seems from two different parties that it solves when related table also, so it's good to merge then.

jokd:
Agree, merge this

PR #1769, opened 2023-05-16
feature: pick feature in editor

### LINKED ISSUE #612: Editor: To select feature covered by other feature of same layer seems impossible
Tested a build of today and when we're editing one layer, say of polygons, and one polygon happens to be completely covered by another polygon, the one beneath cannot be selected. I appreciate that this situation probably doesn't occur a lot (several layers would normally be employed I imagine) but as it did just now and since it was discussed in Slack prior I thought it'd make a good issue. Even if the result is "use several layers" : )

jokd:
Fixes #612 List the features to be able to select overlapping features.
Can be turned off by setting `"featureList": false` in the editor options

Attributes to be showed in the infowindow is setup like this on layer level. Empty attributes won't be showed and the default is to just show the feature ID:
`"featureListAttributes": ["id", "name", "type"]`

chrome-capture-2023-4-16

### DIFF
--- a/src/components/infowindow.js
+++ b/src/components/infowindow.js
@@ -1,0 +1,117 @@
+import { Component, Button, Element as El, dom } from '../ui';
+import utils from '../utils';
+
+const Infowindow = function Infowindow(options = {}) {
+  const {
+    closeIcon = '#ic_close_24px',
+    title = 'Inforuta',
+    viewer,
+    type = 'floating',
+    contentComponent = El({
+      tagName: 'div',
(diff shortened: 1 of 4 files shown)


Grammostola:
Thumbs up for the indicate-selection-on-hover. Would it make sense for a selection to stick when clicked? Like infowindow I guess. As it stands one has to take a certain care in order to move the pointer away from the info-like window to leave the indended line selected.

Grammostola:
lgtm

johnnyblasta (approved):
Tested and LGTM

PR #1770, opened 2023-05-17
fix: styling of draganddrop layers

### LINKED ISSUE #1764: featureStyles for drag-and-drop layers have no visible effect

jokd:
Fixes #1764, alternative solution to https://github.com/origo-map/origo/pull/1765

### DIFF
--- a/src/controls/draganddrop.js
+++ b/src/controls/draganddrop.js
@@ -4,7 +4,6 @@ import GeoJSONFormat from 'ol/format/GeoJSON';
 import IGCFormat from 'ol/format/IGC';
 import KMLFormat from 'ol/format/KML';
 import TopoJSONFormat from 'ol/format/TopoJSON';
-import Style from '../style';
 import { Component, InputFile, Button, Element as El } from '../ui';
 
 const DragAndDrop = function DragAndDrop(options = {}) {
@@ -116,7 +115,6 @@ const DragAndDrop = function DragAndDrop(options = {}) {
           }
         }]
       };
-      const vectorStyles = Style.createGeometryStyle(featureStyles);
       dragAndDrop = new olDragAndDrop({
         formatConstructors: [
           GPXFormat,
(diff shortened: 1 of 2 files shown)


Grammostola:
Looks good to me. Neat solution, make the viewer aware of the style, works well in print and after (legend too if less `[234, 234, 17, 1]` like in my example in the issue and more `'rgba(234, 234, 17, 1)'` and makes sense as viewer.addLayer is now employed. Well done

Grammostola (approved):

PR #1772, opened 2023-05-17
fix: legend icon color set as array

### LINKED ISSUE #1771: Legend rendering with color array
The legend icon won't render properly when the color is set as an array `[234, 234, 17, 1]`  instead of rgba-value `'rgba(234, 234, 17, 1)'`. 

Both array and rgba-string should render the icon correct.

jokd:
Fixes #1771 
Legend icon renders correct when set like `"color": "rgba(234, 234, 17, 0.5)"`  as well as`"color": [234, 234, 17, 0.5]`


johnnyblasta (approved):
Tested and LGTM

PR #1775, opened 2023-05-29
fix: let img attribute have precedence before name

### LINKED ISSUE #1774: Image and editor can't coexist img attribute
**Describe the bug**
When setup a editor with image attribute the image isn't displayed in infowindow instead the base64 is shown.

**To Reproduce**
1. Setup a editor with attribute `"type": "image"`
2. Add an image to a feature and close the editor
3. Click on the feature to show a infowindow
4. See the base64 instead of the image

**Expected behavior**
The following config:
```
        {
          "img": "bild",
          "name": "bild",
          "title": "Bild: ",
          "type": "image"
        }
```
Should result in that both the editor and infowindow should work as expected which is show the image and be able to edit the attribute.

johnnyblasta:
Fixes #1774 

Test config:
```
        {
          "img": "bild",
          "name": "bild",
          "title": "Bild: ",
          "type": "image"
        }
```

### DIFF
--- a/src/getattributes.js
+++ b/src/getattributes.js
@@ -223,8 +223,6 @@ function getAttributes(feature, layer, map) {
         } else {
           if (attribute.img || attribute.type === 'image') {
             val = getContent.img(feature, attribute, attributes, map);
-          } else if (attribute.name) {
-            val = getContent.name(feature, attribute, attributes, map);
           } else if (attribute.url) {
             val = getContent.url(feature, attribute, attributes, map);
           } else if (attribute.html) {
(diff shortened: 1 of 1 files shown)


steff-o:
Why not put _name_ last? That would fix the other attribute types as well.

johnnyblasta:
> Why not put _name_ last? That would fix the other attribute types as well.

Yeah, thought about that also after I sent in the PR.

johnnyblasta:
Moved name to last in getting attributes to let all special case have precedence over the general case.

jokd (approved):

PR #1780, opened 2023-06-08
feature: thematic styling

### LINKED ISSUE #1629: Hide features when thematic style
It would been nice to be able to hide some of the features in a thematic style.

On a thematic style there could be a checkbox in front of each style and if you want to hide a certain type you just uncheck the box and the features disappears from the map.
image

The function should perhaps be off by default and can be activated on a layer level.

jokd:
Fixes #1629
"thematicStyling": true as a layer option will enable this feature. Works hopefully with both vector and wms services. However, using stylePicker will override this setting for now.

### DIFF
--- a/src/controls/legend/overlay.js
+++ b/src/controls/legend/overlay.js
@@ -29,6 +29,7 @@ const OverlayLayer = function OverlayLayer(options) {
   let moreInfoButton;
   let popupMenu;
   let hasExtendedLegend = false;
+  let thisComponent;
 
   const checkIcon = '#ic_check_circle_24px';
   let uncheckIcon = '#ic_radio_button_unchecked_24px';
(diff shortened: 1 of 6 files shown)


Grammostola:
Very nice!! 
A couple of observers:
- What Johnny said, the remaining wms bits disappear upon changing zoom level
- I can't click the radio buttons inside the info / legend pane of the layer
- ( What classes are active in the layer isn't saved to a mapstate)
- ( What classes are active in the layer isn't understood by the Print module)

jokd:
> Very nice!! A couple of observers:
> 
> * What Johnny said, the remaining wms bits disappear upon changing zoom level
> * I can't click the radio buttons inside the info / legend pane of the layer
> * ( What classes are active in the layer isn't saved to a mapstate)
> * ( What classes are active in the layer isn't understood by the Print module)
> * ( (  (
> 
> ```json
>       "thematicStyling": true,
>       "hasThemeLegend": true
> ```
> 
> looks a little funny but it doesn't break anything. Logically `thematicStyling` isn't relevant if not `hasThemeLegend`. Perhaps `selectiveThemes` or something in stead of `thematicStyling`? ) ) )

The wms issue should be solved now. Apparently you can't send in an empty filter.
The radio buttons on the overlay properties are intentionally not clickable since that would require some new logic to handle the synchronizing of the different legends. Therefore you have the default cursor and not the pointer.
Haven't looked into mapstate but that will complicate the permalink handling quite a bit but if someone wants it it is definitely up for grabs.
I can probably fix the print legend quite easy.

Cheers!

Grammostola:
The WMS image lingering appears fixed. 
The radio buttons on the overlay props - ah, I see. Perhaps less obvious on a mobile platform? I don't suppose they can easily be removed there?
Sharing, hmm, I guess we would need the index of the turned on classes of the `thematic` array of a layer's style with a `thematicStyling` (that is also active) (and where at least one class is turned off, ie filter in effect). In theory we might `encodeURIComponent(aJSON.stringifiedArrayOfIndexNumbers)`

johnnyblasta (approved):
Great stuff!
The sharing would be a plus to have, but see no problem to let this in as is and deal with that later if needed.
When having a featureinfo popup and click on hidding the style for the feature, the popup is still visible. Not nescerraly something that should be fixed either before merge.

jokd:
I merge this and your comments can be enhancements for the future

Issue #1781, opened 2023-06-13
simpleExport for all layers


jokd:
**Is your feature request related to a problem? Please describe.**
I would like to be able to omit `layers` in the properties of the simpleExport, just like how the clientExport works. This way it can be used with every layer in the map.

**Describe the solution you'd like**
Omit `layers` property and it will kick in for every layer.

asemoller:
@jokd That would be really nice, unfortunately there's a little but here…

Currently the three export functions cannot work together on the same layer, only one at a time with the following priority: layerSpecificExport, simpleExport and ClientExport. For example, if you omit layers in clientExport and simpleExport, only simpleExport will be available on the layers.

This is not really a preferred solution. At least clientExport should work even if simpleExport or LayerSpecificExport exists for a layer. But the best would be if all three could coexist on the same layer.

I think the export function needs an overhaul regarding buttons, coexistence, merged info windows with listLayout and so on...

jokd:
I worked a bit on this today and pushed a PR. It works well in my environment but I have a basic setup so please test if it works for you as well.

Issue #1746, opened 2023-04-24
simpleExport doesn't work when using listLayout


jokd:
**Describe the bug**
When using listLayout to present the selection, the simpleExport button will not show up. Setting listLayout to false and the button shows up.

asemoller:
I guess this is related to #1711?

jokd:
Yes probably though I wasn't using round buttons so export doesn't seem to work no matter what button type

PR #1782, opened 2023-06-14
fix: rewrote the infowindow export

### LINKED ISSUE #1711: Infowindow listLayout - no round export buttons
When using round export buttons in the infowindow with listLayout: true, the export button is not shown. When listLayout: false the round export buttons work just fine except the tooltip text has some css issues.

The first one is because infowindow_expandableList.js doesn't contain any code regarding round export buttons. We would of course have tested it before submitting a PR but unfortunately we didn't. I can give it a try.

jokd:
Fixes #1711 #1746 #1781 
Merged parts of infowindow and infowindow_expandableList
Fixes round buttons in list view
simpleExport works for all layers if omitted in config


asemoller:
@jokd Wow, I set this up for us to fix after the holidays but you're super fast 👍 . Nice that you put everything in exporthandler.js. I'll test and get back!

asemoller:
I've only tested it locally and this is what I found (new and old ones 🙂):

If the layer property for clientExport is omitted like this `"layers": []`, the button (round or text button) is not displayed. However, it works fine for simpleExport now.

Default text for text buttons is "Export" for server-side exports and "Exportera" for clientExport. For round buttons the default text is "Ladda ner" for both server side and client exports. Maybe they all should have "Ladda ner" as default text, regardless of button.

If `"roundButton": true` for simpleExport, the property for "roundButtonTooltipText" doesn't work.

If `"listLayout": true` then there's some css-problems:
- lower left corner lacks rounding
- if the tooltipText is too long it will not be displayed outside the window.

This is really a great fix! We're so grateful for this contribution 🙏 !

asemoller:
I just realized that I never tested omitted layer just by removing the layer property for clientExport! If I do that, it works fine for clientExport 🙂.

jokd:
> I just realized that I never tested omitted layer just by removing the layer property for clientExport! If I do that, it works fine for clientExport 🙂.

Now it should both without layers but also an empty array. And hopefully the rest of your comments should be solved. 
The easiest way to show the tooltips without messing too much with css or rearranging the elements is to show the tooltip to the right so I just changed the direction.

asemoller:
Yes, everything seems to work now 👍!

PR #1806, opened 2023-08-17
fix: make imageLoadFunction revokeObjectURL

### LINKED ISSUE #1805: Slight potential issue with wms imageLoadFunction for requestMethod: post
**Describe the bug**
The URL.revokeObjectURL method may never get called (it should as per mdn for mem manag)

**To Reproduce**
Define a map with a wms-layer with prop `requestMethod: 'post'`
Edit
https://github.com/origo-map/origo/blob/master/src/layer/wms.js
so that a console message appears when the event handler containing the mentioned method runs like so:
```js
      img.addEventListener('loadend', () => {
        console.log('img loadend, going to revoke')
        URL.revokeObjectURL(url);
```
There will probably not be anything logged to the console

**Expected behavior**
'img loadend, going to revoke' for every tile I think

**Additional context**
The way I understand it (may be insufficiently) the HTMLImageElement doesn't seem to emit a loadend event. It has an 'onload' handler property which _appears_ to do the job. I will whip up the smallest pr

Grammostola:
Aims to fix #1805 . (Seems to createObjectUrls before revokingObjectUrls)

### DIFF
--- a/src/layer/wms.js
+++ b/src/layer/wms.js
@@ -119,7 +119,7 @@ function imageLoadFunction(loadedImage, src) {
     if (data) {
       const img = loadedImage.getImage();
       const url = URL.createObjectURL(data);
-      img.addEventListener('loadend', () => {
+      img.addEventListener('load', () => {
         URL.revokeObjectURL(url);
       });
       img.src = url;


jokd (approved):
Skilled to spot this and the fix looks spot on.

PR #1809, opened 2023-08-18
fix: check for filter before adding as thematic style

### LINKED ISSUE #1808: ThematicStyling can't handle TextSymbolizer in style
**Describe the bug**
When using `"thematicStyling": true` and the style contains TextSymbolizer it breaks, see image
image

**To Reproduce**
1. Setup a style which has a textsymbolizer in Geoserver
2. Add layer to a map with the style and set as `"thematicStyling": true`
3. Click on thematic icon

**Expected behavior**
The TextSymbolizer should be ignored and not shown in the legend.

**Screenshots**
The legend from Geoserver only contains two items
image

**Additional context**
There are a [bug in Geoserver that gives a error for TextSymbolizer if not all default properties are set](https://osgeo-org.atlassian.net/browse/GEOS-10820)

johnnyblasta:
Fixes #1808

### DIFF
--- a/src/utils/legendmaker.js
+++ b/src/utils/legendmaker.js
@@ -172,13 +172,15 @@ async function setIcon(src, cmp, styleRules, layer, viewer, clickable) {
       searchParams.set('FORMAT', 'image/png');
       searchParams.set('RULE', row.name);
       const imgUrl = decodeURIComponent(searchParams.toString());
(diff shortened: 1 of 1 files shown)


jokd (approved):

jokd:
LGTM

PR #1817, opened 2023-08-22
fix: remove suggestions when clearing search field

### LINKED ISSUE #1816: Clearing search field should also clear search results
**Describe the bug**
When searching and then clearing the search field you will get the previous search result when putting focus in the search field. This can be a bit confusing. I suggest the search results are cleared when the search field is.

jokd:
Fixes #1816


jokd:
I changed so that awesomplete handles the result list when focusing on the input field

filleg:
LGTM

Issue #1019, opened 2020-11-12
User manual


mulfvik:
It would be nice to have a user manual connected to this repo. It could be a control that renders a manual in a modal perhaps and even only contains information about active controls and menuobjects.

tonnyandersson:
Perhaps something like [this](http://www.globalcarbonatlas.org/en/CO2-emissions), instead of cramming everything into one single modal?

mulfvik:
That looks very nice. Back to the drawing table :)
Just did one i modal for our release soon to come..
[https://github.com/haninge-geodata/origo/tree/manual](https://github.com/haninge-geodata/origo/tree/manual)
[https://github.com/haninge-geodata/origo/commit/eb61bb7eac1f8edb142a7ce9cd81834dd75c687b#commitcomment-44521254](https://github.com/haninge-geodata/origo/commit/eb61bb7eac1f8edb142a7ce9cd81834dd75c687b#commitcomment-44521254)

tonnyandersson:
@mulfvik Just an idea. For inspiration, if you will. No need to throw all your work in the trash. :)

mulfvik:
Maybe have both a "guide" as @tonnyandersson tip and a "help" that diggs a little bit more deep into the controls and functions

johnnyblasta:
Since both are working on a help function on could hope that they would collaborate @paho @mulfvik

tonnyandersson:
A few suggestions:

- Use a json file for configuration. Easier to maintain, no need to rebuild just to fix a typo, etc.
- _Automatically_ check for controls present in the map and only show help for those controls. No need to produce help sections for controls not available.
- Please consider a mix between @paho's neatly organized sidebar solution and @mulfvik's concisely created control. I.e. a help control (rather than a plugin), presented in a sidebar-ish thing that does not close on map interaction and does not cover any other controls, with the information organized in groups.

mulfvik:
Hi @tonnyandersson, the status is its been a while since I did any on it. Got stuck on implementing oglide.js for a guide which I wanted to be a part of this PR. I was going to contact you for help in this regard... when time was available.
The manual, that this PR was refering to from the start :) has just a little more work to be done upon.

Yes, we use some of the code from the manual for a help modal.

I really would like to finish this but I understand if users wants this yesterday. If I promise to get it done during this summer?

tonnyandersson:
[...]

PR #1831, opened 2023-09-10
Guide

mulfvik:
#1019


mulfvik:
Hi, there is still some work to be done on this. If the guide should be able to be active on start, as the only option as it is now, it needs to relate to splash screen or if it should have a button to activate it with.

jokd:
Looks nice. I would like the guide to be started by the user from the menu, so having it configurable would be great.

mulfvik:
@jokd This is ready for review. I slimmed it down a bit to only be available through a button. Thinking that if it should be started up when initializing the map it can be done from the splash control.

Flodkvist:
Nice!
This is a good addition.

Some things that could be discussed

**Should the guide "be aware" if it´s being used on a large or small screen?**

- The pulsation does not work on collapsed legend when configuration/default is set to be expanded. 
(On small screens the legend is never expanded when map loads?).

- If using the guide on small screen the position tool is not present and the guide explaining this might be confusing.


**To be able to place the activation button in the mapmenu would be nice**

- What it does is more intuitive if combined with a text, like the other functions inside of mapmenu.

-  When using a map with more tools in o-tools-left the space runs out very quickly on a laptop screen.

**Could the popup be used for the guide? Might make the map interface more cohesive interface. Also having the dots to show the user the number of steps in the guide would be nice**
image

tonnyandersson:
I also suggest using Material Icons instead of Font Awesome as much as possible. In this case, it looks like the FA signpost icon get clipped. The MI equivalent is a better fit out of the box.

steff-o:
@mulfvik Does the latest commits mean that is is ready for a new test? Anyhow, there are merge conflicts.

mulfvik:
@steff-o No, there are still work to do to meet the requirements

PR #1835, opened 2023-09-20
feature: origin as tileGridOptions

### LINKED ISSUE #1834: origin as tileGridOptions
To load tiles outside the cached area of a tile layer, `origin `should be able to be set as a `tileGridOptions`.
Eg:
```
	"tileGridOptions": {
		"extent": [-72234.21, 6092807.47, 832132.7, 7702218.01],
		"origin": [120000,6556000],
		"tileSize": 512
	}
```

That way all requests can be aligned to the gridset, being inside the cached extent or not. If origin is omitted, the origin will be picked from top left (if `alignBottomLeft === false`) or bottom left coordinate of the extent.

jokd:
Fixes #1834
If origin is set as a `tileGridOptions` it is used instead of top/bottom left coordinate

### DIFF
--- a/src/maputils.js
+++ b/src/maputils.js
@@ -45,7 +45,9 @@ const maputils = {
   tileGrid: function tileGrid(settings, defaultSettings = {}) {
     const tileGridSettings = Object.assign({}, defaultSettings, settings);
     const extent = tileGridSettings.extent;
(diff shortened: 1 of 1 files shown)


jokd:
OK to merge this?

Issue #1773, opened 2023-05-24
Attributes named "state" gets wiped when using infowindow


steff-o:
**Describe the bug**
If a layer has an attribute named "state" it is removed when highlighted in InfoWindow

**To Reproduce**
1. Create an layer with an attribute named "state"
2. Configure origo to use infowindow
3. Select a feature from that layer by clicking it
4. Close infowindow and select the same feature again
5. The value of "state" is missing

A more intricate scenario involves editing the feature after it has been highlighted. In that case the cleared value is persisted to the database.

**Expected behavior**
Well, the value should remain intact.

**Additional context**
The problem is that _SelectionManager_ sets an attribute named "state" to indicate that a feature is highlighted in order to style it as highlighted. The obvious solution would be to use another name that is less likely to appear in a real table. More complicated solutions would be deep copying the feature when it is added to selectionmanager, but that could have some side effects if other parts of the code assumes that it accesses the actual feature. Might also be possible to use another mechanism to determine which features should have the highlight style.

steff-o:
Another approach would be keeping a reference to the highlighted features which would avoid tampering with the features completely. It would be pretty simple but there is an exposed function that makes it possible to highlight several features, which makes it necessary to save a list of highlights. Internally _selectionManager_ can only highlight one feature.  In a simplified form it would be built around:

```javascript
let highlightedFeatures = [];

function highlightFeatureById(id) {
    selectedItems.forEach((item) => {
      const feature = item.getFeature();
      if (item.getId() === id) {
        highlightedFeatures = [feature];
        component.dispatch('highlight', item);
      }
    });
 // we need to manually refresh other layers, otherwise unselecting does not take effect until the next layer refresh.
    urval.forEach((value) => {
      value.getFeatureStore().changed();
    });
  }

function featureStyler(feature) {
    if (highlightedFeatures.includes(feature)) {
         return Style.createStyleRule(multiselectStyleOptions.highlighted);
    }
    return Style.createStyleRule(multiselectStyleOptions.selected);
  }
```

In reality it needs some more changes to handle more complicated cases with remove and refreshing and the exposed api functions. The current implementation utilizes the fact that changing a feature emits a change event that triggers a redraw of that layer. 

I have written a proof of concept for this. I can clean it up and make a PR from it if desired.

Grammostola:
An array of refs to highlighted feature objs sounds nicer and somewhat more involved. Go for it : )

jokd:
[...]

PR #1846, opened 2023-09-29
feature: new multiselectionstyle for features in active layer

### LINKED ISSUE #1836: Display only selection for active layer in infowindow
It can be hard to see what features belong to a certain layer when selecting features from multiple layers, eg when a layer has `"imageFeatureInfoMode": "always"` and is always queryable.

A setting to only show the infowindows active layer. In case of listLayout: true perhaps only show features in groups that are expanded? Or toggle the groups so that only one can be expanded.

A new selection style for features in layers that are not active in the infowindow. 

**Additional context**
listLayout:false
image

"listLayout": true
image

jokd:
Fixes #1836 
However since similar work will be done in #1773 I mark it as Draft.

### DIFF
--- a/src/controls/print/print-legend.js
+++ b/src/controls/print/print-legend.js
@@ -61,7 +61,10 @@ const LayerRow = function LayerRow(options) {
     const layerName = aLayer.get('name');
     const style = viewer.getStyle(aLayer.get('styleName'));
     if (style && style[0] && style[0][0] && style[0][0].icon) {
-      return `${style[0][0].icon.src}&format=${format}`;
+      if (style[0][0].icon.src.includes('?')) {
+        return `${style[0][0].icon.src}&format=${format}`;
+      }
+      return `${style[0][0].icon.src}?format=${format}`;
     }
     return `${url}?SERVICE=WMS&layer=${layerName}&format=${format}&version=1.1.1&request=getLegendGraphic&scale=401&legend_options=dpi:300`;
   };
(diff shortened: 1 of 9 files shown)


steff-o:
Does not work for me. It was clever to use the first part of id to determine if the feature belongs to the active layer. Maybe too clever as not all layers have the layer name as a part of the id. For example the _origo-cities_ geojson does not work. Also AGS layers probably won't work. Also I'm not sure how that id corresponds to the selection group name, especially when the layer name parameter does not equal the layer id (when using the id parameter). Probably some lookup mechanism is needed if the feature itself shall not carry information of which layer it belongs to, something that we want to avoid.

Besides that you left a console.log and setting of property _inAvctiveLayer_ in _onItemAdded()_ which is probably never used.

jokd:
> Does not work for me. It was clever to use the first part of id to determine if the feature belongs to the active layer. Maybe too clever as not all layers have the layer name as a part of the id. For example the _origo-cities_ geojson does not work. Also AGS layers probably won't work. Also I'm not sure how that id corresponds to the selection group name, especially when the layer name parameter does not equal the layer id (when using the id parameter). Probably some lookup mechanism is needed if the feature itself shall not carry information of which layer it belongs to, something that we want to avoid.
> 
> Besides that you left a console.log and setting of property _inAvctiveLayer_ in _onItemAdded()_ which is probably never used.

Right on. Hopefully this is fixed now

steff-o:
LGTM. That was a clever fix. If I could approve stuff I would.

As a side note: it is not only useful when using Multiselect as you wrote, it also works for ordinary featureInfo when infowindow is used.

PR #1855, opened 2023-10-11
fix: don't use state attribut for highlight

### LINKED ISSUE #1773: Attributes named "state" gets wiped when using infowindow

steff-o:
Fixes #1773 by making SelectionManager keeping track of highlighted items instead of adding an attribute to highlighted features.

No new configuration is added or api:s affected.


jokd:
Works well however I would prefer if the highlighted feature was unhighlighted when switching to another layer.

PR #1857, opened 2023-10-17
fix: style for all types in draganddrop

### LINKED ISSUE #1856: Dragandrop only shows one geometry type in FeatureCollection
**Describe the bug**
If you have a GeoJSON FeatureCollection that contains for example both LineString and Point types only the first type will be shown on the map.

**To Reproduce**
1. Drag a GeoJSON that contains both lines and points.
2. Only the type that is first in the features list will have a style and be shown in the map.

**Expected behavior**
There should be a style for every type of geometries there is in the GeoJSON.

**Screenshots**
Original drawn objects with both points and lines:
image
Same area when drag and dropped (the point is missing):
image

johnnyblasta:
Fixes #1856

### DIFF
--- a/src/controls/draganddrop.js
+++ b/src/controls/draganddrop.js
@@ -162,7 +162,15 @@ const DragAndDrop = function DragAndDrop(options = {}) {
           features: event.features
         };
         if (!styleByAttribute) {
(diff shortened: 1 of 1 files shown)


jokd (approved):
LGTM

PR #1864, opened 2023-10-20
feature: add zoomToExtent option for draw layers

### LINKED ISSUE #1862: zoomTo option for draw control
**Describe the bug**
It should be possible to set zoomTo option for draw layers, making it possible to zoom to layer extent from the layer (three-dot) menu

jokd:
Fixes #1862 
To NOT use this feature:
`"zoomToExtent": false`

Defaults to true


johnnyblasta (approved):
LGTM

PR #1865, opened 2023-10-20
fix: respect layers parameter in getFeatureInfoRequests

### LINKED ISSUE #1863: Customize the getFeatureInfoRequests request
The getFeatureInfoRequests function should use the layers parameter in case it is set. This makes it easier to make requests from a plugin, eg the multiselectiontool.

jokd:
Fixes #1863 
With this fix the multiselection tool will fully work and getfeatureinfo will exclude layers that are outside their zoom limits.

### DIFF
--- a/src/featureinfo.js
+++ b/src/featureinfo.js
@@ -634,7 +634,6 @@ const Featureinfo = function Featureinfo(options = {}) {
     const pixel = evt.pixel;
     const map = viewer.getMap();
     const coordinate = evt.coordinate;
-    const layers = viewer.getQueryableLayers();
     const clientResult = getFeatureInfo.getFeaturesAtPixel({
       coordinate,
       clusterFeatureinfoLevel,
@@ -646,7 +645,6 @@ const Featureinfo = function Featureinfo(options = {}) {
     if (clientResult !== false) {
       getFeatureInfo.getFeaturesFromRemote({
         coordinate,
-        layers,
         map,
         pixel
       }, viewer)
--- a/src/getfeatureinfo.js
+++ b/src/getfeatureinfo.js
@@ -215,7 +215,8 @@ function getFeatureInfoRequests({
         });
       }
     });
-  };
+  }
+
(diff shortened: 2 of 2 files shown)


Grammostola:
The change appears what is suggested, very neat.
 It seems to affect the pointer ms tool. I can't quite parse what you mean that the suggested change means in practice for what type of layers with the ms-tool (but I haven't detected any problem with the suggested change).

jokd:
> The change appears what is suggested, very neat. It seems to affect the pointer ms tool. I can't quite parse what you mean that the suggested change means in practice for what type of layers with the ms-tool (but I haven't detected any problem with the suggested change).

The `layers` parameter sent to the `getFeatureInfoRequests` function was never used before. Now if layers is set it will be used instead looping through queryableLayers in the function.

Grammostola:
Yes. It might be very useful for other users of `getFeaturesFromRemote`, like the MS tool (based upon your suggestion). I can see that the intended changes apply to the pointer tool, I merely couldn't quiet fathom what you meant with "out of zoom" and the practical application (for the MS tool) . But it has been a while since I tested the original MS plugin.

jokd:
> Yes. It might be very useful for other users of `getFeaturesFromRemote`, like the MS tool (based upon your suggestion). I can see that the intended changes apply to the pointer tool, I merely couldn't quite fathom what you meant with "out of zoom" and the practical application (for the MS tool) . But it has been a while since I tested the original MS plugin.

The `shouldSkipLayer` (https://github.com/origo-map/multiselect-plugin/blob/main/src/multiselect.js#L246) checks if layer is queryable, shouldn't be excluded according to config and is within zoom range

Grammostola:
In the small map I tried the MS with a WMS layer and the origo cities layer I did not see that function being called. I'm not sure that matters. If you are happy with this change then I'm happy. On the basis of it not seeming to have any adverse affects and employing a logical seeming change, it lgtm.

PR #1869, opened 2023-10-20
fix: draw control: set layer as visible when activated

### LINKED ISSUE #1868: Activating draw layer should also set it to visible: true
**Describe the bug**
When activating a layer in the draw control, the layer is not automatically set to visible: true which it should.

jokd:
Fixes #1868


johnnyblasta (approved):
LGTM

PR #1873, opened 2023-10-24
fix: no legend message

### LINKED ISSUE #1858: Draw legend is hiding text
**Describe the bug**
The text "Legend saknas" is hiding in the Draw legend

**To Reproduce**
1. Add the draw control to the map and activate it
2. Click on the 3 dots in the layer list to open the legend
3. The text "Legend saknas" is hiding

**Screenshots**
origo_draw_ritlager_legend

Grammostola:
Aims to fix #1858

### DIFF
--- a/src/utils/legendmaker.js
+++ b/src/utils/legendmaker.js
@@ -239,7 +239,7 @@ export const renderExtendedThematicLegendItem = function renderExtendedThematicL
 export const Legend = function Legend({
   styleRules, layer, viewer, clickable = true, opacity = 1
 } = {}) {
-  const noLegend = 'Legend saknas';
+  const noLegend = 'Teckenförklaring saknas';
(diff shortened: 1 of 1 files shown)


Grammostola:
I was thinking two things here in order to more solidly fix the no legend message.
First, what with the recent PR to change "Legend" to "Teckenförklaring" in the Print control, the same reason should apply to the Legend control and the "No legend" message, no?

If so then 
bild
which looks like the message did prior in that it is slightly clipped. Change the padding class or perhaps change the text size rather as the text seems a bit large at present. Making it slightly smaller makes it as large as the layer title. Therefore I propose making it slightly smaller still:
bild
This is currently what the pr looks like.

jokd (approved):
LGTM

PR #1874, opened 2023-10-24
fix: make splash not assume config file

### LINKED ISSUE #1859: var origo=Origo(json) gives error if json contains splash control
Following code returns an error unless the splash control is removed from json.
```
<html>
	<head>
		<link href='css/style.css' rel='stylesheet'>
		<script src="js/origo.min.js"></script>
		<script>
			var json={
				"controls": [{
					"name": "splash",
					"options": {
						"content": "Hej!"
					}
				}],
				"projectionCode": "EPSG:3857",
				"extent": [
					-20026376.39,
					-20048966.1,
					20026376.39,
					20048966.1
				],
				"center": [
					1770000,
					8770000
				],
				"resolutions": [
					1222.99
				],
				"layers": [{
					"name": "osm",
					"title": "OpenStreetMap",
					"type": "OSM",
					"visible": true
				}]
			};
		</script>
	</head>
	<body>
		<div id="app-wrapper"></div>
		<script>
			var origo = Origo(json);
		</script>	
	</body>
</html>
```

Error in browser:

> origo.min.js:363 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'split')

No error is given if json is put in separate file and loaded with var origo = Origo('file.json').

Grammostola:
Seeks to fix #1859

### DIFF
--- a/src/controls/splash.js
+++ b/src/controls/splash.js
@@ -98,8 +98,9 @@ const Splash = function Splash(options = {}) {
       component = this;
       viewer = evt.target;
       target = viewer.getId();
-      contentKey = `splashContent;${window.location.pathname};${viewer.getMapName().split('.')[0]}`;
-      visibilityKey = `splashVisibility;${window.location.pathname};${viewer.getMapName().split('.')[0]}`;
+      const mapName = viewer.getMapName() ? viewer.getMapName().split('.')[0] : 'inline';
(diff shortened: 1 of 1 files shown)


Grammostola:
I've checked for other instances of `.getMapName()` and not found any similarly troublesome instances, have also tried a wider controls definition inline.

jokd (approved):

PR #1877, opened 2023-10-25
fix: allow percent sign in getFeature filter

### LINKED ISSUE #1523: Configured vector layer extent not fully respected

### LINKED ISSUE #1848: getFeature crashes when WFS filter contains percent sign
**Describe the bug**
getFeature can not handle percent wildcard in a filter 

**To Reproduce**
1. Configure a WFS layer with the _filter_ property set to something that contains a percent sign e.g. `"filter": "text LIKE '%tter%'"` 
2. Make a getFeature request on that layer
3. Watch error in console

**Expected behavior**
It should return those features that passes the filter

**Additional context**
Error occurs because when _getFeature_ sends its request it is not URL encoded and instead relies on the automatic URL encoding that happens for all requests from a browser. Unfortunately that automatic URL encoding does not encode percent signs as that would lead to double encoding if the URL was already URL encoded.

This is one more example that getFeature should be rewritten as just a wrapper for WFSSource to avoid double implementation that in some cases behaves a bit different (And WFSSource is always the one that does it best).

steff-o:
Fixes #1848 by moving the wfs part of getFeture() to WfsSource class as it already urlecodes arguments correctly.
Also fixes #1523 when strategy = 'all' as that wasn't fixed in #1579
Probably also solves at least part of #1534, but on the other hand, it makes #1535 and #1536 almost impossible to merge.

As a bonus some redundant code from getFeature is removed and hopefully it will be easier to implement filters for QGIS and other servers and support for different WFS versions etc by keeping all code in one place. 

I have tried to make it as backwards compatible as possible, but there is a small risk that if someone utilized or tried to compensate for the weaknesses (e.g. no coordinate transform) in a plugin or external code that may break. To my knowledge getFeature is used in search (internally in Origo) and from the MultiSelect plugin, which should be unaffected. The things that are deliberately not entirely backwards compatible are:
- The WFS version in getFeature was 1.0.0. This implementation will use the same version as WfsSource, which right now is hardcoded as 1.1.0
- The request will follow the method (GET, POST) that the layer is configured with. Old implementation of getFeature was always GET.

None of those should pose any problem unless the layer is configured only to be used by getFeature and you have a special 1.0.0 server for that that or it does not support post requests but you have configured the layer as 'POST' anyway.

### DIFF
--- a/src/getfeature.js
+++ b/src/getfeature.js
@@ -71,7 +71,7 @@ sourceType.WFS = function wfsSourceType(id, layer, serverUrl, extent) {
   // Create a temporary WFS source if layer is WMS.
   // This is a special case for multiselect which utlizes the fact that Geoserver usually has an WFS endpoint
   // at the same place as an WMS endpoint
-  if (layerType === 'WMS') {
+  if (layerType !== 'WFS') {
     // Create the necessary configuration to create a request to WFS endpoint from a WMS layer
     const sourceOpts = {
       geometryName: layer.get('geometryName'),
--- a/src/getfeature.js
+++ b/src/getfeature.js
@@ -1,11 +1,22 @@
 import EsriJSONFormat from 'ol/format/EsriJSON';
-import GeoJSONFormat from 'ol/format/GeoJSON';
-import replacer from './utils/replacer';
+import WfsSource from './layer/wfssource';
 
 let projectionCode;
 let projection;
 const sourceType = {};
 
+/**
+ * Fetches features from a layer's source but does not add them to the layer. Supports WFS and AGS_FEATURE altough functionality differs. Mainly used by search, but
+ * is also exposed as an api function that MultiSelect uses. As q quirky bonus it also support fetching features from WMS layers if there is a WFS service at the same endpoint.
+ *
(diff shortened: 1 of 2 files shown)


filleg:
Works well with QGIS server, nice cleanup! 

Except the filters with feature id, but that's a separate issue as you mentioned. 

#1535 is no longer necessary since QGIS server 3.28, which added support for different projections with GeoJSON.

PR #1878, opened 2023-10-26
fix: hide legend button when hideWhenEmbedded is true

### LINKED ISSUE #1830: Not expanded legend no hide when embedded
When configure legend  like below it isnt hidden when map is embedded
`{
      "name": "legend",
      "options": {
        "hideWhenEmbedded": true,
        "expanded": false
      }
 },`

jokd:
Fixes #1830


Grammostola (approved):
Does what it says on the tin

PR #1880, opened 2023-10-26
fix: set max height for stylepicker dropdown

### LINKED ISSUE #1849: stylePicker list objects doesnt fit window
If one have many styles in stylePicker not all styles fit in "list window". 
One solution could be that the list window is vertical scrollable

image

jokd:
Fixes #1849


Grammostola:
Looks good, albeit with one observation: would an `overflow-y` value of `auto` fit better, in order to only show the scrollbar when needed?

jokd:
> Looks good, albeit with one observation: would an `overflow-y` value of `auto` fit better, in order to only show the scrollbar when needed?

Absolutely, was a bit fast with this one

Grammostola (approved):

PR #1881, opened 2023-10-26
fix: Infowindow -headers scroll & window max-height

### LINKED ISSUE #1866: Infowindow group header should be scrollable
**Describe the bug**
When there are lots of layers in the infowindow, the list can be long and it isnt scrollable.

image

Flodkvist:
- Added scrollable container for layers listed in selection.
- Added max-height for infowindow to avoid window expanding out of screen when in default position

Fixes #1866


jokd (approved):
LGTM

PR #1883, opened 2023-10-27
Fix:support for MultiLine, MultiPolygon and MultiPoint in drawtools

### LINKED ISSUE #1872: Drawtools does not support multigeometries
**Describe the bug**
It is not possible to configure drawtools for a layer that uses a multivariant geometry (MultiPolygon, MultiLine or MultiPolygon). depending on how the configuration is made the error manifests itself in different ways. 

If the drawtools configuration is set on editor level it is according to documentation not possible to use the Multi-variants, but if you do it anyway it will work, but the label in the dropdown will read "undefined" for the default tool.

If the configuration is made on an individual layer and the tool name is set as _Polygon_ for a multipolygon layer it will not be possible to save as the geometry created is _Polygon_ in a multipolygon layer.

If the configuration is made on an individual layer and the tool name is set to _MultiPolygon_ it works but the label in the dropdown will read "undefined" for the default tool.

**To Reproduce**
See description

**Expected behavior**
The tool name should read either Polygon or Multipolygon (and the corresponding for line and point) and the created geometry should be of the same kind that they layer is configured as.

**Additional context**
There are Multivariants of the tools in OpenLayers, but they do not provide any additional functionality to actually draw multivariants, they use the same kind of interaction as its single geometry counterpart but returns a geometry of the correct type.

With that in mind it would be appealing to make a difference between tool and geometry type, i.e. it should be possible to use the "Polygon" tool for a "MultiPolygon" layer, as the "MultiPolygon" tool would indicate that you actually could draw a multipolygon. When the day comes that someone actually implements a multipolygon tool it could be added. Or we just use the geometry type as tool name as well by requiring the "MultiPolygon" tool for a MultiPolygon layer. If someone actually creates a MultiPolygon tool they will be upgraded for free to use that cool tool.

steff-o:
Fixes #1872.

It is now possible to configure drawtools on _MultiLine, MultiPolygon and MultiPoint_ both on editor level and on layer level.

No actual new configuration needed, but it is possible to use the aforementioned geometry types in configuration. Note that layers with geometry typ _MultiLineString_ is configured as _MultiLine_ as _LineString_ already was configured as _Line_.

Also note that there are no draw tools available for any kind of points and lines, but that's outside the scope of this PR. If there ever emerges any tools for these geometry types it will be possible to configure them.

Documentation will be updated accordingly.

### DIFF
--- a/src/controls/editor/copyTool.js
+++ b/src/controls/editor/copyTool.js
@@ -1,15 +1,14 @@
 import Select from 'ol/interaction/Select';
 import Feature from 'ol/Feature';
-import { MultiPolygon, MultiLineString, MultiPoint } from 'ol/geom';
+
 import dispatcher from './editdispatcher';
 
 // Point of entry. Create on of these each time the tool is selected
 //  viewer: the viewer
 //  editLayer: the destination layer for edits
 //  options: the current drawTools configuration for this layer
-const copyTool = function copyTool(viewer, editLayer, options) {
+const copyTool = function copyTool(viewer, options) {
   const map = viewer.getMap();
-  const destinationLayer = editLayer;
   let selectLayers = [];
   let hasGoups = false;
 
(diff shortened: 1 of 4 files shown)


jokd (approved):
LGTM

PR #1882, opened 2023-10-27
fix: fixed broken image issue for dynamically styled layers

tonnyandersson:
Fixes #1837, simply by not adding an icon if the layer styleName is "origoStylefunction".

PR #1885, opened 2023-10-27
fix: only issue getlegendgraphic requests for wms layers

### LINKED ISSUE #1884: Print legend handles geojson layers without a title in a suboptimal way
**Describe the bug**
When no style property is given a geojson layer (at least) the print legend will assume it's a WMS layer and issue two getlegendgraphic requests that ill fit a non-WMS layer. Further the print legend will look a bit funny.

**To Reproduce**
1. Start a default Origo instance
2. Remove the style tag from the origo-cities layer
3. Reload the map
4. Open the print panel

**Expected behavior**
Non-repeated title and an empty space for an icon visually, which rhymes with a recent PR for OrigoStyleFunction-layers and  keeps the structure of the print legend in terms of a hierarchy. In addition, no redundant WMS requests.

**Screenshots**
Current:
bild
Suggested:
bild

**Additional context**
print-legend.js:
```js
else {
        content = await getWMSJSONContent(title);
     }
```
 appears unsuitable. `else` ought to be the empty space in the shape of an icon

Grammostola:
Aims to fix #1884

### DIFF
--- a/src/controls/print/print-legend.js
+++ b/src/controls/print/print-legend.js
@@ -237,19 +237,17 @@ const LayerRow = function LayerRow(options) {
     async render() {
       const title = layer.get('title') || 'Titel saknas';
       let content = '';
-
       const style = viewer.getStyle(layer.get('styleName'));
       if (style && style[0] && (!style[0][0].extendedLegend)) {
         content = getStyleContent(title, style);
-      } else if (!layer.get('type') || layer.get('styleName') === 'origoStylefunction') {
+      } else {
         content = getTitleWithIcon(title, '');
-      } else if (layer.get('type').includes('AGS') || /\/arcgis\/services\/[^/]+\/[^/]+\/MapServer\/WMSServer/.test(getOneUrl(layer))) {
-        content = await getAGSJSONContent(title, layer.get('id'));
-      } else if (layer.get('type').includes('WMS')) {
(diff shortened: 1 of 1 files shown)


tonnyandersson:
Why not do like below?
```
} else {
  content = getTitleWithIcon(title, '');
}
```

Also, these changes make the changes in PR #1882 a bit redundant. Maybe the whole `else if` block on lines 244-245 can be removed since those situations will be covered by these changes? Or...?

Grammostola:
I see what you are seeing, the block is a bit of an eyesore.

Something like this might work.. I'll check it on monday.
```js
  const title = layer.get('title') || 'Titel saknas';
  let content = '';

  const style = viewer.getStyle(layer.get('styleName'));
  if (style && style[0] && (!style[0][0].extendedLegend)) {
    content = getStyleContent(title, style);
  } else {
    content = getTitleWithIcon(title, '');
    const lType = layer.get('type');
    if (lType && lType.includes('WMS')) {
      content = await getWMSJSONContent(title);
    } else if (lType && lType.includes('AGS') || /\/arcgis\/services\/[^/]+\/[^/]+\/MapServer\/WMSServer/.test(getOneUrl(layer))) {
      content = await getAGSJSONContent(title, layer.get('id'));
    }
  }
```

tonnyandersson (approved):

PR #1886, opened 2023-11-01
fix: set maxheight for layer search

### LINKED ISSUE #1827: Legend search results
If you configure a higher number of hits for `searchLayersLimit` the search results will overflow the screen.

A maximum height and a scrollable list could be an option. Otherwise avoid going above a certain number of hits.

jokd:
Fixes #1827 
Set max height element containing search result so it won't overflow the legend container and make it scrollable. Also set the width to match the legend.

image

### DIFF
--- a/scss/externs/_awesomplete.base.scss
+++ b/scss/externs/_awesomplete.base.scss
@@ -47,6 +47,8 @@
   top: auto;
   bottom: 100%;
   z-index: 9999999;
+  max-height: 100%;
+  overflow-y: auto;
 }
 
 .awesomplete > ul:empty {
--- a/scss/externs/_awesomplete.base.scss
+++ b/scss/externs/_awesomplete.base.scss
@@ -39,8 +39,8 @@
   left: 0px;
   list-style: none;
   margin: 0;
-  min-width: 180px;
-  max-width: 100%;
+  width: 100%;
+  min-width: unset;
   overflow-wrap: break-word;
   padding: 0;
   position: absolute;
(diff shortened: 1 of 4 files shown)


asemoller:
I haven't tested this one but how are searches handled if the legend only contains a collapsed group with many layers like in issue #1641?

jokd:
> I haven't tested this one but how are searches handled if the legend only contains a collapsed group with many layers like in issue #1641?

It will look half decent. The max-height is set as if the legend was full height so it will overflow the legend when it is smaller so I was lying a bit in my previous comment:
image

Dunno if there's a better way of doing it.

jokd:
Maybe its enough to set max height to 50% of the legend max height, at least this scenario will look better

asemoller:
Tested it and 50% works fine. I don't get a scroll in the search result though. And without turn-off-layers button and turn-on-layers button in the legend, neither close button, visible layers button nor background maps work. 😞

jokd:
> ~Tested it and 50% works fine. I don't get a scroll in the search result though.~ I had the option `"searchLayersLimit": 10"` which made it look like 50% and of course the scrolling didn't appear, my bad! 😄
> 
> However, this still applies: And without turn-off-layers button and turn-on-layers button in the legend, neither close button, visible layers button nor background maps work. 😞

Right on, now its fixed and the height is set to approx half the max height of the legend. Fixed scrolling for the search control as well

asemoller:
Yes, the legend search works well!

PR #1891, opened 2023-11-05
feature: spinner for print map loading indicator

### LINKED ISSUE #1871: Print loading no visual feedback
**Describe the bug**
When printing a PDF you get no feedback that work is being done in the background. Using a higher DPI or other settings that are demanding it can be mistaken for that something has gone wrong and the user tries to reload the page etc. 

**To Reproduce**
1. Print a A3 PDF in 300 DPI

**Expected behavior**
Some kind of spinner maybe, open for suggestions.  

**Additional context**
The possibility to abort a print would also be nice, without having to reload the page.

Grammostola:
Suggestion for an indicator in addition to tiles appearing at the edges of the print map and the two print buttons being greyed out that the print map is loading. Aims to fix #1871 

It looks like the larger "global" print spinner that shows when printing -> png and shows in the middle of the print map preview, more directly tied to the map not being finished that way I thought.

Draft pr less because I think it needs anything else and more because it may need to be something different. There are various ways of implementing a spinner or progress bar. As for the resident options in Origo, the "global" spinner is in a different place in the DOM (for a reason I guess) and has logic for that purpose (well looks rather). The "progress" control doesn't take configuration presently and is purposed as a singleton of the main map. The "spinner in utils/ appears a bit visually unappealing.

### DIFF
--- a/scss/_print.scss
+++ b/scss/_print.scss
@@ -118,11 +118,11 @@ $print-scale-line-font-size: $font-size-smallest;
 
 .print-map-loading-spinner {
   display: inline-block;
-  width: 35px;
-  height: 35px;
-  border: 2px solid rgba(255,255,255,.3);
+  width: 45px;
+  height: 45px;
+  border: 16px solid rgba(255, 255, 255, 0.884);
   border-radius: 50%;
-  border-top-color: #fff;
+  border-top-color: #008ff5;
   animation: spin 1s ease-in-out infinite;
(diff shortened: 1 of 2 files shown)


Grammostola:
Alright I had some local feedback and instead of the prior wispy thing the current incarnation has some gravitas:
bild

jokd (approved):

Issue #1899, opened 2023-11-17
Exclude parts of the mapstate in the sharemap control


jokd:
**Is your feature request related to a problem? Please describe.**
In some applications it is useful to be able to exclude certain controls/info in the mapstate of the sharemap control so this should be possible to configure.

PR #1904, opened 2023-11-23
fix: sharemap popup not displayed

### LINKED ISSUE #1900: Sharemap with wfs no longer opens overlay or infowindow
**Describe the bug**
The pop-up no longer opens when sharing a clicked feature from a WFS-layer using sharemap. Seems to be introduced since 2.8.

**Additional context**
Tested with both Geoserver and QGIS as backend. Also tested with infowindow and overlay. The sharemap link is correct with `&feature=layername.id`

steff-o:
Fixes #1900

Moved the rendering of the popup to the _featuresloadend_ event of the source as the layer event _postrender_ seems to be fired before features have been fetched from the source. Or possibly fired several times where the first is in a useless extent that does not load any features. This is probably due to a change in open layers.

As a bonus it fixes a bug in fetureInfo.showFeatureInfo that threw an exception when only one feature was sent. Previous implementation of sharemap did not suffer from this as the _postrender_ event caller swallowed the exception and the functionality that was cut short due to the exception was zooming to the feature, which no one was missing as sharemap always includes center and zoom in the links so the map was already panned to the desired extent.

Bonus 2: It is now possible to actually zoom to the extent of the feature instead of the sharemap center and zoom by creating a link that omits center and zoom.

### DIFF
--- a/src/featureinfo.js
+++ b/src/featureinfo.js
@@ -601,8 +601,8 @@ const Featureinfo = function Featureinfo(options = {}) {
   };
 
   /**
-  * Shows the featureinfo popup/sidebar/infowindow for the provided feature and fit the view to it.
-  * @param {any} featureObj An object containing layerName and feature
+  * Shows the featureinfo popup/sidebar/infowindow for the provided features and fit the view to it.
+  * @param {any} featureObj An object containing layerName and feature. "feature" is either one Feature or an Array of Feature
   * @param {any} opts An object containing options. Supported options are : coordinate, the coordinate where popup will be shown. If omitted first feature is used.
   *                                                                         ignorePan, do not autopan if type is overlay. Pan should be supressed if view is changed manually to avoid contradicting animations.
   * @returns nothing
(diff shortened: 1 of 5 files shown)


johnnyblasta:
Tested and works for a WFS layer, but not GEOJSON for which it also worked previously.

steff-o:
New attempt. Made other layers emit the _featuresloadend_ event. Also support for the (undocumented) static GeoJson type, which is impossible to attach to before the event is emitted (so it doesn't emit it).

Maybe another event should be more general, but it seems hard to avoid a race condition at startup using the various render events on both layers and map as they may be triggered several times before any features actually has been loaded which makes it hard to determine when to stop waiting.

johnnyblasta (approved):
Tested with WFS and GEOJSON and it works, saw that AGS Feature should be supported, but I haven't the means to test that.

jokd:
Looks good and making the behaviour more customizable is in line with #1899

Issue #1640, opened 2022-12-20
Layer info menu - expected behavior?


asemoller:
**Is your feature request related to a problem? Please describe.**
The behavior of the layer info menu is still a little bit confusing. Shouldn't the layer info menu close when you click "Zooma till" or "Spara ner lager"? The menu is still open if you click on either of the two options. It closes if you click the more option button again or do something else in the legend like expanding a group or making a layer visible.

I think the menu should close when clicking "Zooma till" or "Spara ner lager" or is there a reason for current behavior?

![layer-info-menu](https://user-images.githubusercontent.com/17123002/208651598-9ee438f0-aa71-4780-a551-e3e0ac6a88d9.gif)

tonnyandersson:
I don't think there's a specific reason for the current behavior. But I agree that the menu should close after clicking an option inside, I think, mainly because it's more annoying to have to close it manually every time you open it than it is to re-open it the few times you want to use several of the tools inside on the same layer. So, I vote for close-on-click.

jokd:
@asemoller Can we close this since  #1910 is merged?

Issue #1492, opened 2022-03-08
Legend - more-option button for groups and subgroups


asemoller:
**Is your feature request related to a problem? Please describe.**
It should be possible to have a more-option button even for groups and subgroups if there is an abstract for the group/subgroup.  

Read some more and look at images in the related issue #1480.

asemoller:
How should this work? If you choose to have a button instead of the text area, what would be the most user-friendly behavior for the button?

1. The text area expands and collapse.

2. A page with "group info" is displayed just like layers and "layer info".

3. The abstract is displayed in the popup menu

Any other idea?

jokd:
Let's go with #2

PR #1910, opened 2023-11-28
feature: new subgroup options

### LINKED ISSUE #1896: New group options
We have a map where we would like to be able to set the opacity for all layers in a group atonce instead of doing it for each individual layer.

Submenu for groups similar to the three-dot one on layer level with the three dots and a "settings page". Perhaps also zoom to layer(group) is useful in the submenu.

jokd:
Fixes #1896 
Adds three new options for subgroups:

"description": "Group descripton", // If set, this is shown on second page along with opacitySlider (if set)
"opacityControl": true, // If set, this is shown on second page along with description (if set). Changing the opacity will change the opacity for all layers in the subgroup. It will not affect groups in the subgroup. (Should it?!)
"zoomToExtent": true, // If set to true and there is an extent set for the group a link in the popupMenu will be created

Example config:
```
{
      "name": "grupp",
      "title": "grupp",
      "expanded": true,
      "groups": [
        {
          "name": "grupp2",
          "title": "grupp2",
          "expanded": true,
          "abstract": "Kort abstract",
          "description": "Detta är en grupp för test, visas på sida 2",
          "opacityControl": true,
          "zoomToExtent": true,
          "extent": [
            -20026376.39,
            -20048966.10,
            20026376.39,
            20048966.10
          ]
        },{
          "name": "grupp3",
          "title": "grupp3"
        }
      ]
    }
```

### DIFF
--- a/src/controls/legend.js
+++ b/src/controls/legend.js
@@ -555,6 +555,7 @@ const Legend = function Legend(options = {}) {
     },
     getuseGroupIndication() { return useGroupIndication; },
     getOverlaysCollapse() { return overlaysCmp.overlaysCollapse; },
+    getOverlays() { return overlaysCmp; },
     setVisibleLayersViewActive,
     addButtonToTools(button, buttonGroup) {
       if (buttonGroup === 'addLayerButton') {
--- a/src/controls/legend/group.js
+++ b/src/controls/legend/group.js
@@ -1,5 +1,7 @@
 import { Component, Button, Collapse, CollapseHeader, dom } from '../../ui';
 import GroupList from './grouplist';
+import createMoreInfoButton from './moreinfobutton';
+import LayerProperties from './overlayproperties';
 
 /**
  * The Group component can be a group or a subgroup,
@@ -22,7 +24,10 @@ const Group = function Group(viewer, options = {}) {
     exclusive = false,
     toggleAll = true,
     draggable = false,
-    zIndexStart = 0.1
+    zIndexStart = 0.1,
+    opacityControl = false,
+    zoomToExtent = false,
+    description
   } = options;
 
   const stateCls = {
@@ -39,6 +44,8 @@ const Group = function Group(viewer, options = {}) {
   const listCls = type === 'grouplayer' ? 'divider-start padding-left padding-top-small' : '';
   const groupList = GroupList({ viewer, cls: listCls, abstract });
   visibleState = groupList.getVisible();
+
+  const thisGroup = viewer.getGroup(name);
 
   const getEl = () => groupEl;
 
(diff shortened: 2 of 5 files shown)


jokd:
> Why only for subgroups? Isn't there a case where you'd want this on a main group?

I thought subgroups are more similar to layers and could have this functionality but perhaps you are right and it should be possible. However I think groups are handled a bit different than subgroups but can have a look at it.

johnnyblasta:
> > Why only for subgroups? Isn't there a case where you'd want this on a main group?
> 
> I thought subgroups are more similar to layers and could have this functionality but perhaps you are right and it should be possible. However I think groups are handled a bit different than subgroups but can have a look at it.

Yes, the handling is different and therefore I approved it after testing, but the user might not understand why it wouldn't be possible to set on main group also.

jokd:
> > > Why only for subgroups? Isn't there a case where you'd want this on a main group?
> > 
> > 
> > I thought subgroups are more similar to layers and could have this functionality but perhaps you are right and it should be possible. However I think groups are handled a bit different than subgroups but can have a look at it.
> 
> Yes, the handling is different and therefore I approved it after testing, but the user might not understand why it wouldn't be possible to set on main group also.

I had another go at it and now it will work on top level groups as well.
It will also fix #1640

filleg:
Tried it with the latest changes and it works fine on the main group, but I can't get the moreinfo button to work on the subgroups.
Nice feature btw! Seems to cover #1492 as well.

jokd:
> > > Tried it with the latest changes and it works fine on the main group, but I can't get the moreinfo button to work on the subgroups. Nice feature btw! Seems to cover #1492 as well.
> > 
> > 
> > Had totally forgotten about that issue and it looks spot on this. Whats your issue with subgroups?
> 
> The moreinfo button works when only setting opacityControl:true, but nothing happens when I click if I also add zoomToExtent. Works well on the main group however.

Right on; might have messed up the popup when adding the group functionality. Will fix.

asemoller:
You bet I do!

Abstract, description and showAbstract work fine. My only input is that the abstract/description is a bit confusing. Abstract's main position should be in the group info page just like it is for layers and not in the legend as it is now. If you want additional information that appears directly in the legend, it should be called description. So switching these two should be more user friendly for admins. But I understand if you don't want to make that change because backwards compatibility will be a bit damaged but I think it will be better in the end. As for showAbstract, I'm not convinced. Personally, I don't think that the same information should be shown in two places. But that's just me! 

Really nice improvements!

jokd:
OK new suggestion: 
abstract and description are both to be found on the properties page, unless showAbstractInLegend is set to true and then it will behave like before. I suggest we default this to false so the abstract will move to the properties page unless configured to do so.

The reason for having both abstract and description as options is that some plugins/functions use abstract for adding layer information, the desciption option is a way to add another string.

jokd:
> > > Tried it with the latest changes and it works fine on the main group, but I can't get the moreinfo button to work on the subgroups. Nice feature btw! Seems to cover #1492 as well.
> > 
> > 
> > Had totally forgotten about that issue and it looks spot on this. Whats your issue with subgroups?
> 
> The moreinfo button works when only setting opacityControl:true, but nothing happens when I click if I also add zoomToExtent. Works well on the main group however.

Fixed now, managed to trigger click event twice

johnnyblasta:
Tested again and still approve.
Good work!

Issue #1801, opened 2023-08-07
Mapbox vector tile layers aren't rescaled properly in Print -> 300dpi


Grammostola:
**Describe the bug**
If a mapbox style vector tile layer is employed in a map and the print control is activated and the print resolution is upped to 300 dpi then the features of the layer will become thinner and smaller.

**To Reproduce**
- Have a vector tile layer with a mapbox style definition in a file / adress , for instance the LM sample mentioned in https://github.com/origo-map/origo/pull/929 
- Print : 300dpi

**Expected behavior**
It should rescale more properly like most other vector features nowadays

**Additional context**
Like most other vector features the style rules need to be altered. These style rules are fetched and not reused in src/extensions/olms.js
In order to construct a scaler-function for mapbox style rules the contents of the style file/location need to be accessible after it has been read. Changing the `olmsStyle` function to async in order to await the fetch in order to return the read style has repercussions further along (it needs to be awaited / then:ed and so on). 

It makes sense for the style to be fetched once at any rate.

PR #1911, opened 2023-12-04
Fix: Rescale vector tiles depending on dpi in print

sweco-seguro:
The olms extension now rescales mapbox vector tiles when changing the dpi of a print.

The extension creates one style for each configured print resolution, which is then applied when said print resolution is chosen.

Fix for [#1801](https://github.com/origo-map/origo/issues/1801)

### DIFF
--- a/src/controls/print/print-resize.js
+++ b/src/controls/print/print-resize.js
@@ -3,6 +3,7 @@ import ImageLayer from 'ol/layer/Image';
 import TileLayer from 'ol/layer/Tile';
 import VectorLayer from 'ol/layer/Vector';
 import VectorImageLayer from 'ol/layer/VectorImage';
+import VectorTileLayer from 'ol/layer/VectorTile';
 import { OSM, WMTS, XYZ, ImageWMS, ImageArcGISRest, Cluster } from 'ol/source';
 import TileArcGISRest from 'ol/source/TileArcGISRest';
 import agsMap from '../../layer/agsmap';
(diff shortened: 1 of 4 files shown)


jokd (approved):
Unable to test this but it looks good so I suggest we merge it

Grammostola:
There are one or two changes incoming

Grammostola:
There's a minor niggle with an error when the print preview exits, a default param is given a `null` value, else it looks good. 

There are two potential discussion points:
- The Olms extension now takes a config object with a possible list of resolutions in the case that the print has somehow been initialized with different resolutions than the current ones (75, 150, 300). 
- The `fill-pattern` paint prop for polygons can look slightly funny at 75 and 300 dpi. Not "obviously wrong" that I've seen but not optimal always either. The prop can acc to the spec refer to a raster (not a vector graphic, I asked the mapbox style spec resident AI) in the spritesheet or another image. https://docs.mapbox.com/style-spec/reference/layers/#paint-fill-fill-pattern . With "spritezero" a spritesheet can be created from .svg-graphics for hiDPI displays and standard displays and  ol-mapbox-style I'd guess detect the display and apply the "2x" sprite sheet or the normal one. Inside one sprite sheet there can be several rules with several different raster images for the same layer at different zoom levels. "Different print dpis" isn't so much a concept unless we choose to apply it, that is to say make the print control suppose style rules with a distinct print-dpi-related name and those rules and images would need to be in (each) style sheet.

PR #1918, opened 2023-12-07
feature: new search params

### LINKED ISSUE #1917: Another search param when hitting enter
When searching and hitting enter the results are shown in the searchlist (if set to do so). It would be useful if this search request had a param that could be used for the search endpoint. Eg do a wildcard search which would be too demanding to do for every figure typed.

Add a query param to the get request.

jokd:
Fixes #1917

### DIFF
--- a/src/controls/search.js
+++ b/src/controls/search.js
@@ -508,11 +508,21 @@ const Search = function Search(options = {}) {
       infowindow.changeContent(listcomponent, `${searchlistTitle.replace('{{value}}', searchVal)}`);
     };
 
-    function makeRequest(reqHandler, obj, opt, ignoreGroup = false) {
+    function makeRequest(params) {
+      const {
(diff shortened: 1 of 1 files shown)


asemoller:
Tested this with the associated changes in origo server and it works like a charm!
You're a ⭐!

jokd:
> Tested this with the associated changes in origo server and it works like a charm! You're a ⭐!

Good to hear! The search function on Origo server could use some tlc to prevent sql injection but that is perhaps another issue.

steff-o (approved):
LGTM.

Implementation seems OK, but it actually introduces more features than advertised as the "queryType"-parameter is out of scope of the original issue. I expect it to be well documented.

And next time use URLSearchParams instead of that hideous string concatenation. I know that it would have required rewriting some old stuff to make it work in this case.

PR #1920, opened 2023-12-07
feature: infowindow for search

### LINKED ISSUE #1228: Search hits could have a choice of infowindow: type

jokd:
Fixes #1228
Use infowindow for search result if set

### DIFF
--- a/src/controls/search.js
+++ b/src/controls/search.js
@@ -3,7 +3,6 @@ import Point from 'ol/geom/Point';
 import Awesomplete from 'awesomplete';
 import { Component, Element as El, Button, Collapse, CollapseHeader, dom } from '../ui';
 import generateUUID from '../utils/generateuuid';
-import getAttributes from '../getattributes';
 import getCenter from '../geometry/getcenter';
 import getFeature from '../getfeature';
 import mapUtils from '../maputils';
@@ -134,7 +133,8 @@ const Search = function Search(options = {}) {
           let featureWkt;
           let coordWkt;
           if (res.length > 0) {
-            showFeatureInfo(res, layer.get('title'), getAttributes(res[0], layer, map));
+            const featLayerName = layer.get('name');
+            featureInfo.showFeatureInfo({ feature: res, layerName: featLayerName }, { maxZoomLevel });
           } else if (geometryAttribute) {
             // Fallback if no geometry in response
             featureWkt = mapUtils.wktToFeature(data[geometryAttribute], projectionCode);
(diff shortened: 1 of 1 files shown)


steff-o:
From code inspection it looks like there are still two cases where it is hardcoded to use overlay via the internal _showFeatureInfo_ function instead of calling the function on featureInfo.

I haven't figured out if it not possible to use the featureInfo-version of showFeatureInfo, but if it is possible it should be changed everywhere so it always obeys the setting and the internal showFeatureInfo can be removed.

jokd:
> From code inspection it looks like there are still two cases where it is hardcoded to use overlay via the internal _showFeatureInfo_ function instead of calling the function on featureInfo.
> 
> I haven't figured out if it not possible to use the featureInfo-version of showFeatureInfo, but if it is possible it should be changed everywhere so it always obeys the setting and the internal showFeatureInfo can be removed.

In that case the featureInfo.showFeatureinfo has to be changed to allow features that are not in a layer. Maybe that can be adressed in another issue?

steff-o:
Ok, that was what I expected, just didn't have time too dig into the details but I remembered that there was something special about how search (mis)used featureInfo.render().

If it is a larger rewrite we could probably settle for this PR at the moment, but it would be nice fix all cases to avoid confusion.

jokd:
> Ok, that was what I expected, just didn't have time too dig into the details but I remembered that there was something special about how search (mis)used featureInfo.render().
> 
> If it is a larger rewrite we could probably settle for this PR at the moment, but it would be nice fix all cases to avoid confusion.

I think it's quite a rewrite to fix the other cases so I suggest we fix that in another issue. 
So if it's ok it would be nice if this was merged.

steff-o (approved):
LGTM. 

But...
The case in line 147 does not make any sense. The newly created feature does not have any attributes, so infowindow will be empty. But that was also the case before this PR. Maybe the intention was that all attributes should be read from search response? In that case it would make sense to point out the layer to get the correct attribute formatting. Well, at least it will obey the "infoWindow"-setting without breaking anything.

Overall I don't think that the documentation is describing the five cases correctly. For instance, case 2 seems impossible to achieve.  #1420 would be welcome.

jokd:
> LGTM.
> 
> But... The case in line 147 does not make any sense. The newly created feature does not have any attributes, so infowindow will be empty. But that was also the case before this PR. Maybe the intention was that all attributes should be read from search response? In that case it would make sense to point out the layer to get the correct attribute formatting. Well, at least it will obey the "infoWindow"-setting without breaking anything.
> 
> Overall I don't think that the documentation is describing the five cases correctly. For instance, case 2 seems impossible to achieve. #1420 would be welcome.

Yeah it is a bit messy but this PR aims to use the configured infowindow where possible. Further improvements can be addressed and solved in another PR, such as issue #1420

PR #1923, opened 2023-12-07
fix: allow 0 as ID

### LINKED ISSUE #1922: searchlist, allow 0 as ID
**Describe the bug**
When clicking objects in the searchlist with ID of 0, you will select every item in the layer since FeatureId isn't part of the getFeature request. Despite being bad practice to use 0 as ID, sometimes it is still a fact and should be handled.

jokd:
Fixes #1922


Grammostola:
I don't have a quick way of testing this at hand but the changes appear to make crystal sense so lgtm : )

tonnyandersson (approved):

PR #1925, opened 2023-12-07
fix: change word wrapping for subgroups

### LINKED ISSUE #1924: Change word wrapping for groups
**Describe the bug**
Groups in the legend has word-break: break-all rule which cuts names in a often not so nice way. I suggest we use overflow-wrap: anywhere instead

**Screenshots**
word-break: break-all
image

overflow-wrap: anywhere
image

jokd:
Fixes #1924


Grammostola:
Lgtm (the issue appears for me only if the legend is given a specific width, else it expands horizontally in order to accomodate long group names)

PR #1929, opened 2023-12-08
fix: group header issue

### LINKED ISSUE #1928: Group with context menu jinxes the legend layout
**Describe the bug**
If group options are given a group so that it yields a context menu, activating said menu will make the legend display with issues

**To Reproduce**
Have a group (or a subgroup) with the following options 
```json
          "opacityControl": true,
          "zoomToExtent": true,
          "extent": [
            -20026376.39,
            -20048966.10,
            20026376.39,
            20048966.10
          ]
```

**Expected behavior**
The same as when activating a layer's context menu

**Screenshots**

bild

**Additional context**
The flex element that the context menu creates doesn't seem to stretch south of the item row as it does with layers ( I am not the css wizards)

A context menu requires two context menu options for a group which makes this slightly more of an edge case

jokd:
Fixes #1928

### DIFF
--- a/scss/ui/_flex.scss
+++ b/scss/ui/_flex.scss
@@ -99,6 +99,10 @@ $flex-grow-all: 2;
     align-items: stretch;
   }
 
+  .basis-0 {
+    flex-basis: 0;
+  }
+
   .basis-25 {
     flex-basis: 25%;
   }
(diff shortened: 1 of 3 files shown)


Grammostola:
Lgtm

PR #1938, opened 2023-12-22
Fix: Correct print scales even when constrainResolution is true

### LINKED ISSUE #1934: constrainResolution + print scales issue
`constrainResolution:true` for the map and a scales array for the print control do not dwell well together

**To Reproduce**
- Setup an Origo map with a series of zoom levels that suitably match a cached background map's zoom levels
- use `constrainResolution:true` for the map
- setup the print control with a series of scales that do not match the map's zoom levels
- Goto print preview and change the scales of the dropdown. The scale dropdown will suddenly show a map scale rather than a print scale as the print map shares the constrainResolution prop

**Expected behavior**
Whether or not the map has constrainResolution and scales that make sense to no man if the print control is configured with a series of (probably common for printed maps) scales the print control should do a better job of serving those scales

**Screenshots**
bild.

**Additional context**
The scales configured with the print control should rule supreme to all possible extent since "human readable scales" make sense only as pdfs with known physical sizes.  Either constrainResolutions but with these scales or do not constrainResolution in the print preview even if the map is configured with this prop. 

Edit. Of course the type of background map service determines what the print control scales array may dictate. If there's a cached image based service with no available underlying WMS service then the map author will need to determine (the same) zoom levels / scales for the map as well as the print control. This issue is for when there's a reason to give the print control an array of scales that fit printed maps with a known physical size because the (main) map is not using those zoom levels. Typically a vector tile background map (the intermittent scales might be taken care of via `renderMode: vector` in the print control), but it could be a cached WMS map with an accessible underlying WMS service as well.

sweco-seguro:
Correct print scales are now retained even when the ConstrainResolution option is true.

ConstrainResolution gets temporarily set to false when using the print component and is restored once the print component is closed.

Fixes #1934 in regard to ConstrainResolution. The scales are however still constrained by the max and min resolutions of the map, regardless of if ConstrainResolution is true or false.

### DIFF
--- a/src/controls/print/print-component.js
+++ b/src/controls/print/print-component.js
@@ -271,7 +271,8 @@ const PrintComponent = function PrintComponent(options = {}) {
     titleComponent,
     descriptionComponent,
     createdComponent,
-    closeButton
+    closeButton,
+    constrainResolution: view.getConstrainResolution()
   });
 
   const setScale = function setScale(scale) {
--- a/src/controls/print/print-resize.js
+++ b/src/controls/print/print-resize.js
@@ -19,7 +19,8 @@ export default function PrintResize(options = {}) {
     titleComponent,
     descriptionComponent,
     createdComponent,
-    closeButton
+    closeButton,
+    constrainResolution
   } = options;
 
   let {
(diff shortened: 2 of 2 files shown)


Grammostola:
It appears to work as intended, i.e if print scales are supplied then those are available in the scale dropdown, if they aren't supplied then the dropdown shows the scales of the map. I haven't seen a need or `renderMode: vector` for vector tiles either as I see no quality loss when employing even print scales that do not rhyme with the vector tile cache. 

A question I see is a cached WMTS or WMS map with no underlying WMS and zooming with pinch or scroll. The MDK cached background map service doesn't seem to have a problem but I'm not aware of how it is setup. That map does not employ constrainResolution and the tile requests never seem to miss. If someone has a setup with a cached map that needs constrainResolution (because there is no underlying wms service) then maybe they could advice what happens without constrainResolution. If a problem then this PR could simply be a print config option.

PR #1943, opened 2024-01-11
fix: make draganddrop use existing style function for features withou…

### LINKED ISSUE #1935: Drag and drop styles for more geometry types would be a boon
If one enables the drag and drop control and drags and drops the data/origo-mask.geojson the app crashes because there is no default style for "MultiPolygon" for the drag and drop control. (And the documentation isn't super explicit regarding this) 

**Expected behavior**
I would expect a default style for Point, LineString, Polygon, MultiPolygon, MultiLineString and MultiPoint and maybe MultiGeometry (kml).

Grammostola:
…t a defined style. Seeks to fix #1935 . (The fix includes attempting to make kml files that from what I gather employ their own style not have the wrong legend. Yes it is a bonus, I was testing my proposed fix with different file formats. It can be removed.)

Creds to @jokd for the idea.

### DIFF
--- a/src/controls/draganddrop.js
+++ b/src/controls/draganddrop.js
@@ -74,49 +74,7 @@ const DragAndDrop = function DragAndDrop(options = {}) {
       const styleByAttribute = options.styleByAttribute || false;
       const zoomToExtent = options.zoomToExtent !== false;
       const zoomToExtentOnLoad = options.zoomToExtentOnLoad !== false;
-      const featureStyles = options.featureStyles || {
-        Point: [{
-          circle: {
-            radius: 5,
-            stroke: {
-              color: [0, 255, 255, 1],
-              width: 0
-            },
-            fill: {
-              color: [0, 255, 255, 1]
-            }
-          }
-        }],
-        LineString: [{
-          stroke: {
-            color: [255, 255, 255, 1],
-            width: 5
-          }
-        },
-        {
-          stroke: {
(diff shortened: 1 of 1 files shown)


johnnyblasta:
Tested with a KML and got a error.
image

Fixed it with this code with a check if geometry exists, on row 128 in draganddrop.js:

```
            if (feature.getGeometry() !== null) {
				const geometryType = feature.getGeometry().getType();
				if (featureStyles && featureStyles[geometryType]) {
				  if (!types.has(geometryType)) {
					styles.push(...featureStyles[geometryType]);
					types.add(geometryType);
				  }
				} else {
				  getStyleFunction(feature);
				}
            }
```

johnnyblasta:
> Cheers, I hadn't thought of adding a layer without geometry. What usage scenario are you thinking of? Feel free to push the proposed fix for that anyway.

I had a KML which was exported from Autocad which apparently had some placemark without geometry. Don't know why that is, but would be nice if it can handle empty geometry without crashing.

johnnyblasta:
@Grammostola made a commit in your branch.

johnnyblasta (approved):
LGTM

Grammostola:
> I had a KML which was exported from Autocad which apparently had some placemark without geometry. Don't know why that is, but would be nice if it can handle empty geometry without crashing.

Indeed, nice testing!

PR #1946, opened 2024-01-17
fix: wrap in infowindow to stopp overflow

### LINKED ISSUE #1939: Infowindow don't break long attribute values
**Describe the bug**
When there are long attribute values and using infowindow the word don't break and stretches the element and adds a scrollbar which can be used to view the whole attribute value. The dropdown arrow is pushed away and isn't visible unless scrolled to the right.

**To Reproduce**
1. Set 
```
  "featureinfoOptions": {
    "infowindow": "infowindow"
  }
```
2. Have a long none break attribute value
3. See error

**Expected behavior**
Found to ways to resolve this.

1. In _infowindow.scss remove the `width: 100%;` in .o-identify-content

```
.o-identify-content {
  display: inline-block;
  width: 100%;
}
```

2. In _infowindow.scss add `word-break: break-all;` in .o-identify-content

```
.o-identify-content {
  display: inline-block;
  width: 100%;
  word-break: break-all;
}
```

Which would be the preferred way of handling this?

**Screenshots**
image
image

johnnyblasta:
Fixes #1939

### DIFF
--- a/scss/_infowindow.scss
+++ b/scss/_infowindow.scss
@@ -174,6 +174,7 @@
 .o-identify-content {
   display: inline-block;
   width: 100%;
+  overflow-wrap: anywhere;
 }
 
 .folded {


Grammostola:
Lgtm

PR #1947, opened 2024-01-18
feature: rotate text and point in draw control

### LINKED ISSUE #1941: Draw - enable rotate for objects
Many users have requested a rotation function for objects in the draw control, especially for texts.

A rotate slide bar (like print) in the style window could work.

johnnyblasta:
Closes #1941 
Enables rotation for text and point in draw control.

RoteraTextPunkt

### DIFF
--- a/src/style/drawstyles.js
+++ b/src/style/drawstyles.js
@@ -10,7 +10,7 @@ import {
 import { getArea, getLength } from 'ol/sphere';
 import { LineString, MultiPoint, Point } from 'ol/geom';
 
-function createRegularShape(type, pointSize, pointFill, pointStroke) {
+function createRegularShape(type, pointSize, pointFill, pointStroke, pointRotation) {
   let style;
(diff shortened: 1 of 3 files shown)


Grammostola:
Looks good overall, nice feature! One note I have is the UI and "degrees rotation", if I pull the slider all the way to "-360°" (from the default 0) my cube rotates half a rotation, which seems appropriate function-wise but doesn't quite rhyme with the UI text. (And maybe "0" could be included?)

johnnyblasta:
Corrected the wrong degrees text and added a zero.

Grammostola:
Cheers. Hm, I've not seen a rotate control in the print preview but based on the opacity slider would it be easier to revert to the default value if the scale went from 0 (default) to 360?

johnnyblasta:
Sure, this makes it easier to get back to zero.
image

Grammostola:
Lgtm  : )

Issue #1945, opened 2024-01-16
Minimize/hide button on selection infowindows that doesn't clear selection


huggla:
Infowindow shown when selecting a search match can not be hidden without the selection being cleared (the close button clears the selection). A minimize/hide button is needed.

PR #1949, opened 2024-01-22
Added minimize button to popup

huggla:
This is my attempt to add a minimize button to popup (fix for https://github.com/origo-map/origo/issues/1945). It works, but I would appreciate if someone with more experience could make the code less ugly.

### DIFF
--- a/src/popup.js
+++ b/src/popup.js
@@ -5,7 +5,7 @@ function render(target) {
       <div class="o-popup o-card">
         <div class="flex row justify-end">
           <div id="o-card-title" class="justify-start margin-y-smaller margin-left text-weight-bold" style="width:100%;"></div>
-          <button id="o-minimize-button" class="small round margin-top-smaller margin-bottom-auto margin-right-small icon-smallest grey-lightest no-shrink" aria-label="Minimera">
+          <button id="o-minimize-button" class="small round margin-top-smaller margin-bottom-auto margin-left-small margin-right-small icon-smallest grey-lightest no-shrink" aria-label="Minimera">
             <span class="icon ">_</span>
           </button>
           <button id="o-close-button" class="small round margin-top-smaller margin-bottom-auto margin-right-small icon-smallest grey-lightest no-shrink" aria-label="Stäng">
@@ -63,8 +63,12 @@ function setContent(config) {
 function minimizePopup() {
   const oidentify = document.getElementById('o-identify');
   const opopup = document.getElementById('o-popup');
+  const ocardtitle = document.getElementById('o-card-title');
   oidentify.style.display = oidentify.style.display === 'none' ? 'block' : 'none';
   opopup.style.width = oidentify.style.display === 'none' ? 'auto' : null;
+  opopup.style.height = oidentify.style.display === 'none' ? '78px' : '58px';
+  opopup.children[0].style.position = oidentify.style.display === 'none' ? null : 'sticky';
+  ocardtitle.style.display = oidentify.style.display === 'none' ? null : 'none';
 }
 
 /**
--- a/src/popup.js
+++ b/src/popup.js
@@ -66,9 +66,9 @@ function minimizePopup() {
   const ocardtitle = document.getElementById('o-card-title');
   oidentify.style.display = oidentify.style.display === 'none' ? 'block' : 'none';
   opopup.style.width = oidentify.style.display === 'none' ? 'auto' : null;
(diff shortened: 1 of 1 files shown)


huggla:
image
image

Grammostola:
I think it's a neat feature. Running the default map config and infoclicking an Origo city makes the overlay not show its contents and move to the right? 

As for pretty you could formulate it like the close button's event, meaning no event handler declaration in the html string:
```html
          <button id="o-minimize-button" class="small round margin-top-smaller margin-bottom-auto margin-right-small icon-smallest grey-lightest no-shrink" aria-label="Minimera">
            <span class="icon ">_</span>
          </button>
```

then a function where the other functions reside:

```js
function minimizePopup() {
  const oidentify = document.getElementById('o-identify');
  const opopup = document.getElementById('o-popup');
  oidentify.style.display = oidentify.style.display === 'none' ? 'block' : 'none';
  opopup.style.width = oidentify.style.display === 'none' ? 'auto' : null;
}
```
then the event handler by the close one:
```js
  function bindUIActions() {
    const closeel = document.querySelector('#o-popup .o-popup #o-close-button');
    closeel.addEventListener('click', (evt) => {
      closePopupInternal(closeCb);
      evt.preventDefault();
    });

    const minel = document.querySelector('#o-popup .o-popup #o-minimize-button');
    minel.addEventListener('click', (evt) => {
      minimizePopup();
      evt.preventDefault();
    });
  }

```

Something along those lines is what I would try anyways.

huggla:
Thanks for your input. I tried your code and it seems to work as intended, but I don't understand the part below. Could you please explain the syntax? Is (evt) => {} shorthand for function(evt) {}? What is the purpose of evt.preventDefault()? Is it strictly needed? Why not simply write minel.addEventListener('click', minimizePopup);?

minel.addEventListener('click', (evt) => {
      minimizePopup();
      evt.preventDefault();
});


________________________________________
Från: David ***@***.***>
Skickat: den 16 februari 2024 08:22
Till: origo-map/origo
Kopia: Henrik Uggla; Author
Ämne: Re: [origo-map/origo] Added minimize button to popup (PR #1949)

I think it's a neat feature. Running the default map config and infoclicking an Origo city makes the overlay not show its contents and move to the right?

As for pretty you could formulate it like the close button's event, meaning no event handler declaration in the html string:

          <button id="o-minimize-button" class="small round margin-top-smaller margin-bottom-auto margin-right-small icon-smallest grey-lightest no-shrink" aria-label="Minimera">
            <span class="icon ">_</span>
          </button>

then a function where the other functions reside:

function minimizePopup() {
  const oidentify = document.getElementById('o-identify');
  const opopup = document.getElementById('o-popup');
  oidentify.style.display = oidentify.style.display === 'none' ? 'block' : 'none';
  opopup.style.width = oidentify.style.display === 'none' ? 'auto' : null;
}

then the event handler by the close one:

  function bindUIActions() {
    const closeel = document.querySelector('#o-popup .o-popup #o-close-button');
    closeel.addEventListener('click', (evt) => {
      closePopupInternal(closeCb);
      evt.preventDefault();
    });

    const minel = document.querySelector('#o-popup .o-popup #o-minimize-button');
    minel.addEventListener('click', (evt) => {
      minimizePopup();
      evt.preventDefault();
    });
  }

Something along those lines is what I would try anyways.

huggla:
Also, what are the pros to adding event listeners at runtime vs. having them in the static html? (I'm trying to learn)

Grammostola:
> s (evt) => {} shorthand for function(evt) {}?

yes

> What is the purpose of evt.preventDefault()? Is it strictly needed? 

Based on my non-complete understanding the preventDefault method prevents the default browser method. What's the default browser method of a button click? There doesn't seem to be one. Removing it doesn't have an obvious effect so I would probably suggest that. (I merely echoed the existing button's setup in the example above. You are right to question what doesn't seem right)

> Why not simply write minel.addEventListener('click', minimizePopup);?

No reason but to keep things consistent. That's not a hard requirement though so I'd call it optional. And your way still has access to the event object (even if we don't appear to need it)

> Also, what are the pros, in this case, to adding event listeners at runtime vs. having them in the static html? 

Writing code as a string is unpleasant and hard to maintain for one

huggla:
I think it should be fixed now.

Grammostola:
Lgtm (the preventDefault() can both be removed in a separate pr some other time)

Grammostola (approved):

PR #1953, opened 2024-01-29
feature: more option for draw-controls text

### LINKED ISSUE #1950: More style options to text in draw control
It would be nice if there was a option to add a background to the text in the draw control. One advantage is that the contrast can be enhanched making it easier to read the text. 

TextBakgrund

johnnyblasta:
Closes #1950


Grammostola:
A nice addition. I noticed two things:
- Rotate resets though its slider does not whenever another prop is changed 
- If draw is activated and a point is drawn first thing the map starts panning with the cursor

johnnyblasta:
> A nice addition. I noticed two things:
> 
> * Rotate resets though its slider does not whenever another prop is changed
> * If draw is activated and a point is drawn first thing the map starts panning with the cursor. This does not appear to happen in the current master.

Fixed the problem with point drawn first.
Removed the rotate setting to zero since saw no good way of accomplish what I had actually intended to happen. At least doing it from the style window, could have done something with the drawtools and drawhandler perhaps.

Grammostola (approved):

Issue #1951, opened 2024-01-24
Sharemap doesn't work with thematicStyling


dahlalex:
**Describe the bug**

I've done the folowing: 

1. Cloned [Origo repository](https://github.com/origo-map/origo)
2. Added `"thematicStyling": true` to the layer "origo-cities"
```json
{
      "name": "origo-cities",
      "title": "Origokommuner",
      "group": "root",
      "source": "data/origo-cities-3857.geojson",
      "style": "origo-logo",
      "type": "GEOJSON",
      "thematicStyling": true,
      "attributes": [
        {
          "name": "name"
        }
      ],
      "visible": true
    }
```

3. Changed the style "origo-logo" to the following:
```json
"origo-logo": [
      [
        {
          "label": "Hallstahammar",
          "filter": "[name] == 'Hallstahammar'",
          "circle": {
            "radius": 10,
            "stroke": {
              "color": "rgba(0,0,0,0.5)",
              "width": 2.5
            },
            "fill": {
              "color": "rgba(255,0,0,0.9)"
            }
          }
        }
      ],
      [
        {
          "label": "Annat",
          "filter": "[name] !== 'Hallstahammar'",
          "circle": {
            "radius": 10,
            "stroke": {
              "color": "rgba(0,0,0,0.5)",
              "width": 2.5
            },
            "fill": {
              "color": "rgba(37,129,196,0.9)"
            }
          }
        }
      ]
    ],
```
4. Here is a screenshot to the result

![thematic](https://github.com/origo-map/origo/assets/96343158/96a3357b-d2b3-4818-b20d-7bc5e5f47ebf)

5. Deselected "Annat" so that only "Hallstahammar" is visible

![thematic2](https://github.com/origo-map/origo/assets/96343158/9d26d048-46e8-4ded-84ab-84bc769d692c)

6. Clicked on "Dela karta" and got this link [http://localhost:9966/#layers=osm/v/1/s/0/o/100,origo-mask/v/1/s/0/o/25,origo-cities/v/1/s/0/o/100&center=1810000,8390000&zoom=5.7&legend=expanded&map=index](http://localhost:9966/#layers=osm/v/1/s/0/o/100,origo-mask/v/1/s/0/o/25,origo-cities/v/1/s/0/o/100&center=1810000,8390000&zoom=5.7&legend=expanded&map=index)

**The problem**
"Dela karta" ignore that "annat" is deselected.

[...]

PR #1962, opened 2024-02-09
feature: add thematicStyling to sharemap

dahlalex:
Fixes [#issuenr 1951](https://github.com/origo-map/origo/issues/1951)

### DIFF
--- a/src/utils/legendmaker.js
+++ b/src/utils/legendmaker.js
@@ -259,11 +259,8 @@ export const Legend = function Legend({
       const style = viewer.getStyles()[styleName];
       if (layer.get('type') !== 'WMS') {
         for (let i = 0; i < style.length; i += 1) {
-          let idStr;
-          if (style[i][0].id) {
-            idStr = style[i][0].id.toString();
-          }
-          style[i][0].visible = activeThemes.includes(idStr || style[i][0].label);
+          const combinedStr = style[i][0].id?.toString() || style[i][0].label?.toString();
+          style[i][0].visible = activeThemes.includes(combinedStr);
         }
       }
     }


johnnyblasta:
Tested and works fine for GeoJSON and WFS (I assume, from the code, didn't test that), but it don't work for WMS.

johnnyblasta:
One thought:

- The config part would be more simple if the label was used instead of constructing an array with zeros and ones.

johnnyblasta:
> I've made som changes to work also for WMS and found an issue that
> 
> ```js
> origo.api().getStyles()[styleName][0].thematic[index].visible = false
> ```
> 
> has no effect.
> 
> It seems that this issue should be solved even if the label is used (if i understand correctly).

ThematicStyling for WMS works by filtering the request, so I guess a new request has to be triggered.

Grammostola:
It appears to work for WMS now, nice : ) However the link uses rule titles rather than rule names (whereas getLegendGraphic for instance is forced to use rule names). Rule names are ids and titles are for humans so I think it would be better if it the link was saved with rule names.

Grammostola:
Cool, looks good now for WMS. For vector layers no themes are active when opening the shared map. I don't have an opinion on the separator, tilde is less common and this would likely do well described in the documentation somewhere too.

Grammostola:
I used your example theme style for the origo cities layer and added an `id` prop to both. http://localhost:9966/#layers=osm/v/1/s/0/o/100,origo-mask/v/1/s/0/o/25,origo-cities/v/1/s/0/o/100/th/2&center=1810000,8390000&zoom=5.7&legend=expanded&map=index

...And the problem seems to be Number vs string. I gave the `id` prop a number value. Maybe both could be acceptable?

PR #1973, opened 2024-02-29
feature: expose loader and spinner

### LINKED ISSUE #1972: Expose global loader in Api
When writing plugins it is error prone and tedious work to have to implement a spinner knowing that it already exits in the core. Not to mention it probably breaks the common look and feel.

The global spinner functions should be exposed in the global Origo object such as: Origo.Loader.Show(), Origo.Loader.Hide(), Origo.Loader.WithLoading()

There is also an inline spinner generator function in utils. Probably best to expose that one as well. It could be named Origo.Loader.GetInlineSpinner(). (It creates some HTML that can be injected at any place in the document to display a small image)

Copy the code from loading.js, but that is error prone as implementation may change at any time.

**Additional context**

steff-o:
Fixes #1972 by adding the following properties to the global Origo object:
- Origo.Loader.show
- Origo.Loader.hide
- Origo.Loader.withLoading
- Origo.Loader.getInlineSpinner

### DIFF
--- a/origo.js
+++ b/origo.js
@@ -27,6 +27,8 @@ import 'elm-pep';
 import 'pepjs';
 import 'drag-drop-touch';
 import permalink from './src/permalink/permalink';
+import * as Loader from './src/loading';
+import Spinner from './src/utils/spinner';
 
 const Origo = function Origo(configPath, options = {}) {
   /** Reference to the returned Component */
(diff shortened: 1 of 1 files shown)


johnnyblasta (approved):
LGTM

PR #1978, opened 2024-03-13
feature: cycleButton for background layers

### LINKED ISSUE #1379: Toggle background maps
I have a simple map without a legend but want to be able to switch between different background maps and thinking of some kind of toggle button. 
Is this useful for core? 
Any idea on layout?

Without input I will probably fix something in place of the legend icon bottom right.

Grammostola:
Suggestion to fix #1379 

Background layers can have a `cycleGroup` prop. Those of the same `cycleGroup` value are part of one cycleButton (where the normal kind is a backgroundButton). 

The cycleButtons are delineated via a greyish background color. Clicking them once turns on the currently shown (icon, tooltip) layer for that button, clicking on an already active (grey border) cycleButton does not bring up the details pane, it cycles the button to the next layer in the cycleGroup.

It is in no shape or form a perfect solution since the "Örebro" solution is inherently somewhat limited. Local feedback has shown a preference for an unspecified form that appears when a background layer icon is clicked, that would be a kind of visual selector. There isn't anything like that that I know of currently in Origo, I guess the Legend layout could be rearranged.

But the current solution is like the Örebro button and it does, at least temporarily, solve a problem we were having with too many background maps (put two similar ones behind one cycleButton).

cycleButton

### DIFF
--- a/scss/ui/_button.scss
+++ b/scss/ui/_button.scss
@@ -314,13 +314,8 @@ button.active,
   }
 }
 
-.cycle-even {
+.cycle {
   background: linear-gradient(45deg, #8db2d7, #ddddee);
-  width: 37px; 
-  height: 37px; 
-}
-.cycle-odd {
-  background: linear-gradient(45deg,  #ddddee, #8db2d7);
   width: 37px; 
   height: 37px; 
 }
--- a/scss/ui/_button.scss
+++ b/scss/ui/_button.scss
@@ -313,3 +313,14 @@ button.active,
     border-right: 0;
   }
 }
+
+.cycle-even {
+  background: linear-gradient(45deg, #8db2d7, #ddddee);
+  width: 37px; 
+  height: 37px; 
+}
+.cycle-odd {
+  background: linear-gradient(45deg,  #ddddee, #8db2d7);
+  width: 37px; 
+  height: 37px; 
+}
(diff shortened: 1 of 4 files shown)


tonnyandersson:
@Grammostola

I don't think this solves the original issue since that issue specifically asked for "_simple map **without a legend** but want to be able to switch between different background maps_". I understand that it solves your internal issue with too many background maps, but that seems like a different issue to me.

A few random thoughts on the subject:

- Mixing regular buttons and cycle buttons is not very intuitive
- Too many background maps could be solved by:
    - Implementing a list/dropdown/dropup type solution for background maps
    - Add the more "exotic" background maps as regular layers in an exclusive layers group
- An option to place the background maps cycle switch directly on the map (Örebro style) or in the legend (E-tuna style) would solve both issues, but might be hard to implement (One Size Fits All rarely fits all...)

Grammostola:
Fair enough about the "without a legend" part. It seemed to me that Origo used to employ a legend in the lower left corner of the map and I wanted to keep background maps where they are now, potentially with "one button to rule them all", ie you can just have one cyclebutton. Agreed and as mentioned it isn't particularly easy to make the cycleButtons distinct. I'll close this and maybe the discussion can continue in the issue : )

PR #1984, opened 2024-03-25
fix: make items in toolbars and navigation use the align-items: flex-…

### LINKED ISSUE #1976: New class in _flex.scss
I am working on a plugin that should add a button to navigation tools/map tools and that should fold out a row of buttons on the right.
The problem is that I can not use both flex column and flex row for buttons in the same tool without them being affected by each other, i.e. so that the buttons keep the same size.

A new class needs to be created in _flex.scss

In the new class that is proposed to be called .flextoolsleft, you need to define this

```css
.flextoolsleft {
    display: flex;
    align-items: flex-start;
    width: 35px;
```

The rest of the code should be the same as in the class .flex

This below also then needs to be changed in main.js for the tools that should use it. For this purpose in the navigation och maptools that are placed in o-tools-left

Row 9 och 10:
```js
const navigation = El({ cls: 'o-navigation flextoolsleft column relative transparent spacing-vertical-small' });
const mapTools = El({ cls: 'o-maptools flextoolsleft column relative transparent spacing-vertical-small' });
```

Row 21:
```js
<div id="o-tools-left" class="box top-left transparent flextoolsleft column">
```

I have tried to use the class flex but it is not working

**Additional context**

Dalneberg:
…start strategy and tooltips in the same containers relate to the items. Aims to fix #1976

### DIFF
--- a/scss/_viewer.scss
+++ b/scss/_viewer.scss
@@ -17,6 +17,14 @@
   margin-top: .5rem;
 }
 
+.o-navigation button {
+  position: relative;
+}
+
+.o-maptools button {
+  position: relative;
+}
+
 .o-footer {
   background-color: $black;
   background-color: rgba(0, 0, 0, 0.8);
--- a/src/components/main.js
+++ b/src/components/main.js
@@ -6,8 +6,8 @@ export default function Main(options = {}) {
   } = options;
 
   const cls = `${clsSettings} o-main transparent relative flex column grow transparent height-full no-margin`.trim();
-  const navigation = El({ cls: 'o-navigation flex column relative transparent spacing-vertical-small' });
-  const mapTools = El({ cls: 'o-maptools flex column relative transparent spacing-vertical-small' });
+  const navigation = El({ cls: 'o-navigation flex column relative transparent spacing-vertical-small align-start' });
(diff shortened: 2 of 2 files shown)


tonnyandersson:
Seems to work just fine, except for one thing.

The tooltips seem to lag a bit on the horizontally placed buttons, initially rendering behind the buttons to the right. And sometimes, if you move the mouse too quickly, they seem to get stuck (see the last tooltip in the gif below).

Not a complete dealbreaker, but it's not a very smooth and pleasant experience.
tools

Flodkvist:
**Looks good!**
Running origo 2.8.1 (and also this PR) on editor with related layers, tooltip in editor form get stuck while pointer is placed over tooltip. 
image

**Running this PR with the addition from @tonnyandersson that issue is resolved.**

Another thing (unrelated to this PR), is that there is a white border on the tooltip anchor(?) see the red circle in image above.
I´m guessing this is something old that has gone unnoticed since it´s often placed on a white background. Removing the border fixes this...
image

This might be better adressed in a new issue?
The main editor tools in the bottom of the screen does not have tooltips?
So a new issue adressing editor tooltips might be better?

tonnyandersson:
Perhaps we should view all the tooltips related stuff as a separate issue after all? In that case, I'm all for approving this PR as it stands right now.

Issue #1981, opened 2024-03-21
Make it possible to have more backgroud layers


huggla:
The number of background layers is limited by the width of the layerswitcher. Find a way to fix this limitation.
![image](https://github.com/origo-map/origo/assets/5260503/c2442277-9a54-4c23-b61d-40d786d6a01c)

Wrap the line of background icons?:

![image](https://github.com/origo-map/origo/assets/5260503/c3e8663e-be42-418e-98d9-6dc4cd11f2e3)

Option to have a dropdown for backgrounds (would also make it possible to add text description)?:
![image](https://github.com/origo-map/origo/assets/5260503/8e0feb71-80e5-422d-8e5c-96b23e194498)

Grammostola:
Dropdown would need to be dropup as the legend control's background layer pane is in the bottom of the window and then it would cover parts of the legend. That might and might not look problematic.  The alternative is the way your last image shows , that the pane grows vertically to accomodate the dropdown. That might look a bit funny if there's a longer list of background layers I think.

If the line simply grows up in the second picture then that looks better. Maybe that's the simplest solution.

In a recent pr I was reportedly trying to solve this and we discussed it locally prior and "one button and a chooser that appears" is a popular thought, a bit like what maps.google.com does when the background layer button is hovered over. In order to fit something like that in the Origo legend ui something would need to stretch, though. A new element could be introduced above the background layer pane and the one above, with all the background layer choices and which would disappear when a choice was made. Can't tell for sure that that would look any better than your second image though. I kind of like the second image - wrap - for its simplicity.

tonnyandersson:
My first thought was also a new element, floating on top of the rest of the legend, kinda like the [Add Layer button](https://github.com/origo-map/origo/pull/1685#issuecomment-1437975188).

But the wrap thing seems like the best option here. What it lacks in looks and elegance, it makes up for in simplicity and ease of use and implementation.

huggla:
All that is needed for the wrap solution is to remove height:50px and add max-width:300px to the top div for the background buttons. But in which source file do I find the code to change?

tonnyandersson:
[src/controls/legend.js, line 701](https://github.com/origo-map/origo/blob/1e51a6ca46683a41d3ed03c86473431d12eab62d/src/controls/legend.js#L701), I presume.

PR #1985, opened 2024-03-25
Make it possible to have more backgroud layers

huggla:
Makes the background layer buttons wrap. https://github.com/origo-map/origo/issues/1981
bild

### DIFF
--- a/src/controls/legend.js
+++ b/src/controls/legend.js
@@ -711,7 +711,7 @@ const Legend = function Legend(options = {}) {
         components: mainContainerComponents,
         style: {
           'max-height': `${maxHeight}px`,
-          'width': 'min-content'
+          width: 'min-content'
         }
       });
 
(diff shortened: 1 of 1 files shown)


Grammostola:
Hi, nicely done. In the case of a legend control with the following prop:
```json
        "contentStyle": {
          "width": "350px"
        }
```
then the show active layers and minimize legend buttons should probably be situated in the bottom right, like where they are without a multirow of background map buttons?

huggla:
Thanks for your feedback. I think it should be fixed now.

> Hi, nicely done. In the case of a legend control with the following prop:
> 
> ```json
>         "contentStyle": {
>           "width": "350px"
>         }
> ```
> 
> then the show active layers and minimize legend buttons should probably be situated in the bottom right, like where they are without a multirow of background map buttons?
> 
> image

huggla:
I've changed the look. Better?

bild

bild

johnnyblasta (approved):
LGTM

huggla:
When I merged the original PR the hide and close buttons were perfectly aligned with the background layer buttons. But when I checked on a different computer a day later the alignment was off and I submitted a fix. Now, a few days later, the alignment is once again off by a few px. I have disabled cache in my browsers. Any ideas what causes this? Is it better to set margins in em/rem instead of px?

PR #1994, opened 2024-04-08
Make background buttons and close buttons in legend align on all browsers

huggla:
Style changes to make background buttons and close buttons in legend align on all browsers. Also removes some minor blank spaces.

bild

### DIFF
--- a/src/controls/legend.js
+++ b/src/controls/legend.js
@@ -652,7 +652,7 @@ const Legend = function Legend(options = {}) {
       visibleOverlaysCmp = VisibleOverlays({
         viewer, cls: `${contentCls} hidden`, style: contentStyle, labelOpacitySlider
       });
-      const baselayerCmps = [toggleGroup];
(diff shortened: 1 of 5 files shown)


tonnyandersson:
The close button still seems to be slightly misaligned, and it's not just that it's larger. 

image

huggla:
> The close button still seems to be slightly misaligned, and it's not just that it's larger.
> 
> image

What browser and OS are you using?

tonnyandersson:
Edge/Windows.

huggla:
I've updated the style to align a little better on Edge/Windows. If I nudge the close button more downwards it will look bad on Firefox.

tonnyandersson:
@huggla What does the firefox say if you set `margin: "3px 0"` instead of the current bottom only margin on the close button element?

huggla:
That looked better. Thanks! I changed px to rem though.

tonnyandersson (approved):

PR #1998, opened 2024-04-19
Fix missing typenames for QGIS Server sources

### LINKED ISSUE #1997: Typenames missing when using filterType "qgis"
**Description**
After the most recently merged PR#1536 - [Add filterType source param and use QGIS EXP_FILTER when param is qgis](https://github.com/origo-map/origo/pull/1536) - typenames are not included correctly when fetching features using IDs.

**To Reproduce**
1. Add a source with the option "filterType" = "qgis"
2. Search (using search option 1 to search a map layer) within a layer using the above source.
3. Press the makeSelection button to open the selection in the infoWindow
4. The call to fetch feature fails because of missing typename prefixes.

**Expected behavior**
The typenames were supposed to be included so that QGIS Server can return features...

**Additional context**
PR incoming...

MattiasSp:
Fixes #1997.

Corrects a mistake that caused typenames to not be prepended when fetching features from a QGIS Server source using IDs.

### DIFF
--- a/src/layer/wfssource.js
+++ b/src/layer/wfssource.js
@@ -204,8 +204,8 @@ class WfsSource extends VectorSource {
     if (ids || ids === 0) {
       switch (this._options.filterType) {
         case 'qgis': {
-          const idArray = ids.toString().split(','); // Split to array
-          idArray.map(id => {
+          let idArray = ids.toString().split(','); // Split to array
+          idArray = idArray.map(id => {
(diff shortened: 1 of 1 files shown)


steff-o (approved):
LGTM. 

Don't have a Qgis server to test on though, but the request seems legit and it doesn't crash or interfere with geoserver-style filters.

I will merge soon, giving others some more time to speak up,

PR #2002, opened 2024-05-02
Moved attachments and related tables from featureinfo to SelectedItem

### LINKED ISSUE #1975: Add related tables and attachments to SelectedItem
When using _MultiSelect_ plugin, attributes from related tables and attachments are missing. This creates a inconsistent result depending on if infowindow was displayed using featureInfo or MultiSelect. This is because the code for adding the related tables and attachments attributes are hidden inside the featureInfo implementation, which MultiSelect can not use as it is not exposed in api.

There should be an easy way to create attachments and related tables content from api. It could be done in numerous ways, for instance:
1. By adding a function to SelectedItem: addAsyncContent or create a factory method that do both ```new SelectedItem()``` and adds async content.
2. Expose the methods in featureInfo that adds the related tables and attachments.

Probably the most flexible and easy to maintain solution would be adding a function to SelectedItem.

Copy paste all related tables and attachment code to MultiSelect, which obviously is not appealing. It is quite a lot of code.

Utilize one of ShowInfo or  ShowFeatureInfo api function on featureInfo. But that does only support a new complete set of features while MultiSelect also can add to current selection. Also MultiSelect uses SelectionManager and not featureInfo. 

**Additional context**
The reason why the content for attachments and related tables are not added in the SelectedItem constructor is that adding that content is an async operation. When I wrote (yes it was me) attachments and related tables I settled for adding the content only when using featureInfo. I regret this today.

steff-o:
Fixes #1975 by adding the _async createContentAsync()_ method on _SelecedItem_ class. While doing that all code that creates the async content has been moved from _featureinfo.js_ to _getfeatureinfo.js_ while keeping the functionality intact.

For "standard" use of Origo, this PR will not change anything. The main benefit is that is is now possible to create SelectedItems containing async content and display them in InfoWindow from api or plugins, as the _SelectedItem_ class is exposed in the global Origo object.

### DIFF
--- a/src/featureinfo.js
+++ b/src/featureinfo.js
@@ -12,9 +12,7 @@ import StyleTypes from './style/styletypes';
 import getFeatureInfo from './getfeatureinfo';
 import replacer from './utils/replacer';
 import SelectedItem from './models/SelectedItem';
-import attachmentclient from './utils/attachmentclient';
-import getAttributes, { getContent, featureinfotemplates } from './getattributes';
-import relatedtables from './utils/relatedtables';
+import { getContent, featureinfotemplates } from './getattributes';
 
 const styleTypes = StyleTypes();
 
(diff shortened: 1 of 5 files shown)


Flodkvist:
Tried running this on a related tables configuration, looks good to me.

Issue #1927, opened 2023-12-07
Sharemap doesn't work when configuration is defined inline.


huggla:
The sharemap control is able to save the mapstate, but the given link doesn't retrieve it. Works when config is stored in separate json file.
```
<html>
	<head>
		<link href='css/style.css' rel='stylesheet'>
		<script src="js/origo.min.js"></script>
		<script>
			var json={
				"controls": [
					{
						"name":"mapmenu",
						"options":{
							"isActive":true
						}
					},
					{
						"name":"sharemap",
						"options":{
							"title":"Dela karta",
						        "storeMethod": "saveStateToServer",
							"serviceEndpoint": "../origoserver/mapstate"
						}
					}
				],
				"projectionCode": "EPSG:3857",
				"extent": [
					-20026376.39,
					-20048966.1,
					20026376.39,
					20048966.1
				],
				"center": [
					1770000,
					8770000
				],
				"resolutions": [
					1222.99
				],
				"layers": [{
					"name": "osm",
					"title": "OpenStreetMap",
					"type": "OSM",
					"visible": true
				}]
			};
		</script>
	</head>
	<body>
		<div id="app-wrapper"></div>
		<script>
			var origo = Origo(json);
		</script>	
	</body>
</html>
```

huggla:
Could someone please take a look at this issue? It breaks some features in our Origo administration tool.

huggla:
Yes, still an issue.

To reproduce:
1. Load the html above (no .json-file).
2. Pan the map to new location.
3. Create url with the sharemap tool "Dela karta".
4. Open the url and notice that the map is loaded at the default location, not the location you paned to in step 2. 

The created url from step 3 works if you change the html to load the config from separate .json-file

PR #2016, opened 2024-05-24
Fix map name parsing in permalink store getState() function

### LINKED ISSUE #2014: Change permalinkstore state.map split param
Sometimes there can be more than one dot in the path to the configuration file so it would be a better solution to split getMapName() on ".json" instead of on "." for state.map in permalinkstore.

mulfvik:
Fixes #2014

### DIFF
--- a/src/permalink/permalinkstore.js
+++ b/src/permalink/permalinkstore.js
@@ -104,7 +104,7 @@ permalinkStore.getState = function getState(viewer, isExtended) {
       .join();
   }
   if (viewer.getMapName()) {
-    state.map = viewer.getMapName().split('.')[0];
+    state.map = viewer.getMapName().split('.json')[0];
   }
 
   Object.keys(additionalMapStateParams).forEach((key) => additionalMapStateParams[key](state));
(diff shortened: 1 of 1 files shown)


tonnyandersson:
Is it 100% certain that the map name contains '.json'? What if someone decides to use .geojson or .txt or something?

Perhaps it would be safer to split at the last occurence of '.'?

Not asking you to change this, just raising the question.

mulfvik:
Sure, why not

huggla:
I don't think this works with inline configuration. Related to this issue https://github.com/origo-map/origo/issues/1927

mulfvik:
> I don't think this works with inline configuration. Related to this issue #1927
> 
> Nevermind, I think I mixed things up. This is reading from mapstate-json not config-json, right?

Is there (still?) an issue?

huggla:
If there is a problem after https://github.com/origo-map/origo/issues/1927 has been fixed then I'll open a new issue. I think this should be ok to merge.

tonnyandersson (approved):

PR #2032, opened 2024-06-27
feature: expose layertype

### LINKED ISSUE #2031: Expose layertype
A plugin may bring support for a new layer type and therefore it would be useful to have the layertype exposed by Origo.

johnnyblasta:
Closes #2031


Grammostola:
mulfvik wrote in the related issue:
> Tested this and it seems, due to that Origo loads config before plugin(?), that it would take some promiseAll operation to load them when them all is finished.

> A plugin may bring support for a new layer type
Is what you wrote in the issue as a reason for exposing Layertype.


I at least agree that the purpose appears somewhat opaque. Is it to be able to add a layer type to Origo so that all parts of it and potentially other plugins can use the new layer type? (There would seem to be parts missing if that's the case like a method to add layerTypes). As mulfvik deduces the plugins are loaded after Origo so Origo wouldn't be able to use the new type during init. This all makes me think this is not your indended scenario.

Or is it to access Origo's existing layer type definitions in case a plugin wants to do something with them, possibly extend one?

johnnyblasta:
If the order of loading of types isn't resolved somehow, the purpose of this PR does not resolve the intended issue.

johnnyblasta:
Closing this since it won't work for intended purpose for plugins.
To be able to add layers type is still something that might be useful, but has to be reworked a bit to be useful.

PR #2041, opened 2024-09-11
fix: Left toolbar scrollable

### LINKED ISSUE #2021: The toolbar needs to know the size of the browser
If a user have many tools expanded at the same time, some icons will not appear in the browser, the toolbar expand without knowing it hit the bottom of the browser.

The toolbar should sense the size of the browser and do something when it hits the bottom, like the footer and the legend. I'm not really sure what the best solution would be, maybe a scrollbar, maybe a new bar on the right, but I'm not convinced by either. All I know is that the user should be able to see all available tools in the toolbar.

Even if we move the measure tool and the selection tool to the bottom just like the draw and editor control, as we mentioned in issue #1409, I believe the problem remains since many controls can be placed in the left toolbar. Users can have different resolutions, zoom settings and sizes of screens, browsers and devices and so on. This is something that should work much better than it does today.

toolbar_expanded

jokd:
Fixes #2021 
Makes left toolbar scrollable if too long for browser window

### DIFF
--- a/scss/_viewer.scss
+++ b/scss/_viewer.scss
@@ -4,6 +4,9 @@
   flex-flow: column;
   left: 0;
   top: 0;
+}
+
+#o-tools-left {
   width: 100%;
   scrollbar-width: none;
   overflow: auto;
--- a/scss/_viewer.scss
+++ b/scss/_viewer.scss
@@ -2,11 +2,11 @@
   display: block;
   display: flex;
   flex-flow: column;
-}
-
-.o-tools-left {
   left: 0;
   top: 0;
+  width: 100%;
+  scrollbar-width: none;
+  overflow: auto;
 }
 
 .o-toolbar-maptools {


steff-o:
What is the rationale of using an id-selector instead of class? Wouldn't that cause problems if a page hosts several origo applications using different app-wrapper containers? 

While it currently probably is not possible to actually host several origo-applications in the same document, I think it unnecessary to make it even harder to make it possible. Not that I have no idea why anyone would like to host several instances in the same document, but anyway. Overview map maybe?

jokd:
> What is the rationale of using an id-selector instead of class? Wouldn't that cause problems if a page hosts several origo applications using different app-wrapper containers?
> 
> While it currently probably is not possible to actually host several origo-applications in the same document, I think it unnecessary to make it even harder to make it possible. Not that I have no idea why anyone would like to host several instances in the same document, but anyway. Overview map maybe?

Due to specificity and a bit of laziness. The .box.transparent css rule sets overflow:visible which we don't want and to ensure that rule is overridden I used the id selector. Otherwise we need to change or create new classes for the toolbar element.

steff-o:
The id-selector issue aside (it would work with multiple identical ids, it just is "wrong" to have multiple identical id:s and it is just a hypothetical case anyway), I think it is pretty limited and hard to discover. There is no indication what so ever that it is possible to scroll the toolbar and it only works with scroll wheel, which makes it useless on phones which probably is the situation where it would be most useful.

Just for fun I tried to show the toolbar but it appeared on the far right of the screen and was not scrollable by dragging the bar, possibly due do OL:s interactions taking precedence.

Also when the screen gets small the attributions control changes from a bar at the bottom to an icon at bottom left, which will overlay the toolbar. The scrollable panel would have to compensate for that area as well, even if it is possible to scroll the last tool above the "i" icon as there seems to be an empty row last in the toolbar.

image

Ideally there would appear clickable scroll up and scroll down arrows above and below the tool bar and the tool buttons themselves should be possible to drag in order to scroll the toolbar, but that will take implementation to a completely new level.

jokd:
>it only works with scroll wheel, which makes it useless on phones which probably is the situation where it would be most useful.

Now I recall why the tooltips wouldn't work the last time I looked into it. pointer-events can't be auto when the div is broader than the toolbar (which is necessary to show the tooltips) and keeping `pointer-events: none` will prevent touch scrolling. I close this now and maybe will get back to it later.

PR #2046, opened 2024-09-11
Always show infowindow

### LINKED ISSUE #2026: SelectionManager can not display InfoWindow if featureInfo is not configured to use infowindow
**Describe the bug**
If featureinfo is configured to use `overlay` or `sidebar` infowindow is not displayed when a selection is made with SelectionManager. As multiselect uses SelectionManager's api to display the selection, no infowindow will be show.

**To Reproduce**
1. Configure featureinfo to use overlay or sidebar
2. Add some features to SelectionManager using api or MultiSelect
3. See error: the selection is made in the map, but no infowindow is displayed

**Expected behavior**
Infowindow or at least overlay or sidebar should be visible.

**Additional context**
This may or may not be a bug. Displaying infowindow is explicitly prevented i SelectionManager when featureinfo is not configured to use infowindow. This has been in the code since the first version of SelectionManager, but I can not see the reason why. The only reason I can see is that allowing InfoWindow would create inconsistent look if selection is made with featureinfo or multiselect. But preventing the InfoWindow to be displayed does not really help in that case. It would then have been better to actually use overlay or sidebar, but due to implementation that would require a major rewrite of featureinfo.

steff-o:
Fixes #2026 by making SelectionManager to display the selection in InfoWindow even if the "infoWindow"-setting is `overlay` or `sidebar`.

This will result in api-calls to SelectionManager always showing the result in infoWindow while featureInfo will obey the infoWindow-setting. It may seem a bit confusing but the old behavior was to only display the selection in the map and not show any infoWindow at all during the same circumstances.

This will only affect calls directly to SelectionManager using the api, e.g. from the MultiSelect-plugin.

### DIFF
--- a/src/selectionmanager.js
+++ b/src/selectionmanager.js
@@ -32,7 +32,6 @@ const Selectionmanager = function Selectionmanager(options = {}) {
   let highlightedFeatures = [];
 
   const multiselectStyleOptions = options.multiSelectionStyles || styleTypes.getStyle('multiselection');
-  const isInfowindow = options.infowindow === 'infowindow' || false;
   const infowindowManager = options.infowindowOptions && options.infowindowOptions.listLayout ? infowindowManagerV2 : infowindowManagerV1;
 
   function alreadyExists(item) {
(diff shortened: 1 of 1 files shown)


Grammostola (approved):
lgtm

PR #2051, opened 2024-09-23
deps: OL v10.2.0 and some other deps

### LINKED ISSUE #2049: OpenLayers v10.2.0
New version

jokd:
Fixes #2049 and upgrades some other dependencies

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,26 +1,26 @@
 {
   "name": "origo",
-  "version": "2.9.0",
+  "version": "2.10.0-dev",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "origo",
-      "version": "2.9.0",
+      "version": "2.10.0-dev",
       "license": "BSD-2-Clause",
       "dependencies": {
-        "@glidejs/glide": "^3.6.1",
-        "awesomplete": "^1.1.5",
+        "@glidejs/glide": "^3.6.2",
+        "awesomplete": "^1.1.7",
         "cuid": "^2.1.8",
         "downloadjs": "^1.4.7",
         "drag-drop-touch": "^1.3.1",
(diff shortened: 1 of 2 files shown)


steff-o:
Haven't tested yet, but if I remember correctly, there are some breaking changes that might affect us. Are you sure that no code changes are necessary?

steff-o:
Still haven't tested ...

Looks like breaking changes are few: https://github.com/openlayers/openlayers/blob/main/changelog/upgrade-notes.md

It was _getFeaturesInExtent()_ in vector tiles I was uncertain about, but I can't find any direct references to _getFeaturesInExtent()_ in our code but OL seems to use it internally in other calls as it is a part of the origo bundle. But I assume OL will be smart enough to figure out what to do when it is called internally when it  is a VectorTile layer.

Grammostola:
On a tangent `ol/source/imageTile` has appeared, so far with the intent of removing `ol/source/tileImage` if I interpret things correctly, (which may not be the case). Print uses `ol/source/tileImage`. So far so good in other words.

I aim to test this with a few different layer types and most controls during the day.

steff-o:
Tested some basic stuff (WMS, WFS, Edit WFS incl. snapping, measure, print, featureInfo) in order to activate as much OL interactions as possible. Looks OK. Only concern is that OL 10.2.1 is out now... Just a minor bugfix regarding tiles.

I don't have any test setup for VectorTiles so I haven't tested that.

jokd:
I changed to OL 10.2.1 and changed the version in package.json to 2.9.1-dev to show that this is a dev version of the upcoming release and not the latest release.

steff-o (approved):
Did some testing on searchList (awesomplete) and carousel (glide) as well as rerun tests for 10.2.1. Looks good.

PR #2053, opened 2024-09-27
feature: Extend layerSpecificExport for APIs

### LINKED ISSUE #2040: Extension of layerSpecificExport to allow more config options
**Description**
The layerSpecificExport options for the infoWindow are great, but could be more flexible with the request specifications.

Adding the option to include URL-parameters and choose between making a POST- or a GET request would go a long way to making this feature more general and make it possible to export selected features to many different types of APIs.

**Proposed solution**
Adding a bunch of config parameters to the layerSpecificExport which control how the requests are made.

**Alternatives considered**
Adding more hardcoded export modes other than simpleExport and layerSpecificExport would make the configuration simpler as each mode could be better tailored to a specific type of API. But I prefer the proposed solution because hard coded modes would make the code more bloated and repeated, and would also require the code to be updated to match any changes to the APIs.

MattiasSp:
Fixes #2040.

The PR includes several improvements to the layerSpecificExport handling to make the configuration more flexible and adaptable to more different types of APIs. It does not include authentication in any way, but that would be a useful next addition.

- Allow setting `exportedFileName` and `button` per layer or specific URL
  - The handling of export options was not entirely consistent, but now these options (and all the others except `url`) can be set as defaults per layer, or overriden for specific export URLs.
- New export option: `urlParameters`
  - Add URL parameters as key-value pairs
  - Fill parameter values with values from selected objects' attributes by specifying an object with an `attribute` property: `{ "attribute": "objekt_id", "separator": ";"}`. The separator character can be customized but uses semicolon by default.
  - Set a parameter's value to `{{no_value}}` to add the parameter without the equals sign or value.
  - The example config below will result in e g the following url: https://my.api.se/postSomething?param1=value1&featureid=1;3;12&silent
<pre>
{
  "url": "https://my.api.se/postSomething",
  "urlParameters": {
    "param1": "value1",
    "featureid": { "attribute": "id", "separator": ";"},
    "silent": "{{no_value}}"
  },
  "attributesToSendToExport": ["id","comment"],
  "button": {
    "buttonText": "Post IDs"
  }
}
</pre>
 - New export options `requestMethod` (**`POST_JSON`** | `OPEN` | `GET`) and `displayExportResponse` (`true` | **`false`**) added
   - `POST_JSON` is the default option and posts JSON the usual way as the layerSpecificExport works.
   - `OPEN` opens the URL in a new window.
   - `GET` makes a GET request
     - Includes any `urlParameters` specified (but not `attributesToSendToExport` of course)
     - If `displayExportResponse` is set to `true` (for the layer or the specific url) the request's response will be displayed in the infowindow. Supports images and text (if from the same domain or otherwise allowed by CORS restrictions). Defaults to false.
 - Added a material design icon `article` because it looks nice for opening reports etc.


steff-o:
A lot of things to test, so I haven't had the time to test yet as it requires some hefty set up to actually test the export result handling. Guess I will just test that old functionality is not broken and configuration is backwards compatible.

But the silent parameter, why not just set it to _null_ or empty string? Null could be sent as completely empty and empty string "PARAM=" if there is a difference between "PARAM" and "PARAM=" in the backend. 

 The {{}} notation is not really intuitive and can potentially interfere with a future  variable substitution syntax by disallowing the variable name "no_value" (whoever calls a variable that)

MattiasSp:
> But the silent parameter, why not just set it to _null_ or empty string? Null could be sent as completely empty and empty string "PARAM=" if there is a difference between "PARAM" and "PARAM=" in the backend.
> 
> The {{}} notation is not really intuitive and can potentially interfere with a future variable substitution syntax by disallowing the variable name "no_value" (whoever calls a variable that)

We have a backend that cares about the trailing equals sign, so I changed it to either _null_ or empty string which results in &silent or &silent= respectively.

steff-o:
Why not use the json literal `null` for representing null or an object with a property `"empty": true` ? I think it would be less confusing than using magic strings.

If `URLSearchParams` does not serialize null parameters as desired a temporary replacement could be added. By doing so the magic string can be anything as it is not visible to the user.

steff-o:
Finally got my act together to test this. It seems to do what it is supposed to do without breaking how it used to work.

However `displayExportResponse` only works for the traditional infowindow. If `displayExportResponse` is set to true and `"listLayout": true,` it results in a nasty console error. Since the users can  not switch by themselves this is a configuration problem, but as the list view is gaining popularity I think it should work there as well or at least be clearly documented that exactly that setting does not apply to the list view if that was the intention.

MattiasSp:
Thanks for testing! :)

> However `displayExportResponse` only works for the traditional infowindow. If `displayExportResponse` is set to true and `"listLayout": true,` it results in a nasty console error. Since the users can not switch by themselves this is a configuration problem, but as the list view is gaining popularity I think it should work there as well or at least be clearly documented that exactly that setting does not apply to the list view if that was the intention.

I agree, not considering the other infowindow modes was an oversight. I'll look into supporting them or at least not causing a console error promptly.

PR #2055, opened 2024-10-04
fix: expose layertype

### LINKED ISSUE #2031: Expose layertype

johnnyblasta:
Found out that this will work with plugins that has a layer type which is missing in Origo and want to add it.
The key is that the layer type should be added before starting Origo.

Closes #2031

### DIFF
--- a/origo.js
+++ b/origo.js
@@ -29,6 +29,7 @@ import 'drag-drop-touch';
 import permalink from './src/permalink/permalink';
 import * as Loader from './src/loading';
 import Spinner from './src/utils/spinner';
+import layerType from './src/layer/layertype';
 
 const Origo = function Origo(configPath, options = {}) {
   /** Reference to the returned Component */
@@ -172,5 +173,6 @@ Origo.Loader.show = Loader.showLoading;
 Origo.Loader.hide = Loader.hideLoading;
 Origo.Loader.withLoading = Loader.withLoading;
 Origo.Loader.getInlineSpinner = Spinner;
+Origo.layerType = layerType;
 
 export default Origo;


Grammostola:
Would have been interesting to have partaken in your meeting as this PR is clearly for the initiated : )

It looks just like the old one.

 Origo (origo, origo.js) makes bits available that plugs can utilize to add a layer type. After origo has been instantiated, reasonably. 

Yet no doubt you've solved the riddle so it'll be interesting to hear how.

johnnyblasta:
Yes, it's the same as the old PR.

The trick we missed the last time was to add the new layer type before starting Origo, since it's to late to add it when Origo already is started and the load event is dispatched which the plugin listens to get started.

This wasn't perhaps the neatest solution and we discused and tested other solutions with triggering more events in the startup before the load event, but didn't get it to work as we wanted.

And yes you probably missed an opportunity to dig down in the loading and dispatching of events in Origo.

Grammostola:
Aha, as in
```js

const origo = Origo('index.json');
origo.layerType.threeDtile = ...

origo.on('load', function (viewer) {
```

Indeed, I already had another meeting (a friday afternoon :) ). Oh well, I didn't have time to answer the doodle either, you snooze you lose.

This seems like a realistic solution at any rate.

johnnyblasta:
> Aha, as in
> 
> ```js
> Origo.layerType.threeDtile = ...
> const origo = Origo('index.json');
> origo.on('load', function (viewer) {
> ```
> 
> Indeed, I already had another meeting (a friday afternoon :) ). Oh well, I didn't have time to answer the doodle either in time, you snooze you lose.
> 
> This seems like a realistic solution at any rate. (Edit. I think I figured it out)

Yes, that is correct, the layertype must be declared before kicking of Origo.

steff-o:
Yes, ideally we wanted a neater solution where a plugin could add itself with just one line of code. This way it is a two step operation to add the 3D layer plugin: first add the layer type before origo is initialized, then add the plugin part as usual in the load event.

Emitting more events from origo could have solved that by emitting one event just after the origo component is created but before the viewer is initialized and have the plugin listen to that event and in that handler attach to more events. Problem was that we needed to rework the start up procedure too much and that would be risky, especially if we were to accommodate for async event handlers.

We also discussed adding callbacks to origo() options, but that was scrapped for pretty much the same  reasons as events plus that it wasn't as cool.

This solution, as raw it may seem, does however have some merits: it makes it possible to add functionality to the Origo-library and then use it. Not all plugins needs to add functionality, and not all added functionality needs a plugin to work.

On a personal note, I would have preferred a solution that only exposes an layertypes.add() method to avoid exposing internal implementation details. But I won't fight for it as it is just a personal preference which I can't say is better than anyone else's personal opinion.

PR #2059, opened 2024-10-21
fix: make print legend layer order match legend layer order

### LINKED ISSUE #2057: Print legend order is reversed
**Describe the bug**
The legend in print is in reversed order compared to the legend in map view.

**To Reproduce**
1. Memorize the layer order in the legend
2. Click print in top right menu
3. Behold, the layers in the print view is reversed compared to that of the map view

**Expected behavior**
The legend should be in the same order as in the map view

**Screenshots**
Map view:
image

Print view:
image

**Additional context**
Problem is probably in print-legend.js where the layers array is iterated:
```javascript
overlays.forEach((layer) => {
  if (!layer.get('drawlayer')) {
    overlayEls.push(LayerRow({ layer, viewer }));
  }
});
const layerListCmp = Component({
  async render() {
    const startTime = new Date().getTime();
    const content = await overlayEls.reduce(async (acc, item) => await acc + await item.render()  , '');
    console.log(`Took ${new Date().getTime() - startTime} ms`);

    return `<ul id="${this.getId()}" class="list">${content}</ul>`;
  }
});
```
The overlays array is in the reverse order of index.json, in order to make the last defined layer be drawn first and thus become the lowest layer. The legend control reverses this order correctly to reflect the draw order, but print control does not.

Additionally, the row 
```javascript
    const content = await overlayEls.reduce(async (acc, item) => await acc + await item.render()  , '');
```
makes the awaiting sequential as acc is awaited first. By simply switching order to `await item.render() + await acc`, the order will be reversed and the reduce iteration will be performed in parallel, potentially reducing (no pun intended) the rendering time.

Grammostola:
Aims to fix #2057

### DIFF
--- a/src/controls/print/print-legend.js
+++ b/src/controls/print/print-legend.js
@@ -286,7 +286,7 @@ const LayerRows = function LayerRows(options) {
         async render() {
           const rowPromises = overlayEls.map((item) => item.render());
           const rows = await Promise.all(rowPromises);
-          return `<ul id="${this.getId()}" class="list">${rows.reverse()}</ul>`;
+          return `<ul id="${this.getId()}" class="list">${rows.reverse().join('')}</ul>`;
         }
       });
       return `
--- a/src/controls/print/print-legend.js
+++ b/src/controls/print/print-legend.js
@@ -284,8 +284,9 @@ const LayerRows = function LayerRows(options) {
       });
       const layerListCmp = Component({
         async render() {
-          const content = await overlayEls.reduce(async (acc, item) => await acc + await item.render(), '');
-          return `<ul id="${this.getId()}" class="list">${content}</ul>`;
+          const rowPromises = overlayEls.map((item) => item.render());
+          const rows = await Promise.all(rowPromises);
+          return `<ul id="${this.getId()}" class="list">${rows.reverse()}</ul>`;
         }
       });
       return `


src/controls/print/print-legend.js
```diff
@@ -284,8 +284,9 @@ const LayerRows = function LayerRows(options) {
       });
       const layerListCmp = Component({
         async render() {
-          const content = await overlayEls.reduce(async (acc, item) => await acc + await item.render(), '');
-          return `<ul id="${this.getId()}" class="list">${content}</ul>`;
+          const rowPromises = overlayEls.map((item) => item.render());
+          const rows = await Promise.all(rowPromises);
+          return `<ul id="${this.getId()}" class="list">${rows.reverse()}</ul>`;
```
steff-o, line 289: The implicit rows.reverse().toString() introduces an extra comma as it calls string.join() with comma as separator.

steff-o (changes requested):
I really like the change to get rid of the async reduce, it is really hard to read.

However this solution introduces an annoying comma sign between the list elements:
image
image

src/controls/print/print-legend.js
Grammostola: Well spotted

steff-o (approved):

PR #2061, opened 2024-10-23
feature: multiple choice checkbox in editor

### LINKED ISSUE #1518: Multiple values in editorform
When editing a layer it would be good if there was a possibility to choose multiple values from a dropdown list or be able to tick multiple checkboxes and store in the same attribute.

For example if the attribute is country and the options are:
•	Sweden
•	Norway
•	Denmark
•	Finland
Then being able to select one or many values. 

In the dialogue-plugin there is an excellent solution to the problem, with checkboxes. Here you can also add a text value if no option in the value list matches the desired value “for example Iceland”. 
This functionality in the core would be superb.  

**Additional context**
bild

johnnyblasta:
Closes #1518 
Adds a multiple choice checkbox to the editor with a possibility to have a free text choice.
image

Configurable like this:
```
{
        "name": "atgardsforslag",
        "title": "Åtgärdsförslag: ",
        "type": "checkbox",
	"options": [
		"Kronhöjning minst >4.6m",
		"Kronhöjning minst >3.2m",
		"Utrymme beskärning ",
		"Borttagnings av döda grenar",
		"Återhamling",
		"Hammling",
		"Leverans högstubbe",
		"Kronretireringbeskärning",
		"Formträdbeskärning",
		"Uppbyggnadsbeskärning",
		"Säkerhetsbeskärning",
		"Avlastbeskärning",
		"Borttagnings av vatten skott",
		"TRAQ 2",
		"TRAQ 3; Klätterandebesiktning",
		"TRAQ 3; Resistograph/tomograph",
		"TRAQ 3; Advencerad Rotbesikning",
		"Annat:textbox"
	]  
}
```


johnnyblasta:
I think I prefer `checkboxgroup` if it should be a new input type.
I can take a look at reworking it, so it's less hokus pokus in the implementation.

johnnyblasta:
- Introduced a new input type `checkboxgroup`
- Changed the configuring syntax, so that it's not magical.
- Removed the hardcode for separator.
- Added option for the freetext field so it's not harcoded and makes it possible to add more than one.
- Added the possibility to have the text and value of the choice to be different.

Example configuration:
```
{"name": "atgardsforslag",
 "title": "Åtgärdsförslag: ",
 "type": "checkboxgroup",
 "separator": ";",
 "freetextOptionPrefix": "freetext_option:",
 "freetextOptionValueSeparator": "=",
 "options": [
    { "text": "Kronhöjning minst >4.6m", "value": "Kronhöjning minst >4.6m" },
    { "text": "Kronhöjning minst >3.2m", "value": "Kronhöjning minst >3.2m" },
    { "text": "Utrymme beskärning ", "value": "Utrymme beskärning" },
    { "text": "Borttagnings av döda grenar", "value": "Borttagnings av döda grenar" },
    { "text": "Återhamling", "value": "Återhamling" },
    { "text": "Hammling" },
    { "text": "Leverans högstubbe", "value": "Leverans högstubbe" },
    { "text": "Kronretireringbeskärning", "value": "Kronretireringbeskärning" },
    { "text": "Formträdbeskärning", "value": "Formträdbeskärning" },
    { "text": "Uppbyggnadsbeskärning", "value": "Uppbyggnadsbeskärning" },
    { "text": "Säkerhetsbeskärning", "value": "Säkerhetsbeskärning" },
    { "text": "Avlastbeskärning", "value": "Avlastbeskärning", "type": "textbox" },
    { "text": "Borttagnings av vatten skott", "value": "Borttagnings av vatten skott" },
    { "text": "TRAQ 2", "value": "TRAQ 2" },
    { "text": "TRAQ 3; Klätterandebesiktning", "value": "TRAQ 3; Klätterandebesiktning" },
    { "text": "TRAQ 3; Resistograph/tomograph", "value": "TRAQ 3; Resistograph/tomograph" },
    { "text": "TRAQ 3; Advencerad Rotbesikning", "value": "TRAQ 3; Advencerad Rotbesikning" },
    { "text": "Annat", "type": "textbox" }
  ]  
}
```

src/controls/editor/edithandler.js
```diff
@@ -1235,6 +1300,14 @@ function editAttributes(feat) {
           } else {
             alert('Villkor verkar inte vara rätt formulerat. Villkor formuleras enligt principen change:attribute:value');
           }
+        } else if (obj.type === 'checkboxgroup') {
+          if (obj.options && obj.options.length > 0 && obj.val) {
+            const separator = obj.separator ? obj.separator : ';';
+            obj.val = obj.val.split(separator);
+          }
+          obj.isVisible = true;
+          obj.elId = `input-${obj.name}`;
```
steff-o, line 1309: This is not related tables proof. Id should be truly unique, e.g. by including layer name like in the else clause:
```obj.elId = `input-${currentLayer}-${obj.name}`;```
Not sure why same code is in several if:s though. maybe my fault.

steff-o (changes requested):
No doubt that this will be a great addition and the text and value approach is a great feature, however it seems like we don't share the same vision when it comes to the "other" values. You may have a stronger case, but here is my take on it:
- It could possibly simplify backend operations if all values including the others are stored in the same way in the database. Your approach is robust when it comes to roundtripping in Origo, but if you want a backend integration it will be necessary to parse the string in the database. If all values where stored using only a separator, any list splitting operation will do. I think it would be possible to implement it in that way if you don't require to match the value against a specific "other" label as described in my previous comment.
- The _text_ property on the "others" is a candidate for localization as a part of the GUI, not a part of the values, which makes me think it should be a localized string in editform. Also if made that way, it won't be necessary to store it in the database which makes previous point easier.

Do what you want with that, but I request a change on the id-creation.

steff-o:
... and tabbing does not highlight the checkboxes ...

johnnyblasta:
I think I fixed the unique id @steff-o requested, so take a look.
Added highlight for all inputs i form, since all seemed to be missing and also some labelling was missing or incomplete.

As for the the other value I think I will leave it as it is for now since it works and is flexible for possible use cases. I see there is possible ways to do this, for one would to save the values as a json object. Yes it could be localized if it only is a others field, but perhaps some will use it more creatively.

steff-o (approved):
LGTM.

Yes, serializing to JSON can be an option in the future when someone discovers the need for it. The current implementation seems to fulfil the current requirements.

PR #2067, opened 2024-11-01
feature: add another member to mask and cities

sebras:
I know that pull request #1966 also adjusted the transparency mask in order to properly highlight the relevant city. I can't figure out how one knows what coordinates to add for the mask. Maybe someone can help out with that? :)


tonnyandersson:
Hi, @sebras!

The origo-cities layer only contains official members of Origosamverkan, and according to the latest intel, Malmö is not an official member yet.

However, you are still more than welcome to flaunt your stadskarta in the Live examples section of the README if you want to. In that case, just drop the changes in the origo-cities file and we'll review this PR again.

PR #2068, opened 2024-11-01
feature: localize draw

### LINKED ISSUE #2065: Drawing disappears when show length/area or sections
When choosing to show length/area or sections in a drawing the drawing disappears from the map.
image
image

Grammostola:
Aims to fix #2065 (the long way around)

Some testing would be good, around measure and draw. I've done some testing and I have gotten most of the strings and numbers but do tell if there's something missing (or a crash, I've changed a number of function signatures but who likes `formatArea(undefined, 0, undefined, undefined, cat)`?)

 The editor appears to have parallel files and I haven't tested it.


johnnyblasta:
There is a little bit faulty on the translation.
On the types of points.
Check mark refers to a [example check icons](https://www.flaticon.com/free-icons/check-mark)
Cross mark (also known as 'X mark') refers to a [example cross icons](https://www.flaticon.com/free-icons/cross-mark)
Plus mark refers to a [example plus icons](https://www.flaticon.com/free-icons/plus-mark)
Also notice the blank space separation.

Grammostola:
Thanks for the feedback. I did experience difficulties with the English terms for "kors" and "kryss"-symbols. (Obviously "Ask an LLM" wasn't immediately the right move. Humans are still better at some things :P )

So 
```json
           "pointTypeCross": "Kors",
            "pointTypeCheckmark": "Kryss",
```

can become

```json
           "pointTypePlus": "Kors",
            "pointTypeX": "Kryss",
```

In English

```json
           "pointTypePlus": "Plus",
            "pointTypeX": "Cross",

```
What was that about a blank space separation?

johnnyblasta:
> 
> What was that about a blank space separation?

You had "Checkmark", but the correct would have been "Check mark".
If you don't use "mark" or "sign" as in your suggestion above there is no need for "särskrivning"

Grammostola:
Should be ok now.

johnnyblasta:
> Should be ok now.

Nah, seems to still be confusing between cross and plus.

Grammostola:
Please specify what you would like to see then, I did provide suggestions earlier : )

src/loc/en_US.json
```diff
@@ -80,7 +116,35 @@
...
+            "rotation": "Rotation",
+            "degreesRotation": "Degrees",
+            "marginForText": "Margin",
+            "frame": "Frame",
+            "margin": "Margin",
+            "point": "Point",
+            "pointTypePoint": "Point",
+            "pointTypeCircle": "Circle",
+            "pointTypeX": "X",
+            "pointTypePlus": "Cross",
```
johnnyblasta, line 143: "pointTypePlus": "Cross" should be "pointTypePlus": "Plus"

johnnyblasta (commented):
Here is what I meant was wrong.

src/loc/en_US.json
johnnyblasta: "Kors" works in swedish and "Plus" would also be appropriate in swedish, so I don't know which is best.

Issue #2062, opened 2024-10-24
Link Control:  Ability to be placed directly on the screen


dahlalex:
**Is your feature request related to a problem? Please describe.**
The current "Link" control can only be placed within the mapmenu. This could become a constraint when developers want to place link buttons directly on the screen without requiring a full mapmenu setup. 

**Describe the solution you'd like**
I propose enhancing the "Link" control to include an option that allows it to be placed directly on the screen, in addition to its existing functionality within the mapmenu. This would provide developers with the flexibility to:

- Use the control as it currently functions inside the mapmenu, or/and
- Add the control directly to the screen as a standalone link button, depending on the specific needs of the project.

This approach retains the original functionality of the "Link" control while adding the ability to use it in more flexible layouts.

*Here is a video that shows the potential enhancement:

![Animation](https://github.com/user-attachments/assets/aa40c28b-b700-4dd4-9ea4-d9bc435e1c9a)

PR #2069, opened 2024-11-06
feature: link control on screen

dahlalex:
Aims to fix [#2062](https://github.com/origo-map/origo/issues/2062)
Linked with [#328](https://github.com/origo-map/origo-documentation/pull/328)

### DIFF
--- a/src/controls/measure.js
+++ b/src/controls/measure.js
@@ -750,7 +750,7 @@ const Measure = function Measure({
         });
 
         measureButton = Button({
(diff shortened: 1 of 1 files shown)


johnnyblasta (commented):
Tested and works fine.
However why are there changes in measure.js in the PR?

johnnyblasta:
Seems like this misalignment is present on more buttons. 
Saw it on the draw button also.
Perhaps a separate issue is a good idea and try catch all the buttons that misalign.

Issue #2064, opened 2024-10-26
External url Control:  Ability to open regular external links that do not require map coordinates


dahlalex:
**Is your feature request related to a problem? Please describe.**
Currently, the "External url" control is limited to sending the map's center coordinates in the URL template and showing the center marker. This constraint can make it challenging to use the control for simpler purposes, such as opening regular external links that do not require map coordinates.

**Describe the solution you'd like**
I propose enhancing the "External url" control to add an option allowing it to open regular URLs without needing to send the map’s center coordinates or display the center marker. This would increase the flexibility of the control, enabling it to:

- Function as it currently does, sending the center coordinates and displaying the marker, or
- Open links directly as a normal URL, without displaying the center marker., when coordinates aren’t needed.

This approach retains the original functionality of the "External url" control while broadening its use cases.

_Here is a video that shows the potential enhancement:_

![external](https://github.com/user-attachments/assets/5f42ebaf-904e-42c0-a39f-2b2f1ea61e8f)

PR #2071, opened 2024-11-06
feature: external url control with regular links

dahlalex:
Aims to fix [#2064 ](https://github.com/origo-map/origo/issues/2064)
Linked with [#330](https://github.com/origo-map/origo-documentation/pull/330)

### DIFF
--- a/src/controls/externalurl/externalurlSeveralButtons.js
+++ b/src/controls/externalurl/externalurlSeveralButtons.js
@@ -15,6 +15,7 @@ const externalurlSeveralButtons = function externalurlSeveralButtons(options = {
   const buttons = [];
   const subButtons = [];
   const direction = options.direction;
+  const linkTarget = options.target || '_blank';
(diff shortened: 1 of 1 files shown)


johnnyblasta (approved):
Tested and LGTM

steff-o:
In general the controls should be added in the configured order but there are some controls that are "default" and are added before the "non-default".

The default controls are (from origo.js):
``` javascript
 defaultControls: [
   { name: 'localization' },
   { name: 'scaleline' },
   { name: 'zoom' },
   { name: 'rotate' },
   { name: 'attribution' },
   { name: 'fullscreen' }
 ]
```
Not all of them are in the left toolbar so they don't matter but I can't find any explanation to why home control comes before externalurl.

steff-o:
Alrighty! I missed that subtle part where controls either calls getMain().getNavigation() or getMain().getMapTools() or the other ones to get their parent element.

PR #2075, opened 2024-11-08
feature: Add infowindow component to origo.api()

### LINKED ISSUE #2074: Infowindow in Origo API
It is useful to be able to launch an infowindow ( not THE infowindow but the component ) from outside of Origo core.

jokd:
Fixes #2074


johnnyblasta:
You have a conflict to resolve @jokd

jokd:
Yeah and to spice up things even more we have the infowindow setting in the config where infowindow is one option. Based on your comment a small rework of the component is maybe a good idea.

Infowindow has as far as I know other features than the modal, but things might have changed since I last used it. 
- It has no backdrop and can be present while using the map. 
- You can have it floating or aligned to the left. 
- There are methods for altering the content.

Maybe it should have been put in the ui folder along with the modal but then perhaps the logger should as well.

steff-o:
> Yeah and to spice up things even more we have the infowindow setting in the config where infowindow is one option. Based on your comment a small rework of the component is maybe a good idea.
> 
Are you going to do any rework?
> 
> Maybe it should have been put in the ui folder along with the modal but then perhaps the logger should as well.

The logger is already out there, so i can't be moved, and it is not really an ui class as it is a singelton instance. 

I strongly suggest renaming this infoWindow to something like _floatingPanel_ or something that sounds like a ui class and moving it to the ui folder.

jokd:
> > Yeah and to spice up things even more we have the infowindow setting in the config where infowindow is one option. Based on your comment a small rework of the component is maybe a good idea.
> 
> Are you going to do any rework?
> 
> > Maybe it should have been put in the ui folder along with the modal but then perhaps the logger should as well.
> 
> The logger is already out there, so i can't be moved, and it is not really an ui class as it is a singelton instance.
> 
> I strongly suggest renaming this infoWindow to something like _floatingPanel_ or something that sounds like a ui class and moving it to the ui folder.

Renamed and moved it. Can be used like this:

```
const viewer = origo.api();
const contentComponent = Origo.ui.Element({
  innerHTML: 'Content',
  cls: 'margin-left padding-small text-small',
});
viewer.addComponent(
  Origo.ui.FloatingPanel({ 
    viewer, 
    contentComponent, 
    type: 'floating', 
    title: 'Profile',
    isActive: true
  })
);
```

steff-o (commented):
It seems like it's been there "forever", but calling close actually hides it. So if a Control does not reuse its floatingPanels, the DOM will be littered with invisible floatingpanels unless caller implicitly tracks down the div and removes it. Before exposing it, would it be idea to rename the close to hide and add a new close that removes it from the DOM just like the modal ?

jokd:
> It seems like it's been there "forever", but calling close actually hides it. So if a Control does not reuse its floatingPanels, the DOM will be littered with invisible floatingpanels unless caller implicitly tracks down the div and removes it. Before exposing it, would it be idea to rename the close to hide and add a new close that removes it from the DOM just like the modal ?

I was thinking of that as well. But the default (and expected way) way of "closing" is hiding, that's the way it's been used. But I renamed close to hide and added a remove method that removes the element from the DOM, certainly useful when using it with the API. The default behaviour for the close button is hide but can be overridden by setting removeOnClose option to true.

```
const panel = Origo.ui.FloatingPanel({ 
    viewer: origo.api(), 
    type: 'floating', 
    title: 'Profile',
    isActive: true,
    removeOnClose:true
});
panel.render();
panel.hide();
panel.show();
panel.remove();
```

steff-o (approved):
LGTM

PR #2078, opened 2024-11-08
feature: Make measure layer queryable

### LINKED ISSUE #2076: Make measure queryable
It should be possible to make the measure layer queryable and show the measurements in the popup/infowindow.

jokd:
Fixes #2076

### DIFF
--- a/src/controls/measure.js
+++ b/src/controls/measure.js
@@ -22,6 +22,7 @@ const Measure = function Measure({
   useHectare = true,
   snap = false,
   snapIsActive = true,
+  queryable = true,
   snapLayers,
   snapRadius = 15,
   highlightColor,
@@ -106,8 +107,8 @@ const Measure = function Measure({
     }
     if (
       tip
-    && geomType === 'Point'
-    && !modify.getOverlay().getSource().getFeatures().length
+      && geomType === 'Point'
+      && !modify.getOverlay().getSource().getFeatures().length
     ) {
       tipPoint = geometry;
       tipStyle.getText().setText(tip);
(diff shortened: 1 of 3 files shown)


Grammostola:
Hi, nice addition!
I've two observations:
- It doesn't appear to work on buffers
- The area and length measurements in the infowindow (numbers) could be localized, as well as the text

jokd:
> Hi, nice addition! I've two observations:
> 
> * It doesn't appear to work on buffers
    That is because buffers are points with a radius style, hence got no area. We did something similar in the style window but I remember the figures being a bit off and needs some rework.

> * The area and length measurements in the infowindow (numbers) could be localized (you've already localized the text)
     That formatting is not in the measure control but in the attribute handling. (It would also be nice if zero rows were omitted, eg area for lines.)

jokd:
> > That formatting is not in the measure control but in the attribute handling. (It would also be nice if zero rows were omitted, eg area for lines.)
> 
> I see, the numeral and unit is returned via the `@length` and `@area` functions..wherever they are defined (I had a quick look and it seemed entirely cryptic) ..for yes those functions would need to be modified to return a localized numeral. Point me in a direction and I can have a look.
> 
> Edit. Right, utils/formatareastring.js and formatlengthstring.js need to be localized ...via getattributes.js, selecteditem.js, featureinfo.js and the viewer I think

src/geometry/getarea and /getlength seems to be used by the replacer

Grammostola:
Indeed. And it is possibly the replacer that needs to become aware of "null" results from getArea, getLength. 
 Either way for clarity I am working on localizing the relevant numbers (probably via 
```js
    attributes: [
      { html: `${locLength}: {{@length(1)}}`,
        localization },
      { html: `${locArea}: {{@area(1)}}`,
        localization }
    ],
```
even if it looks a little funny to send the loc object for every instance of `html` use. It seems the less potentially disruptive way)

as well as making something aware in order to not show 0.0 results.

Grammostola:
Info on a polygon should not include "Length: 0.0" now and vice versa for area.

In order to achieve this some things were changed:
- the `getArea()` and `getLength()` functions now return 0 instead of a formatted text string when the length measured is 0 (there's no need to format the response if it is 0, the rationale is)
- the `html` method of getAttributes now returns null instead of a new element with null inside (the latter seemed pointless)
- the replacer method will return null if it received 0 from `getArea()` or `getLength()`

jokd:
@Grammostola Looks good, thanks

Grammostola:
@jokd Cool,  ok to merge do you reckon? I think so : )

jokd:
> @jokd Cool, ok to merge do you reckon? I think so : )

Sure, I'll fix the dox

PR #2082, opened 2024-11-12
feature: Store getCapabilites document for source

### LINKED ISSUE #1588: Global getCapabilities-document
More and more functionality requires the getCapabilites document being fetched from the web server and therefore are requested multiple times.

I suggest the getcap document is requested on map init for a source (if setup to do so) and made available through the API.

jokd:
Fixes #1588

When capabilitiesURL for a source is set, the getCap doc will be saved unless opted out ( `"saveCapabilitiesDoc": false` ). 

Configured like:

```
 	"source": {
		"local": {
			"url": "http://localhost:8080/geoserver/ows",
			"capabilitiesURL": "http://localhost:8080/geoserver/ows?service=wms&version=1.1.1&request=GetCapabilities",
			"saveCapabilitiesDoc": false
		}
	},
```

and accessible with `origo.api().getMapSource()['local'].capabilitiesDoc`

### DIFF
--- a/src/getCapabilities.js
+++ b/src/getCapabilities.js
@@ -21,7 +21,8 @@ const getCapabilities = function getCapabilities(name, getCapabilitiesURL) {
       if (xmlHttp.readyState === 4 && xmlHttp.status === 200) {
         resolve({
           name,
-          capabilites: responseParser(xmlHttp.responseText)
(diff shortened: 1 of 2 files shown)


johnnyblasta (approved):
Tested and got it to work after finding a way to not have CORS interfer the function.

PR #2086, opened 2024-11-20
fix: Make position control work without additional projection codes

### LINKED ISSUE #2085: Position control crashes when no additional projection codes are present in the config
**Describe the bug**
Position control crashes when no additional projection codes are present in the config because Origo tries to convert an undefined object to an array.

Example config:

```
{
  "name": "position",
  "options": {
    "title": "Sweref99 16.30"
  }
}
```

**To Reproduce**
Use the config above, or try to open the editor example (examples/editor.html)

tonnyandersson:
Fixes #2085.


steff-o (approved):
LGTM

Felt a bit guilty about that one.

PR #2087, opened 2024-11-22
feature: add trace to draw interaction

### LINKED ISSUE #2072: Trace tool in edit
When drawing new features it is hard to not create gaps or overlaps against other features in the same layer or other layers. Today it exists a snapping feature but if there are many vertices it is tedious work adding all vertices.

Possibility to trace an existing feature by just following it.
trace

Not sure how to activate it though. The obvious solution would be to activate trace using a modifier key (shift, ctrl, alt or whatever combination isn't already used), but that would not work well on mobile devices. Maybe not a problem as not many will do serious precision editing on a touch screen. Alternatively a new tool would be added to the create feature button. Feel free to discuss!

None, it's a non-negotiable end user request.

**Additional context**
The functionality exists in Openlayers as an option to the draw interaction. It does not exist on modify interaction, so it would only be implemented for new features. The same approach could be applied to the draw tools, but that is beyond the scope of this issue.

Demo (not mine): https://openlayers.org/en/latest/examples/tracing.html

steff-o:
Fixes #2072

Adds possibility to trace other features when drawing new features. As a bonus the snap tolerance can configured.

Tracing starts by snapping to a feature and clicking, thus setting a vertex. All possible paths that are traceable are lit up and the mouse can only be moved along the highlighted paths until clicked, which sets another vertex and leaves tracing but the draw interaction is still active so it is possible to jump to another features or just draw wherever.

Snapping must be active for this functionality to be meaningful as it is almost impossible to hit a line exactly. All snap layers are used for tracing.

Disclaimer: most of the cool stuff are implemented by OL, and as such there are some limitations in how much the user experience can be changed.

This PR only contains a static configuration if trace is active or not, but the foundation is laid to add an on/off switch in a menu or dynamically activate/deactivate by requiring a modifier (alt or ctrl) or a keyboard shortcut.

New configuration i edit.options:
``` jsonc
"traceStyle": "tracehighlight", // Optional named style for displaying possible trace paths. Defaults to some hideous green
"snapTolerance": 5, // Optional snap tolerance in pixels. Defaults to 10 (same as before)
"trace": true // Optional, defaults to false. When true, tracing is active when editing. Requires snap to be active
```
Documentation will be updated accordingly.

_During discussions on the issue a question was raised if the trace functionality creates topologically correct features with respect to overlaps and slivers against other features. As it can not stop a user from deliberately create overlaps or gaps outside the snap tolerance, it still can produce overlaps and slivers when seemingly following existing features. As long as the start and end of a trace snaps to a vertex, all vertexes should have the exact same coordinates, but if a trace starts or ends between vertexes, there will be a small sliver or overlap as the original feature will not have a vertex in the corresponding place and the new features vertex will unlikely be exactly on the line segment mathematically. To avoid that, a new vertex must be inserted on the traced feature, but that is far beyond the scope of this PR._

### DIFF
--- a/src/controls/editor/edithandler.js
+++ b/src/controls/editor/edithandler.js
@@ -4,8 +4,11 @@ import Modify from 'ol/interaction/Modify';
 import Snap from 'ol/interaction/Snap';
 import Collection from 'ol/Collection';
 import Feature from 'ol/Feature';
-import { LineString, MultiPolygon, MultiLineString, MultiPoint } from 'ol/geom';
+import { LineString, MultiPolygon, MultiLineString, MultiPoint, Polygon } from 'ol/geom';
 import { noModifierKeys } from 'ol/events/condition';
+import VectorSource from 'ol/source/Vector';
+import VectorLayer from 'ol/layer/Vector';
+import { squaredDistance, toFixed } from 'ol/math';
 import { Button, Element as El, Modal } from '../../ui';
 import Infowindow from '../../components/infowindow';
 import store from './editsstore';
@@ -23,6 +26,7 @@ import topology from '../../utils/topology';
 import attachmentsform from './attachmentsform';
 import relatedTablesForm from './relatedtablesform';
 import relatedtables from '../../utils/relatedtables';
+import Style from '../../style';
 
 const editsStore = store();
 let editLayers = {};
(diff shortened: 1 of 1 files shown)


johnnyblasta (approved):
LGTM

PR #2091, opened 2024-12-03
feature: localize of zoom, geoposition, about and home control

### LINKED ISSUE #2090: Localize zoom, geoposition, about and home control
Some tooltips which isn't localized.

johnnyblasta:
Closes #2090

### DIFF
--- a/src/controls/about.js
+++ b/src/controls/about.js
@@ -5,11 +5,16 @@ const About = function About(options = {}) {
     buttonText,
     target
   } = options;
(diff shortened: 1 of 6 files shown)


steff-o (approved):
LGTM!

Would be cool to localize _About control_  `content` property too, but that is probably better implemented as a configuration feature of the _About control_ instead of having to provide a custom locale definition, which makes it out of scope for this PR.

PR #2096, opened 2024-12-06
feature: Remove group from legend

### LINKED ISSUE #2093: Be able to remove subgroup from map
We are developing a function where the user can add layers from our geoserver catalogue and in case the layer is a Geoserver Layer Group it is split into the individual layers and put inside a subgroup of Mina lager. This subgroup should be able to be removed similar to how it works for individual layers.

Context menu where the user can remove a subgroup

image

jokd:
Fixes #2093 
Adds a removable option for subgroups


johnnyblasta:
Getting this when testing on a subgroup:
image

jokd:
Had to rewrite the onRemoveGroup method since it didn't actually remove groups, only subgroups.
Configure like this:
```
    {
      "name": "agroup",
      "title": "A group",
      "removable": true,
      "expanded": true,
      "groups": [
        {
          "name": "subgroup1",
          "title": "subgroup1",
          "removable": true
        }, 
        {
          "name": "subgroup2",
          "title": "subgroup2",
        }
      ]
    }
```

jokd:
@johnnyblasta Wanna have another look?

johnnyblasta (approved):
LGTM
Some documentation on it though, so we can merge.

jokd:
> LGTM Some documentation on it though, so we can merge.

Docs are here: https://github.com/origo-map/origo-documentation/pull/345

PR #2097, opened 2024-12-06
fix: Adding layers to map without overlays

### LINKED ISSUE #2095: Error when adding layers to map without overlays
**Describe the bug**
When adding a layer to a map without layers the legend gets corrupt

**To Reproduce**
1. Create a amap with groups but no layers except background layers
2. Add a layer with the API or Layermanager plugin
3. Cry

**Screenshots**
image

jokd:
Fixes #2095

### DIFF
--- a/src/controls/legend.js
+++ b/src/controls/legend.js
@@ -223,6 +223,7 @@ const Legend = function Legend(options = {}) {
   const setVisibleLayersViewActive = function setVisibleLayersViewActive(active) {
     if (!visibleLayersControl) return;
     visibleLayersViewActive = active;
+    overlaysCmp.readOverlays()
     if (visibleLayersViewActive) {
       document.getElementById(overlaysCmp.getId()).classList.add('hidden');
       document.getElementById(visibleOverlaysCmp.getId()).classList.remove('hidden');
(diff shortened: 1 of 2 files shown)


Grammostola:
I'm having difficulties reproducing the issue.
If I
-Create a map with groups but no layers except background layers
-Add a layer with the API or Layermanager plugin

Then what appears are the groups. (A map with groups beside "background" (and root) shows no groups if no layers are present. The groups appear when layers are added and disappear when layers go missing again. This appears the same with your proposed change. )

What am I missing?

jokd:
> I'm having difficulties reproducing the issue. If I -Create a map with groups but no layers except background layers -Add a layer with the API or Layermanager plugin
> 
> Then what appears are the groups. (A map with groups beside "background" (and root) shows no groups if no layers are present. The groups appear when layers are added and disappear when layers go missing again. This appears the same with your proposed change. )
> 
> What am I missing?

Looked at my config and the legend option `"visibleLayersControl": true` is apparently needed to mess it up.

Grammostola:
Right, that and reproducing it in the correct version (current Origo master) made the reported issue reproducible (the groups show initially then disappear when a layer is added and the added layer isn't seen either in the legend. Hitting the visible layers control button twice seemingly remedies the situation. )

I'm not getting this PR to run so well. Unless `"visibleLayersViewActive": true` (it defaults to false) it crashes on start. If `"visibleLayersViewActive": true` then the groups aren't visible on start, they do appear after a layer is added but then the legend looks a bit glitchy.

jokd:
Replaced with #2115

PR #2098, opened 2024-12-06
fix: viewer.removeGroup() also removes subgroups

### LINKED ISSUE #2094: viewer.removeGroup() won't remove subgroups
**Describe the bug**
The viewer.removeGroup() method needs to be rewritten to also remove its subgroups. As of now it won't work the way it is supposed to.

jokd:
Fixes #2094 
viewer.removeGroup() also removes subgroups


johnnyblasta (approved):
LGTM

Issue #2106, opened 2024-12-11
Replace node-sass


sweco-sedalh:
**Describe the bug**
Origo currently uses node-sass to compile SASS to CSS. node-sass is [no longer actively maintained](https://sass-lang.com/blog/libsass-is-deprecated/), and [breaks on newer versions of Node](https://github.com/sass/node-sass/issues/3415).

The recommended replacement is the dart-sass project, through the `sass-embedded` package. A part from some CLI-switches the replacement is fully compatible with the old node-sass, so no adjustments of the SASS code in Origo has to be done.

**To Reproduce**
1. Install Node.js version 22
2. Clone Origo (or remove `node_modules`)
3. Run `npm install`

**Expected behavior**
Origo dependencies are built.

Instead, a long error message is produced (originating from node-gyp, a dependency of node-sass).

**Screenshots**
\-

**Desktop (please complete the following information):**
\-

**Smartphone (please complete the following information):**
\-

**Additional context**
\-

PR #2107, opened 2024-12-11
Replace node-sass by dart-sass

sweco-sedalh:
See #2106 for more background and links.

`node-sass` is no longer actively developed, and breaks on newer versions of Node.js. `sass-embedded` is based on the `dart-sass` project and is almost a drop-in replacement.

Closes: #2106


sweco-sedalh:
Yes, decided to not get into those as part of this PR to keep it focused.

PR #2112, opened 2024-12-13
fix: Test other infoFormat when getFeatureInfo fails

### LINKED ISSUE #2108: Try other info_formats when getfeatureinfo request fails.
For instance, arcgis server have support for `application/geojson` in the getcap document but the correct request info_format is `application/geo+json`. This makes layers automatically added by parsing the getcap document being set up with an invalid infoFormat making getfFeatureInfo-request fail. 

When failing a getfFeatureInfo-request, try a couple of different formats to find a format the server support and change infoFormat for the layer.

jokd:
Fixes #2108 Try different infoFormats if getFeatureInfo request fails with the configured infoFormat (application/geojson or application/geo+json)

### DIFF
--- a/src/getfeatureinfo.js
+++ b/src/getfeatureinfo.js
@@ -143,8 +143,8 @@ async function getFeatureInfoUrl({
   }
 
   if (layer.get('infoFormat') === 'application/geo+json' || layer.get('infoFormat') === 'application/geojson') {
-    const format = layer.get('infoFormat');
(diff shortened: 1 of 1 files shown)


jokd:
Will have to rewrite this since the linter didn't like my loop

johnnyblasta (approved):
LGTM
Nice that it sets the correct format on the layer, so subsequent click don't have to round trip with wrong format.

PR #2114, opened 2024-12-16
Feature: Make subgroups draggable

### LINKED ISSUE #2113: Make subgroups draggable
When a group is set to draggable it is possible to rearrange layers within the group. It is however not possible to move a subgroup which in many cases behaves like a layer and therefor should be handled the same way. The drawing order (z-index) should also be set accordingly.

Be able to move subgroups when a group is set to draggable. 

**Additional context**
Touches #1967 but doesnt aim that high.

jokd:
Fixes #2113 
Subgroups in a group with `draggable: true` can be moved just like other layers.


johnnyblasta (approved):
LGTM

PR #2116, opened 2024-12-20
Fix print composite layer legend

### LINKED ISSUE #2092: Print legend does not work with composite WMS layers
**Describe the bug**
The legend in the print view does not correctly handle WMS composite layers. If a WMS layer is a composite layer (server side, not Origo group layer) only the rules from the first layer is displayed but the images for those rules are missing.

**To Reproduce**
1. Set up a composite layer consisting of at least two layers with thematic styling on Geoserver
2. Configure origo to display the layer having `hasThemeLegend` (but not thematicStyling)
3. Go to print preview
4. Only the legend rules from the first layer is displayed but the legend images are the "image not found place holders". 

**Expected behavior**
I would expect all rules from all layers to be displayed.

**Screenshots**
image
_Image 1: The composite legend in map view with rules from both layers in composite layer_

image
_Image 2: The legend i print only displaying rules from first layer and browser's placeholder images_

**Mapserver**
 - Geoserver 2.25.3

**Additional context**
The steps to reproduce above uses CSS styles, i.e. rules without names, which would explain why there are no images. If I change to SLD using named rules it instead squeezes in the entire group image  on each rule. That seems to be some quirky bug in Geoserver that produces an image of the composite layer when the rule name matches a rule name in an underlying layer's style.

From what I have seen in the code, the problem is that print-legend only loops the rules in `json.Legend[0]` and does not use the correct layer name. It sends the composite layer name and the rule name for the underlying layer. It would probably have to loop the `json.Legend[]` array and make calls for all underlaying layers and rules, or just get the composite layer legend graphic like the map legend does. If it had been configured with `thematicStyling` getting the entire composite layer graphic may not be ideal if it supposed to only display visible rules.

steff-o:
Fixes #2092 by generating valid legend graphics for Geoserver _Layer Groups._

Should generate all rules for all layers in a _Layer Group_ in print legend.

There are some limitations:
1. If a style has several rules, the rules must be named (same as with ordinary WMS layers)
2. A Group Layer must use the default style for each contained layer, i.e. not define another style when used inside the Group Layer, otherwise the map legend and print legend will display different symbols.
3. A top level (without workspace) Group Layer can only contain layers with unique names through all workspaces.

These limitations are all a result of missing information in the response from the  getLegendGraphic as JSON request. To overcome that a bunch of Geoserver api request would have to be performed, but that seemed a bit overmuch.

Also `thematicStyling` is not supported as it not supported by the map legend and would require an issue of its own.

No new configuration is added.

### DIFF
--- a/src/controls/print/print-legend.js
+++ b/src/controls/print/print-legend.js
@@ -50,6 +50,15 @@ const LayerRow = function LayerRow(options) {
   };
 
   /**
+   * Helper that creates a WMS getLegendGraphics request url string
+   * @param {any} url base url
+   * @param {any} layerName name of layer to create legend for
+   * @param {any} format valid mime type
+   * @returns {string} A WMS getLegendGraphics request url string
+   */
+  const createGetlegendGrapicUrl = (url, layerName, format) => `${url}?SERVICE=WMS&layer=${layerName}&format=${format}&version=1.1.1&request=getLegendGraphic&scale=401&legend_options=dpi:300`;
+
+  /**
    * Returns the URL to the WMS legend in the specified format
    *
    * @param {Layer} aLayer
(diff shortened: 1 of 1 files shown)


Grammostola (approved):
Seems to do what it says on the tin. `thematicStyling` layers with a partial theme list active still print ok. Perhaps if different treatment is needed for similar types of composite layers from other map server vendors a check can be made for the source `type` (like print-resize does already).

Anyway based on my testing and no linting issues I approve this. Nice addition.

PR #2118, opened 2024-12-20
fix: resolve deprecation warnings for sass

### LINKED ISSUE #2117: Resolve deprecation warnings for stylesheets
**Describe the bug**
With the move from node-sass to dart-sass there was alot of deprecation warnings that needs to be adressed.

**To Reproduce**
1. Start Origo with npm start
2. See deprecation warnings in console

**Expected behavior**
All warnings should be adresssed to future proof the code.

johnnyblasta:
Fixes #2117


steff-o:
There are still deprecation warnings for glide. As it it beyond our control I guess we'll have to accept it. But is it really necessary to rebuild the stylesheets for glide? The npm package contains pre-built stylesheets, why are we not using them?

There is an issue logged for this in the glide repo: glidejs/glide#704

steff-o:
The glide stuff aside, this PR does not produce the same CSS as before. Basically all definitions have lost a top level selector, either `.o-ui` for ui-elements or `.o-map` for everything else. Here are two examples. but it looks the same for all definitions. Not sure if it affects anything as the selection is only widened. Maybe if map is embedded in a document that has more content than just the app wrapper. Also possibly the print control seems to search for exact style definitions `const h4Rule = getCssRule('.o-ui h4, .o-ui .h4');` to resize them.

Before:
``` css
.o-ui .rounded {
  border-radius: 0.25rem;
  overflow: hidden;
}
```
After:
``` css
.rounded {
  border-radius: 0.25rem;
  overflow: hidden;
}
```
Before:
``` css
.o-map svg,
.o-map object {
  width: 100%;
}
```

After:
``` css
svg,
object {
  width: 100%;
}
```

johnnyblasta:
Seems to be a effect of switching from @import to @use that the scope is lost. I'll try to get it back some how.

johnnyblasta:
Do I have to edit every scss file and add the scope to all that had it before or are I'm missing something?

johnnyblasta:
Now it produces correct code, had to use correct syntax for the nested imports.
The glide warnings remains though.

steff-o (approved):
LGTM

It produces a byte exact result (except for a comment). Probably could have fixed the glide stuff as well, but I don't know if there is a special reason for including the scss instead of the pre built css, so we'll leave it for now. Better merge before the conflicts pile up.

johnnyblasta:
Had to move import order to avoid a warning to last, so comment moved.

PR #2123, opened 2025-01-08
feature: localize legend control

### LINKED ISSUE #2122: Localize Legend control
Some strings has to be localized.

johnnyblasta:
Closes #2122


Grammostola:
Nice. Some thoughts:
- "Show only lit layers" => "Show only visible layers", I'm not certain that we are totally consistent between "active" and "visible" but we toggle layer visibility..
- "Search layer" => "Find layer" or "Search for layer", else it sounds a bit like a noun
- "turnOffTitle" could be "turnOffLayersTitle" to make it consistent with "turnOnLayersTitle"
- The translation for "turnOnLayersTitle" could be "Turn on all layers" to make it consistent with "turnOffTitle"'s translation. It feels intuitive that these operations do not apply to the background layers (as they are radios)
-  "removeLayerQuestion" and "removeLayer"'s translations ought to employ the same verb ("remove"/"ta bort" I think)
- I get a crash when a map layer has `visible: true` . It seems as if overlay.js's localization ref isn't alive

johnnyblasta:
@Grammostola is it better now?

Grammostola:
Yes. The only remaining thing that I can see is "Vill du radera lagret?" => "Vill du ta bort lagret".

Grammostola (approved):
lgtm

PR #2124, opened 2025-01-10
Edit split line tool

### LINKED ISSUE #2073: Cut line in edit
When a line is too long, the only way of shorten it is to delete each vertex from the end. Also, if a line feature needs to be split in two, the only solution is to draw a new line for one of the parts and trim vertices from the other or draw two new lines.

A cut line tool. When active, clicking on a line will split the line in two which will result in two features, one with the first part of the line and the other with the last part and both will have the same attributes.

Not sure how to activate it though. The obvious solution would be to use a modifier key (shift, ctrl, alt or whatever combination isn't already used) when clicking a line that is selected for modification, but that would not work well on mobile devices. Maybe not a problem as not many will do serious precision editing on a touch screen. Alternatively a new tool would be added to the edit menu when a feature is selected for modification.

Let the end users suffer.

**Additional context**
The same could be applied to the draw tools, but has not been requested by the users, so it's beyond the scope of this issue.

The approach on how to activate this functionality would probably be similar to that of #2072

steff-o:
Fixes #2073

Adds a new button to the editor toolbar with modify tools that is activated when the correct geometry for that tool is selected. Currently the only available tool is split-line-by-point, which does just that: select a line feature, select the split tool and click where on the selected line it should be split. It cuts the original feature and creates a new with the rest of the line. The copy gets all attributes from the original feature (except id).

To activate the modify tools menu the editor control must be configured with:
`"modifyTools": true`
The configuration is only on editor level for now. In the future one could imaging having it on layer level as well or specify which tools are available.

As a bonus Edithandler and EditToolbar are rewritten as Components so they can emit events as I didn't wanted to follow the pattern the way the editor toolbar uses DOM events.

### DIFF
--- a/src/controls/editor.js
+++ b/src/controls/editor.js
@@ -172,6 +172,11 @@ const Editor = function Editor(options = {}) {
     createFeature,
     editFeatureAttributes,
     deleteFeature,
+    // TODO: Ta bort, skall ej exponeras, bara anropas från menyn.
+    clip: () => {
+      editHandler.setModifyTool('split-line-by-point');
+    },
+
     changeActiveLayer: (layerName) => {
       // Only need to actually change layer if editor is active. Otherwise state is just set in toolbar and will
       // activate set layer when toggled visible
--- a/src/controls/editor.js
+++ b/src/controls/editor.js
@@ -1,5 +1,5 @@
 import { Component, Button, dom } from '../ui';
-import editorToolbar from './editor/editortoolbar';
+import EditorToolbar from './editor/editortoolbar';
 import EditHandler from './editor/edithandler';
 
 const Editor = function Editor(options = {}) {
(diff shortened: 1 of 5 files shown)


johnnyblasta:
When importing editorToolbar you've used PascalCase instead of CamelCase as we usually do.

steff-o:
> When importing editorToolbar you've used PascalCase instead of CamelCase as we usually do.

I don't agree. According to airbnb, Pascal casing should be used for classes and constructors. As I interpret it, in javascript terminology a constructor is not only the special function used in classes, but also a function that returns a javascript object. This is a function that returns an object. Also, throughout the code Pascal casing are already usually used for Components and Controls.

johnnyblasta:
> > When importing editorToolbar you've used PascalCase instead of CamelCase as we usually do.
> 
> I don't agree. According to airbnb, Pascal casing should be used for classes and constructors. As I interpret it, in javascript terminology a constructor is not only the special function used in classes, but also a function that returns a javascript object. This is a function that returns an object. Also, throughout the code Pascal casing are already usually used for Components and Controls.

OK, seems to be reasonable to use PascalCase for functions that you treat as constructors.

johnnyblasta:
Fix a PR for documenting `"modifyTools": true` and then it can be merged.

PR #2128, opened 2025-01-16
fix: make layer removal happen after setting popupmenu visibility

### LINKED ISSUE #2127: Silent crash when removing a layer
**Describe the bug**
Non fatal crash when removing a layer

**To Reproduce**
- Have a removable layer in your Origo session. (`removable: true` prop for the layer or an added layer which normally comes with it).  (`promptlessRemoval: true` for the layer is irrelevant.)
- Toggle its context menu and choose to remove it. Confirm this action. Check the browser console.

**Expected behavior**
Layer removed from map and no crash in the console

**Additional context**
controls/legend/moreinfobutton tries to do
`popupMenu.setVisibility(false)` which fails since by the time it executes the popupMenu element's id does not reside in the document. It seems as if it disappears when the layer is removed from the map, which just happens before.

No other action of moreinfobutton (export, zoom to, more info) seems to experience the same issue.

Grammostola:
Aims to fix #2127 and preserve current behavior (minus the crash).

### DIFF
--- a/src/controls/legend/moreinfobutton.js
+++ b/src/controls/legend/moreinfobutton.js
@@ -125,11 +125,11 @@ export default function createMoreInfoButton(params) {
         labelEl.addEventListener('click', (e) => {
           const doRemove = (layer.get('promptlessRemoval') === true) || window.confirm(localize('removeLayerQuestion'));
           if (doRemove) {
+            popupMenu.setVisibility(false);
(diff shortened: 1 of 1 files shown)


steff-o (approved):
LGTM.

(But hiding the popup before deleting the layer is not really necessary as it will be removed from he DOM )

PR #2149, opened 2025-02-20
fix: give the iw title area a minimum height

### LINKED ISSUE #2148: infowindow with infowindowOptions title = " " doesn't work so well with the minimize function
Have a relatively fresh Origo clone, set the config for `infowindow` like so:

```json
  "featureinfoOptions": {
    "infowindow": "infowindow",
    "infowindowOptions": {
      "title": " "
  }
},
```

In order to not visibly have a header for the infowindow pane (we've never seen the use of one). 

Then infoclick an Origo municipality point and minimize the conjured infowindow pane.

See 

Image

Grammostola:
Aims to fix #2148  (22.5px seems to be the element's height when text is supplied (incl when the option isn't supplied)

### DIFF
--- a/scss/_infowindow.scss
+++ b/scss/_infowindow.scss
@@ -66,6 +66,7 @@
 .urval-textnode-container {
   font-size: 15px;
   margin-bottom: 0.5em;
+  min-height: 22.5px;
 }
 
 .expandable_list {


johnnyblasta (approved):
LGTM

PR #2150, opened 2025-02-20
feature: new attriubte types audio and video

### LINKED ISSUE #2137: Attribute audio and video
Make it possible to have audio and video as attributes and play theese in the infowindow

A audio attribute with player:
Image

A video attribute with player:
Image

If there are multiple audio/video in the attribute should it be added in a playlist or should it have serveral players with only one track?

johnnyblasta:
Closes #2137 

Config a layer with audio and video attributes. Add urls to audio and video files to test.
```
{
      "name": "origo-cities",
      "title": "Origokommuner",
      "group": "root",
      "style": "origo-logo",
      "type": "GEOJSON",
      "attributes": [
        {
          "name": "name"
        },
        {
          "audio": "ljud",
          "title": "Ljud: ",
          "splitter": ";"
        },
        {
          "video": "film",
          "title": "Film: ",
          "splitter": ";"
        }
      ]
```

image
image

Using base64 data is also supported, but not recommended since the file size get large very quick.


steff-o (changes requested):
I have tested using the hardest use case (but also the most likely)  I can think of: attachments, but I can't get it to work without adding:
```
} else if (attribute.audio) {
            val = getContent.audio(feature, attribute, attributes, map);
```
in the `getattributeHelper` function (and similar for video)

In your example, you set a `title`, but that is not showed anywhere. When using arbitrary numbers of attachments it wouldn't work anyway if the title is for each individual file. In that case the concept of `linktext` attribute type from url could be used if desired.

Not sure how to test the changes in validate.js, as I can't find any place a validation is triggered. If it was possible to upload a file in the editor, that would probably be the place, but audio and video attributes are not handled in the editor at all.

steff-o (commented):
This time I focused on the editor functionality.

Video validate is still never used
Audio validate misses .oga extension which is allowed in the file picker.

However, validation is pretty limited as file selector already filters out all allowed file endings, unless the user manually enters (or copies) a file name in the file picker.

Remove video and audio does not work. It only disappears from from the editor dialog. Once opened again it is still there and not saved to database.

This implementation suffers from the same problem that some other attribute types has, that it is not possible to write a decent attributes configuration for both featureInfo and editor in the same application. This because the editor requires the "name" property to be set and that gets precedence over audio and video in infowindow, which renders the file as its base64 representation as text in the fetaureInfo window. Don't know if there is a way around it or if we have to accept that the usage has to be divided into one edit map and one read only (public) map, where the edit map will have to suffer.

All in all, the idea of adding large media files to columns in the editor probably has a very limited use case, as it will also slow down the layer considerably. One possible usage would be using related tables, which would be lazy loaded on featureInfo.

If there is no data in an audio/video column the title is displayed in the infoWindow, but the player is absent and no explanation why. The attribute should probably be completely missing like most other attributes. Possibly an explaining text that there is no content to play, but that would be pretty ugly if there are several placeholder columns (video1, video2 etc) in the table.

johnnyblasta:
Fixed issues with validation and removing of audio and video.
Also added the possibility to add the audio or video as an URL instead of file.
Which is a much better option than having them directly in the database, the drawback is you'll need a webbserver to add the files  and get the full URL pasted in the editor.

steff-o (approved):
LGTM!

If `splitter` is configured and feature has no audio/video it crashes on featureInfo, but splitter is only meaningful when attachments are used as the gui only supports adding one file per attribute. It is of course possible to enter a list of Url:s or add a list directly to the database (which I did), but that could easily be classified as not supported.

PR #2159, opened 2025-03-11
fix: Remove padding for attribution button

### LINKED ISSUE #2155: Huge attribution icon
**Describe the bug**

**To Reproduce**

**Expected behavior**
Small and discrete

**Screenshots**

Image

jokd:
Fixes large padding for attribution button
Fixes #2155

### DIFF
--- a/scss/externs/_ol.scss
+++ b/scss/externs/_ol.scss
@@ -157,7 +157,6 @@ button.ol-zoom-out {
 .ol-attribution {
   bottom: .5em;
   max-width: calc(100% - 3.5em);
-  padding: .35em 0;
   position: absolute;
   right: .5em;
   text-align: right;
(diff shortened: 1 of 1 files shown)


tonnyandersson:
I think this might be part of a slightly bigger issue - none of the media query based CSS rules seem to apply anymore.

jokd:
> I think this might be part of a slightly bigger issue - none of the media query based CSS rules seem to apply anymore.

You are probably right. Possibly the new sass library?

tonnyandersson:
Probably. It doesn't seem to agree with "&" as... \*looking at notes\* ...top-level parent selector. And without the "&", nothing happens. Haven't had the time to figure out the correct syntax yet, though.

jokd:
Replaced with #2172

PR #2163, opened 2025-03-11
fix: Fix for localization of area/length 

### LINKED ISSUE #2162: Html attribute function area not working
**Describe the bug**
Using the html attribute function `@area` causes an error instead of returning the expected area value.

**To Reproduce**
For example use the origo-mask layer in index. json
Set `"queryable": true` and add the html attribute area function like below.
`"attributes":[
      {
        "html": "<b>Area:</b> {{@area}}"
      }`

Click on the polygon (on the mask surrounding municipalities).
See error message: 
`Cannot read properties of undefined (reading 'getCurrentLocaleId')`

**Expected behavior**
Should show area for the polygon in the informationwindow/overlay.

jokd:
Fixes #2162

### DIFF
--- a/src/utils/formatareastring.js
+++ b/src/utils/formatareastring.js
@@ -10,7 +10,7 @@ function formatAreaString(area, opts = {}) {
     localization
   } = opts;
 
-  const localeNumberFormat = new Intl.NumberFormat(localization.getCurrentLocaleId());
+  const localeNumberFormat = new Intl.NumberFormat(localization?.getCurrentLocaleId() || undefined);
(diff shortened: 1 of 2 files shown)


Owemyr:
Thanks @jokd for the quickfix, works fine.

Grammostola:
Looks good, the util methods will revert to the browser locale which is a lot better than crashing. 
Ultimately the loc object should exist where it is employed so I'll create another issue for that.

Grammostola (approved):

PR #2172, opened 2025-03-20
fix: Rewrite sass includes

### LINKED ISSUE #2155: Huge attribution icon

jokd:
This is a PR that replaces #2159 and fixes #2155 
I ain't no sass ninja but I think we have to use use/include syntax along with mixins instead of `include meta.load-css('file')` to use the scss files without a major rewrite. In this PR I have adjusted ol and viewer, but this needs to be done for all files if it is our solution of choice, hence the draft status.

Some docs: https://sass-lang.com/documentation/breaking-changes/import/#nested-imports

### DIFF
--- a/scss/_ol.scss
+++ b/scss/_ol.scss
@@ -1,29 +1,34 @@
 @use 'externs/ol';
 
-.ol-attribution li a {
-  color: #bbb;
+@mixin default {
+  @include ol.default;
+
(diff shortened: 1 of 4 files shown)


jokd:
Closed since replaced by #2183

PR #2174, opened 2025-03-25
feature: localize area, length functions when used for layer attributes

### LINKED ISSUE #2164: Localize area and length functions for layer attributes
If a layer is configured with 
```js
      "queryable": true,
      "attributes":[ { "html": "<b>Area:</b> {{@area}}" }],
```

(or length) the resulting number will be formatted according to the current browser locale, which is not how the Localization control currently operates (it has a `localeId` and `fallbackLocaleId` with the presumption that there might be missing strings in the chosen `localeId` but there won't be a missing localization obj)

The issue manifests as 

Image

i.e there might be a mismatch between the chosen Origo language and the numbers returned for area or length when querying a layer configured as above.

The target state is to make the numbers shown by querying a layer configured like above match the current Origo language/locale.

Grammostola:
Seeks to fix #2164 .

Random thoughts:
- localizing units of measurement as @MattiasSp mentioned today would include number conversion and unit appending and it would go outside of the scope of intl.NumberFormat afaIct It could be a nice improvement down the line, especially if users in London are accustomed to seeing miles and feet when employing the ruler tool in their local web map.
- I don't pretend to understand why getAttributes and getAttributesAsync are both used when infoclicking a vector layer :)

### DIFF
--- a/src/getfeatureinfo.js
+++ b/src/getfeatureinfo.js
@@ -22,6 +22,7 @@ const isValidJSON = str => {
  * @param {any} layer
  * @param {any} map
  * @param {any} groupLayers
+ * @param {any} [localization]
  * @returns {SelectedItem}
  */
 function createSelectedItem(feature, layer, map, groupLayers, localization) {
--- a/src/getfeatureinfo.js
+++ b/src/getfeatureinfo.js
@@ -24,7 +24,7 @@ const isValidJSON = str => {
  * @param {any} groupLayers
  * @returns {SelectedItem}
  */
-function createSelectedItem(feature, layer, map, groupLayers) {
+function createSelectedItem(feature, layer, map, groupLayers, localization) {
   // Above functions have no way of knowing whether the layer is part of a LayerGroup or not, therefore we need to check every layer against the groupLayers.
   const layerName = layer.get('name');
   let groupLayer;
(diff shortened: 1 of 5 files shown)


src/getfeatureinfo.js
```diff
@@ -24,7 +24,7 @@ const isValidJSON = str => {
  * @param {any} groupLayers
  * @returns {SelectedItem}
```
steff-o, line 25: Param `localization` should be commented and as optional .

src/models/SelectedItem.js
```diff
@@ -12,15 +12,17 @@ export default class SelectedItem {
    * @param {any} map
    * @param {any} selectionGroup
    * @param {any} selectionGroupTitle
+   * @param {any} localization
```
steff-o, line 15: Param `localization` should be commented and as optional .

steff-o (changes requested):
A bit picky, as its only concerns comments, but since both `SelectedItem` and `createSelectedItem` is exposed in the API I think all parameters should be documented. And some IDE:s built in linting can complain.

Other than that, it seems to do exactly what is is supposed to to.

By the way, I think I know why both `getAttributes()` and `getAttributesAsync()` are called. At least I should know as I was the one who wrote it. It is because of backwards compatibility reasons as SelectedItem is exposed in the api and the content is populated in the constructor, which can not be async. When getting attributes from related tables the operation has to be async. So I created getAttributesAsync to be called if you feel that you want to add related tables attributes. And featureInfo will always try to do that. I remember that I was not really happy with that solution, but it seemed the safest way.

steff-o (approved):
LGMT

PR #2179, opened 2025-04-11
fix: moved addLinkListener outside if-statement

### LINKED ISSUE #2177: Modal not working
**Describe the bug**
When you set `"target": "modal-full"` to an attribute, no modal is displayed, instead a new tab is opened in the browser.

**To Reproduce**
1. Add  `"target": "modal-full"`  to a url attribute
2. Click on link
3. A new tab is shown instead of a modal

**Expected behavior**
The old behaviour with showing a modal in full mode.

**Additional context**
It works in the 2.9.0 release.

johnnyblasta:
Fixes #2177

### DIFF
--- a/src/featureinfo.js
+++ b/src/featureinfo.js
@@ -439,16 +439,16 @@ const Featureinfo = function Featureinfo(options = {}) {
             break;
           }
         }
-        const modalLinks = document.getElementsByClassName('o-identify-link-modal');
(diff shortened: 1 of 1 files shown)


steff-o (approved):
LGTM.

I had a hard time reproducing this at first until I realized it was only broken for InfoWindow.

PR #2183, opened 2025-04-23
fix: restore media selectors

### LINKED ISSUE #2147: Cant reach infowindow on smaller units
**Describe the bug**
The infowindow: infowindow is out of reach on smaller units
**To Reproduce**
Fetch a fresh update of Origo

**Expected behavior**
The infowindow should be in center of screen or be available to move the modal

**Screenshots**
Image

### LINKED ISSUE #2155: Huge attribution icon

steff-o:
This PR replaces #2172 and should fix #2155 and fix #2147 as well. 

Restores the media selectors that disappeared in #2118. This PR produces an almost byte exact copy of the style.css as it was before #2107, with the exception of some comments and recent additions. ( I know I said that about #2118 as well, but I overlooked some significant whitespace changes in that one)

Almost all `@include meta.load-css` statements have been rewritten to mixins to preserve parent selectors. I didn't rewrite the ui scss files as they don't have parent selectors.

Also changed glide to use the precompiled css files instead of the scss files as they produced annoying warnings.

Also removed some unused scss files and gitignore the css.map-file.

Many thanks to @jokd for writing the first part in PR # 2172. I stole his parts instead of continuing on his fork.


Grammostola:
Seems to have quelled
- #2155 . The attribution icon looks reasonable again and the bottom bar isn't overly persistent
- #2147 . The infowindow appears in a centered visible position again

Also no more 
> DEPRECATION WARNING: Sass @import rules are deprecated and will be removed in Dart Sass 3.0.0

I would approve but I would rather let someone weller versed in CSS do that.

Grammostola:
I've changed my mind and will approve it. I've tested audio and video content in an overlay and basic functionality elsewise and seeing as how #2155 can be quite a nuisance on many mobile phones and #2147 too, and how the annoying warnings when running attempting to change branches (style.css.map) are gone this PR is at the very least a very good improvement and probably also bullet proof. (And if there is a stray bullet discovered later we can deal with that then) .. so lgtm

PR #2120, opened 2024-12-20
feature: more styling options with RegularShape

johnnyblasta:
Closes #2088 

image

Styling example:

```
"statusrubrik_filter_arrende": [
  [
    {
  "triangle": {
    "radius": 7,
    "stroke": {
      "color": "rgba(0,0,0,1)",
      "width": 1
    },
    "fill": {
      "color": "rgba(0,255,0,1)"
    }
    },
  "filter": "[VernacularName] == 'jättebalsamin'",
  "label": "jättebalsamin"
    }
  ],
  [
    {
  "star": {
    "radius": 7,
    "stroke": {
      "color": "rgba(0,0,0,1)",
      "width": 1
    },
    "fill": {
      "color": "rgba(0,255,0,1)"
    }
    },
  "filter": "[VernacularName] == 'kanadagås'",
  "label": "kanadagås"
    }
  ],
[
  {
"square": {
  "radius": 7,
  "stroke": {
  "color": "rgba(0,0,0,1)",
  "width": 1
  },
  "fill": {
  "color": "rgba(255,255,0,1)"
  }
  },
  "filter": "[VernacularName] == 'vresros'",
  "label": "vresros"
  }
],
[
  {
"cross": {
  "radius": 7,
  "stroke": {
  "color": "rgba(255,255,0,1)",
  "width": 2
  }
  },
  "filter": "[VernacularName] == 'blomsterlupin'",
  "label": "blomsterlupin"
  }
],
[
  {
"x": {
  "radius": 7,
  "stroke": {
  "color": "rgba(255,0,0,1)",
  "width": 2
  }
  },
  "filter": "[VernacularName] == 'lärksopp'",
  "label": "lärksopp"
    }
]
]
```

PR #2184, opened 2025-04-24
fix: default fill in legend symbol

### LINKED ISSUE #2160: Style in legend default to black fill when no fill is set
**Describe the bug**
When a style in the legend has no fill specified, it incorrectly renders with a black fill instead of being transparent. This affects the visual representation of styles in the legend for both lines and polygons.

**To Reproduce**
Set a line-layer with a stroke-style or just use the index.json, change layer origo-mask to "group": "root", and delete fill in mask so it looks lika below.
`  "mask": [
      [
        {
          "stroke": {
            "color": "rgba(0,0,0,1.0)"
          }
        }
      ]
    ],`

**Expected behavior**
Should be transparent, at least for line based layers.

**Screenshots**
Image

**Additional information**
It looks like PR #2120 introduced the bug.
Not sure if it's related at all, but the stroke lineDash in the legend now appears with bigger gaps.

steff-o:
Fixes #2160.

A tricky one this one... Old implementation used `Circle()` to render a legend icon for symbols of type `Line` and `Polygon`. #2120 redefined some defaults for Circle() to match the defaults for OpenLayers' defaults for the regular shape circle in the map. But when called to generate an icon for a polygon or line symbol there shouldn't be any default fill.

I solved it by separating Polygon and Line icon functions from Circle. The diff may look a bit strange but there was already a function called Polygon and one called Line. They were never called, but had to exist as legendmaker checked for their existance against the _styleType_. I copied the old implementation of `Circle()` to `Polygon()` and had `Line()` call `Polygon()` to restore the behavior for lines and polygons without affecting the regular shape "Circle".

In short the _styleType_ is determined from the top level symbol properties. If it has stroke it is a line. If it has both stroke and fill it is a Polygon and if it has a named regular shape, it is that shape. My guess is that the intention was to render different icons for line and polygons, like a snake for lines and an irregular patch for polygons, but in reality they were all circles. With this separation it would be easy to create different svg:s with different shapes depending on the styleType, but since that is a feature request, I didn't do it in this PR. But it would be cool to do so. Maybe give the chance to select from a variety of shapes, like smooth curved lines for water, straight corners for streets, smooth patches for lakes, squared patches for land parcels etc.

### DIFF
--- a/src/utils/legendmaker.js
+++ b/src/utils/legendmaker.js
@@ -98,7 +98,7 @@ export const renderSvgIcon = function renderSvgIcon(styleRule, {
   if (styleType in renderIcon) {
     if (styleType === 'Polygon') {
       const polygonOptions = styleRule.find(style => style.fill);
-      const icon = renderIcon.Circle({
+      const icon = renderIcon.Polygon({
         fill: polygonOptions.fill,
         stroke: polygonOptions.stroke
       });
@@ -106,7 +106,7 @@ export const renderSvgIcon = function renderSvgIcon(styleRule, {
     } else if (styleType === 'Line') {
       const icon = styleRule.reduce((prev, style) => {
         if (style.stroke) {
-          return prev + renderIcon.Circle({
+          return prev + renderIcon.Line({
             stroke: style.stroke
           });
         }
--- a/src/utils/legendrender.js
+++ b/src/utils/legendrender.js
@@ -12,36 +12,44 @@ export const renderSvg = function renderSvg(content, {
 };
 
 export const renderIcon = {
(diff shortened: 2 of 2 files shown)


johnnyblasta (approved):
LGTM and good solution to separate the types so opens up the possibility to render the icons differently.

steff-o:
@owemyr do you want to test this?

Owemyr:
> @Owemyr do you want to test this?

@steff-o tested it now, looks good with my configs, dashed line is also like before, thanks and nice how this opens up for more styling types in the future.

Issue #2193, opened 2025-05-07
Add value and label to dropdown


Grammostola:
The dropdown component is an appealing dropdown/up ui component.
Presently it takes an array of strings and constructs items like so:
```js
      listItems.forEach((listItem) => {
        const itemEl = El({
          tagName: 'li',
          innerHTML: `<span>${listItem}</span>`
        });
```
This appears to mean that what you see is what you get (what you can get from the selected dropdown item), rather than a label:value approach, and indeed this seems to be how dropdown is currently employed around the app, in the print and scalePicker controls for instance.

I propose that we redefine the dropdown component slightly to be able to accept an array like 
```js
[
  {
   label: "1:100,000",
   value: 100000
   },
  {
   label: "1:200,000",
   value: 200000
   }
```
This would make sense when attempting to localize the scale numbers of any dropdown with map scales (and it can be employed for other things too, like an html select).

It would of course need to be 100% backwards compatible, meaning should still be able to take an array of strings as well.

PR #2189, opened 2025-04-28
feature: localize scalepicker

Grammostola:
Aims to fix #2141

PR #2194, opened 2025-05-07
fix: make scale selectors work again

### LINKED ISSUE #2192: Scale selectors issue
The print scale selector does not work (well, it depends a bit on what language the browser is set to, which shouldn't be the case) 

The utils/numberformatter.js was recently updated, to provide a human formatted number (as that seemed the point but as I reflect it needn't have been, entirely. The function has no explanation and it performs a slight rounding of numbers), however the print-component uses
```js
scales = originalResolutions.map(currRes => maputils.resolutionToFormattedScale(currRes, viewer.getProjection()));
```
And that seems to be to show as well as to employ.

It needs to be fixed so that un"formatted" (but rounded) numbers are employed (as before) and localized scale numbers are shown in the selector.

Edit. Further investigation reveals that tellingly, the scalepicker control works worse with US style numerals too.  I think the dropdown component has a weakness in that it doesn't provide a label:value storage, it uses the same string for values and showing, unlike a `select` for instance. 

I also think that this is a separate issue and that more promptly the scalePicker needs to be mostly reverted to its unlocalized form and the numberFormatter needs to treat no localization object param as Do not localize this. I will try and arrange both things.

Grammostola:
This proposed change reverts some of the localization from the scalepicker control and makes the numberformatter not attempt to localize the number it produces if no localization object accompanies the invocation.

The target is to make print's and the scalepicker's scale selectors work again (regardless of the currently chosen language) and as such fix #2192 

(To localize the scale selector numbers I believe #2193 needs to be remedied first)

### DIFF
--- a/src/controls/scalepicker.js
+++ b/src/controls/scalepicker.js
@@ -18,7 +18,7 @@ const Scalepicker = function Scalepicker(options = {}) {
   }
 
   function getScales() {
-    return resolutions.map(resolution => `${listItemPrefix}${mapUtils.resolutionToFormattedScale(resolution, projection, localization)}`);
+    return resolutions.map(resolution => `${listItemPrefix}${mapUtils.resolutionToFormattedScale(resolution, projection)}`);
   }
 
   function setMapScale(scale) {
@@ -29,7 +29,7 @@ const Scalepicker = function Scalepicker(options = {}) {
   }
 
   function onZoomChange() {
-    dropdown.setButtonText(`${buttonPrefix}${mapUtils.resolutionToFormattedScale(map.getView().getResolution(), projection, localization)}`);
+    dropdown.setButtonText(`${buttonPrefix}${mapUtils.resolutionToFormattedScale(map.getView().getResolution(), projection)}`);
   }
 
   return Component({
--- a/src/utils/numberformatter.js
+++ b/src/utils/numberformatter.js
@@ -8,7 +8,10 @@ export default function numberFormatter(numberToFormat, localization) {
     nr = Math.round(nr / factor) * factor;
   }
 
-  // Format acc to locale
-  const formatter = localization?.getCurrentLocaleId() ? new Intl.NumberFormat(localization?.getCurrentLocaleId()) : new Intl.NumberFormat();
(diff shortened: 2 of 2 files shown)


steff-o (commented):
I have tested this PR and from what I can see the scale picker works and when zooming in print preview the scale picker gets updated. The scale picker does not have any thousand delimiter at all.

If I revert to master, the scalepicker does not work if I change locale from swedish to US in the browser. It makes no difference if I swich locale in Origo. The scalepicker will in this case use comma as thousand delimiter.

If I have locale SE in browser master works, but regardless of Origo locale setting scale is formatted with space as thousand delimiter.

Never tested, but by code inspection before localization was introduced the scalepicker would always use space as thousand delimiter no matter what.

So, in conclusion, this PR works, but the behaviour is not the same as before localization as there is no thousand delimeter at all now. Also the return value from numberFormatter will be `number` if no localization object is provided. That is probably not a problem as it will implicitly be cast to `string` when used and I can't find any other usage of this function.

A different approach would be to always provide the localization object to numberFormatter and change the parsing in scalepicker's `setMapScale()`. As there is no parseInt in javascript that uses locale settings that has to be written, and doing a so in a general way is of course pretty complex. But it would be possible to simply replace everything that is not a digit and hope for the best. It wouldn't work for hindi numbers and stuff, but we would cover the only two locales that is currently implemented.

If this is just a temporary solution until there is a value-label version of dropdown it may be OK and not worth fixing further.

Thoughts?

Grammostola:
I think that you have correctly summarized the situation. 

With #2189 the scale number in the scalepicker went from hard-coded S style to Intl.NumberFormat() if no localization object, so browser locale, and with this proposed fix there is no formatting if no localization object which fixes the broken functionality but as you say doesn't look optimal in the scalepicker.

It is a temporary fix until presumed #2195 is in place and the scalepicker can be revised to get its labels (only) formatted (mainly thousand delimiters but as you say there might be other things and with #2195 it won't matter if the delimiter is a collection of bats).

steff-o (approved):
OK. You convinced me, it will do for now.

Scalepicker is (almost) great again.

PR #2195, opened 2025-05-07
feature: make dropdown list able to support a label and a value

### LINKED ISSUE #2193: Add value and label to dropdown

Grammostola:
Aims to fix #2193 by introducing a `data-` attribute to the dropdown item element.

The scalepicker could then do this 
```js
dropdown.setItems([
 { label: '1:100,000',
   value: 1000
 },
 {
  label: '1:25,000',
  value: 25000
 }
 ])
```

instead of this
```js
dropdown.setItems([
'10,000',
'25,000'])
```

And listeners for dropdown select actions:
```js
.addEventListener('dropdown:select', (evt) => {
```
could access the value via the `evt.detail` prop.

For scale selectors this would enable us to localize the numbers without breaking their functionality depending on what locale were chosen.

Draft PR because linked issue not classified (no pr accepted status yet)

### DIFF
--- a/src/ui/dropdown.js
+++ b/src/ui/dropdown.js
@@ -31,11 +31,13 @@ export default function Dropdown(options = {}) {
 
   const style = createStyle(styleSettings);
 
-  const selectItem = function selectItem(item, doClick = true) {
+  const selectItem = function selectItem(itemEl, doClick = true) {
+    const value = document.getElementById(itemEl.getId()).getAttribute('data-value');
     const customEvt = new CustomEvent('dropdown:select', {
-      bubbles: true
+      bubbles: true,
+      detail: value // Pass the value in the event detail
     });
-    document.getElementById(item.getId()).dispatchEvent(customEvt);
+    document.getElementById(itemEl.getId()).dispatchEvent(customEvt);
     if (doClick) dropdownButton.dispatch('click');
   };
 
@@ -51,15 +53,23 @@ export default function Dropdown(options = {}) {
   };
 
   const setItems = function setItems(listItems) {
-    items = listItems;
+    // Check if listItems are strings and if so then convert them to objects for backward compatibility
+    items = listItems.map(item => {
+      if (typeof item === 'string') {
+        return { label: item, value: item };
+      }
+      return item;
+    });
+
     const contentEl = document.getElementById(contentComponent.getId());
     if (contentEl) {
       contentComponent.clearComponents();
       contentEl.replaceChildren();
-      listItems.forEach((listItem) => {
+      items.forEach((listItem) => {
         const itemEl = El({
           tagName: 'li',
-          innerHTML: `<span>${listItem}</span>`
+          innerHTML: `<span>${listItem.label}</span>`, // Use label for display
+          attributes: { data: { value: listItem.value } } // Employ the data prop for the value (will become a data-value attribute)
         });
         contentComponent.addComponent(itemEl);
         contentEl.appendChild(html(itemEl.render()));


jokd:
This is a great addition. But while at it, can it be useful to allow an arbitrary number of data-attributes to set instead of just 'value'? Haven't got any usecase where it is useful but it's just a little tweak to make it work.

Grammostola:
Hi, would this be for fetching them from the dropdown list item element via the document, rather than receiving additional info via the custom event? The custom event has the single prop `detail` for attaching whatever afaik, could be an array of objects?

jokd:
I guess I was thinking that the event detail could be an object with all the element attributes instead of a single value, and then it is up to the listener to handle it.

Grammostola:
Would this work for your intended purpose?

jokd:
Was rather thinking like 
`attributes: { data: listItem}`
or delete the label key first:
`attributes: { data: Object.fromEntries(Object.entries(listItem).filter(([key]) => key !== 'label'))}`

But I haven't tested if this works with the elementcreator in the ui

Grammostola:
I see. In both cases (your suggestion) and the PR as it currently stands the detail prop can send back complex things, I suppose the difference lies in the fact that your suggestion the detail prop will always be an object, in mine it's rather up to what you supplied as a `value` for setItems().

There can be a difference in specifying the listItems too, if that matters. I can't presently see a clear benefit to either method.

jokd:
Whatever works will be great imo

Grammostola:
Currently the suggestion (PR) works like this:

the array argument to `dropdown.setItems()` can be:
```js
      const scaleObjects = [
        {
          label: '1:100',
          value: 100
        }, {
          label: '1:200',
          value: 200
        }, {
          label: '1:300',
          value: 300
        }, {
          label: '1:400',
          value: 400
        }, {
          label: '1:500',
          value: 500
        }
      ];
```
and the listener
```js
document.getElementById(dropdown.getId()).addEventListener('dropdown:select', (evt) => {

```
can then access `evt.detail` and get 
```js
    [
        {
          label: '1:100',
          value: 100
        }, {
          label: '1:200',
          value: 200
        }, {
          label: '1:300',
          value: 300
        }, {
          label: '1:400',
          value: 400
        }, {
          label: '1:500',
          value: 500
        }
      ]
```
and same as before the label via `evt.target.textContent`

Same thing if the array array argument is.

```js
    [
        {
          label: 'jaggedArray',
          value: [
            [4, 4, 3, 2],
            [{ id: 1, color: 'yellow' }, { id: 2, color: 'blue' }],
            ['cats']
          ]
        }
      ]
```

It can also handle
```js
[100, 200, 300, 400, 500]
['1:100', '1:200', '1:300', '1:400', '1:500']
```
the `label` will then be the same as the `value`

Grammostola:
This does require some creativity in order to test but what this PR should not do is break existing functionality, so after its application the scalepicker for instance should still work like it did and the same for draw and editor dropdowns.

After it has been (provided that it will be) merged I will redo the localization of the scalepicker control and make it avail of said changes.

johnnyblasta (commented):
Tested that it doesn't break current function, which it doesn't.

steff-o:
I'm not fully convinced. The way it seems to work is that after an option has been selected, the event handler that detects the change also calls `setButtonText()` using the event detail, which is the backing value, not the label. For this approach, the code that sets up the dropdown must have the capacity to look up the label for the selected value, as the event only contains the value, not the label. Maybe this is not the intended usage of setButtonText, but that is how `title-control` does. Also the dropdown control does not set it by itself, maybe to enable the caller to select the text as something completely different, like the name of property that is changed by that dropdown.

Right now it works however, as no calling code uses value, label other than those that are automatically created, so they will always be the same. But when converting to label, value configuration the caller must also know what to set. Ideally the event should also contain the label, but that will break the event interface. The title-control does (unintentionally) by magic read the label as it in fact does not use the event detail at all, it uses the events `target` property to get the selected item from the DOM, which is exactly the label. Not sure where this is going, maybe there should exist a method to set the button text from value, or let caller use the event's `target`trick, or let caller figure it out by itself as is holds the configuration.

Also to be more backwards compatible the code could look like:
```javascript
      if (typeof item === 'object' && item.hasOwnProperty('label') && item.hasOwnProperty('value')) {
        return item;
      }
      return { label: item, value: item };
```
There is a slim chance that someone has used booleans or some other datatype that could be serialized automatically other than string and number.

Grammostola:
Thanks for your feedback. I'm not completely sure that I follow.

The title-control component of print uses
```js
      document.getElementById(selectSize.getId()).addEventListener('dropdown:select', (evt) => {
        this.onChangeSize(evt.target.textContent);
      });
```
and 
```js

    onChangeSize(evt) {
      this.dispatch('change:titleSize', { class: evt });
      selectSize.setButtonText(evt);
      titleSize = evt;
    },
```

the scalepicker control does the same thing (at least when the dropdown is employed to change the current resolution rather than zooming in the map)

This continues to work with the proposed change.

I concur with the suggested type check change. If someone employs objects with other props, well they would need to amend their ways.

steff-o:
Yes of course, a rewrite would be necessary anyway, but I would prefer that the drop down component's event contains both label and value to avoid the caller to be aware of the technical details that the component uses to store the label in the DOM. Although the caller can fetch the label from DOM (the way it is done now) or do a lookup from its own configurations, I think it would be clearer if the dropdown just emits both label and value.

PR #2205, opened 2025-05-21
fix: backgroundfill for text in draw

### LINKED ISSUE #2198: No background fill for text in draw
**Describe the bug**
It's not possible to add background fill on text in draw control

**To Reproduce**
1. Opn draw control and add a text
2. Try to add a background fill to the text
3. None is added

**Expected behavior**
As it was before a background colour should be possible to add so that the contrast against the text can be controlled.

**Screenshots**
Text without background even though it's specified:
Image

johnnyblasta:
Fixes #2198 
image

### DIFF
--- a/src/style/stylewindow.js
+++ b/src/style/stylewindow.js
@@ -337,7 +337,7 @@ const Stylewindow = function Stylewindow(optOptions = {}) {
     newStyleObj.pointSize *= styleScale;
     newStyleObj.backgroundStrokeColor = stringToRgba(newStyleObj.backgroundStrokeColor, newStyleObj.backgroundStrokeOpacity);
     newStyleObj.backgroundStrokeWidth *= styleScale;
-    newStyleObj.backgroundFill = stringToRgba(newStyleObj.backgroundFillColor, newStyleObj.backgroundFillOpacity);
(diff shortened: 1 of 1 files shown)


steff-o (approved):
It seems to to what it is supposed to do, but I can't get rid of the feeling that it could have been done in a more readable way.

The thing is that `backgroundFillColor` and `backgroundFillOpacity` is not roundtripped in the `origostyle` attribute, so it won't be copied when restoring the style. To be honest, I can't find anywhere where `backgroundFillColor` and `backgroundFillOpacity` could be set on the feature and can only appear from the `swDefaults` assignment.

The `fillColor` property works in a similar way, but in that case the `fillColor` property is used for both the `rgb` and `rgba` notation, which makes it work, but the `fillColorOpacity` is only used when defaulting as `fillColor` is roundtripped in `origostyle` as `rgba`, which already contains opacity. In contrary, the `backgroundStrokeColor` is in `rgb` notation and is accompanied by `backgroundStrokeOpacity` in order to be able to restore it as `rgba` (by reusing the `backgroundStrokeColor` property as `rgba` notation). So there are three different approaches to achieve the same goal.

In my opinion it would have been a lot easier to maintain the code if only `rgba` notation (or rgb + opacity) was used throughout the code, but I guess it would be too much work and risk to clean it up.

PR #2209, opened 2025-06-19
Use icon rotation from attribute if referenced

### LINKED ISSUE #2208: Rotate feature icons by attribute
**Description**
When styling a point layer with icons, it would be great if an attribute could be referenced and have the icon rotated according to the attribute value.

**Proposed solution**
A simple check if the "rotation" property of an icon style is a string and if it includes double curly brackets, in which case the rotation value of that attribute should be used instead.

**Alternatives considered**
Adding a custom style function, but this seems like a useful addition for Origo's default behaviour.

**Additional context**

Image

MattiasSp:
Fixes #2208.

Allows setting an icon style like so;
```
{
  "label": "Origokommuner",
  "icon": {
    "size": [32,32],
    "src": "img/geolocation_marker_heading.png",
    "rotation": "{{rotation_attribute}}"
  }
}
```

If the referenced attribute does not exist, the rotation defaults to 0.

### DIFF
--- a/src/style.js
+++ b/src/style.js
@@ -333,7 +333,7 @@ function checkOptions(options = {}) {
         }
         if (element.icon && Object.prototype.hasOwnProperty.call(element.icon, 'rotation')) {
           let degrees;
-          const rotationFromAttribute = typeof element.icon.rotation === "string" && element.icon.rotation.match(/{{(.+)}}/i)[1];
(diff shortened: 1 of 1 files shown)


tonnyandersson:
What is the difference between this and the existing rotation feature?

MattiasSp:
> What is the difference between this and the existing rotation feature?

None it seems >< Closing the issue and PR. I will add a code example to the documentation instead to make the functionality more visible.

PR #2211, opened 2025-06-26
Add configuration option to icon styles: referenceMapScale, minScale, maxScale

### LINKED ISSUE #2210: Scale feature styles according to a reference map scale and generalize rotation/scale handling for text, icon and circle
**Description**
Sometimes it is desirable to make icons/text/circles/etc smaller as the user zooms out and larger as they zoom in. This could make the map more readable. Other times users want the style to be locked to a specific size in map units, e g a constant diameter of X meters.

**Proposed Solution**
Setting style configuration parameters to control:
 - a reference map scale
   - e g `"referenceMapScale": 10000` to show the icon in its normal size at map scale 1:10 000
 - a minimum size (scale)
   - e g `"minStyleScaleFactor": 0.5` to never show the icon smaller than half its normal size
 - a maximum size (scale)
   - e g `"maxStyleScaleFactor": 3` to never show the icon larger than three times its normal size

**Alternatives Considered**
The users can draw polygons instead to have them keep their true sizes, but this might be much more inconvenient.

**Additional context**

Image

EDIT: Updated the proposed names for the min- and max icon scale factors.

EDIT 2: Updated the names and description again to extend the issue to also generalize the scale and rotation handling so that it works the same for text, icon, circle and other basic shape styles.

MattiasSp:
Fixes #2210.

Lets icons grow and shrink when a user zooms in and out in the map.
<pre>
"origo-logo": [
      [
        {
          "label": "Origokommuner",
          "icon": {
            "src": "img/kompass.svg",
            <b>"referenceMapScale": 50000,
            "minScale": 0.01,
            "maxScale": 0.2</b>
          }
        }
      ]
    ],
</pre>

- _referenceMapScale_ defines the map scale at which the icon will be displayed at its normal size (provide the map scale denominator as a number, e g `10000` for a reference map scale of 1:10 000).
- _minScale_ defines how small the icon is allowed to be relative to its normal size.
- _maxScale_ defines how large the icon is allowed to be relative to its normal size.

Also, if the `"updateWhileAnimating": true` property is added to a layer with an icon style with a referenceMapScale will make the scaling of the icons smooth as the user zooms:
<pre>
"layers": [
    {
      "name": "origo-cities",
      "title": "Origokommuner",
      "group": "root",
      "style": "origo-logo",
      <b>"updateWhileAnimating": true,</b>
      "type": "GEOJSON",
      [...]
</pre>

### DIFF
--- a/src/style.js
+++ b/src/style.js
@@ -336,6 +336,16 @@ function checkOptions(options = {}) {
           const radians = degrees * (Math.PI / 180);
           styleList[j][index].getImage().setRotation(radians);
         }
+        if (element.icon?.referenceMapScale) {
+          let iconScale = element.icon.referenceMapScale / scale;
+          if (element.icon.maxScale) {
+            iconScale = Math.min(iconScale, element.icon.maxScale);
+          }
+          if (element.icon.minScale) {
+            iconScale = Math.max(iconScale, element.icon.minScale);
+          }
+          styleList[j][index].getImage().setScale(iconScale);
+        }
         return null;
       });
       if (Object.prototype.hasOwnProperty.call(s[j][0], 'filter')) {


MattiasSp:
Now the above also works for text or circle styles, e g:
```
{
  "text": {
    "font": "Bold 12px Arial",
    "textAlign": "left",
    "offsetX": 11,
    "offsetY": -9,
    "text": "{{name}}",
    "fill": {
      "color": "rgba(0,0,0,1.0)"
    },
    "stroke": {
      "color": "rgba(255,255,255,0.7)",
      "width": 2
    },
    "referenceMapScale": 10000000,
    "minStyleScaleFactor": 0.4,
    "maxStyleScaleFactor": 3,
    "scale": "{{scale}}",
    "rotation": "{{rotation}}"
  }
}
```

PR #2214, opened 2025-07-25
fix: localize infowindow area and length sum functions

### LINKED ISSUE #2203: Aggregation results are not localized in InfoWindow
**Describe the bug**
Similar to #2164 the area and length attribute functions are not displayed in the currently selected locale's number format. It also applies to any arbitrary attribute that is aggregated.

**To Reproduce**
1. Use the demo application
2.  Configure featureInfo to use infowindow as infowindow:
``` 
   "featureinfoOptions": {
        "infowindow": "infowindow",
        "infowindowOptions": {
            "groupAggregations": [
                               {
                    // No layer specified, apply to all (multi) polygon layers 
                    "function": "sum",
                    "attribute": "@area",
                    "label": "<b>Total area:</b> "
                }
            ]
        }
    },
```
3. Configure origo mask to be queryable
4. Configure the test application to use localization
5. Change locale in Origo to a different than your browsers.
6. Click anywhere on the mask layer
7. The sum does not use the selected locale's number format

**Expected behavior**
The sum should use the the selected locale's number format.

**Screenshots**
Image
This image is taken from when #2174 is in progress, so the area attribute for the mask feature obeys the localization setting, but the aggregation does not.

**Additional context**

Grammostola:
Aims to fix #2203

### DIFF
--- a/src/selectionmanager.js
+++ b/src/selectionmanager.js
@@ -13,7 +13,8 @@ const styleTypes = StyleTypes();
 
 const Selectionmanager = function Selectionmanager(options = {}) {
   const {
-    toggleSelectOnClick = false
+    toggleSelectOnClick = false,
+    localization
   } = options;
 
   let aggregations = [];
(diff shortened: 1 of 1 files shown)


johnnyblasta (approved):
Tested and LGTM

PR #2217, opened 2025-08-20
feature: search on groups in legend search

### LINKED ISSUE #2216: Legend search should include groups
When searching for a layer in legend search the groups are not included, which can cause a confusion when the group name might include words explaining what the layers in that group is about but  the layers them self don't have someof toose words in it's name.

The groups should be included and somehow make it clear that this is a group.

Have more keywords in the abstract so the layers can be found.

johnnyblasta:
Closes #2216

### DIFF
--- a/src/controls/legend.js
+++ b/src/controls/legend.js
@@ -408,15 +408,17 @@ const Legend = function Legend(options = {}) {
         if (layerGroup.exclusive) { // If group is exclusive only make the first layer visible.
           for (let i = 0; i < layers.length; i += 1) {
             if (i === 0) {
-              layers[i].setVisible(true);
+              if (!layers[i].get('secure')) {
+                layers[i].setVisible(true);
+              }
             } else {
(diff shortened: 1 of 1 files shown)


Grammostola:
Nice. It seems to work well, which includes turning on all layers in the selected search hit (group) and not including the layers of any subgroup. And it finds subgroups too. 
- For `exclusive` the last (bottom) layer of the group is activated (maybe the top layer would be more intuitive but I don't have a real opinion here). 
- For `toggleAll: false` subgroups, it activates all their layers. Have no opinion here either since the toggleAll button is gone and maybe that is the specific point of `toggleAll: false`

steff-o (approved):
LGTM!

One could argue that if the first layer in an exclusive group is secure, no layer in the group will be selected, but it seems strange to have an exclusive group where you have access to some layers but not all . But the expected behavior in such case is not obvious anyway and could very well be that no layer should be visible.

PR #2229, opened 2025-09-29
feature: localize splash control

### LINKED ISSUE #2139: Localize Splash control
Some strings has to be localized.

johnnyblasta:
Closes #2139

### DIFF
--- a/src/controls/splash.js
+++ b/src/controls/splash.js
@@ -1,7 +1,16 @@
 import { Component, Button, Modal } from '../ui';
 
 const Splash = function Splash(options = {}) {
-  const defaultTitle = 'Om kartan';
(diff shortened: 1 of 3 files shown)


johnnyblasta:
I suggest keeping configurable strings. If having own preffered strings in the loc-files they have to be in a local copy which has to be considered every time merging from Origo master.

Grammostola:
Yep, and it depends a bit on how we update them in the future but I don't think it could become very complex since it would only ever be a language file.  I think it should be touted as a feature : ) In part because not all or even a majority of the ui strings I would hazard to guess are configurable.

We shall keep the configurable ones then, I will update #2222 

lgtm

PR #2233, opened 2025-10-03
fix: check if layer is secure before making visible

### LINKED ISSUE #2231: Secure layer on when whole group on
**Describe the bug**
When turning on a group which contains a secure layer the secure layer is turned on and the padlock disappears and can't be switched off. The layer obviously isn't shown, but causes network traffic which is stopped by 403 error.

**To Reproduce**
1. Go to a group which has secured layer
2. Click on turn all layer for group.
3. Scroll down to the secured layer
4. See that padlock has disappeared and checkbox icon visible.

**Expected behavior**
The secured layer shouldn't be turned on and therefore the padlock icon would still be visible.

**Screenshots**
Image

johnnyblasta:
Closes #2231

### DIFF
--- a/src/controls/legend/group.js
+++ b/src/controls/legend/group.js
@@ -355,7 +355,9 @@ const Group = function Group(viewer, options = {}) {
           const overlays = groupList.getOverlays();
           overlays.forEach((overlay) => {
             const layer = overlay.getLayer();
-            layer.setVisible(true);
+            if (!layer.get('secure')) {
+              layer.setVisible(true);
+            }
           });
           const groups = groupList.getGroups();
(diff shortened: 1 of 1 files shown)


johnnyblasta:
> Will this interfere with/solve the same problem/complement the solution in #2217 or are there two completely different implementations of turning on a layer?

This is a complement to the other solution, which has another separate way of turning on a layer.

steff-o (approved):
LGTM

It fixes that exact problem. While testing I saw that if a layer is default visible and padlocked, it is displayed as visible and stuck as visible. But that was beyond the scope of this PR as described in the issue.

PR #2235, opened 2025-10-13
feature: handle nested objects when getting attribute values

### LINKED ISSUE #2234: Attribute should handle nested JSON object
If for example a GeoJSON is used as a layer that has a nested JSON object in the attributes the value should be able to find.

If an attribute is defined like this `{"name": "detaljplan.beteckning"}` and the attribute is a JSON object like below it should be possible to get the attribute value by drilling down in the object.
```
{
    "detaljplan": {
        "objektidentitet": "0490dfc9-8dxf-4992-bcbb-9ed8ab6137d9",
        "beteckning": "2281K-DP-xx",
        "namn": "xxxx",
        "status": "laga kraft",
        "typ": "detaljplan",
        "datumLagakraft": "1991-12-05",
        "kvalitet": {
            "korrigeradeGranser": true,
            "kontrolleratPlaneringsunderlag": false
        },
        "anvandbarhet": "god"
    }
}
```

The GeoJSON could be flattened, so all values have a key

johnnyblasta:
Closes #2234

### DIFF
--- a/src/utils/getnestedvalue.js
+++ b/src/utils/getnestedvalue.js
@@ -1,3 +1,13 @@
+/**
+ * Get value for attribute key and drill down to the last key part of the attribute if it is a nested JSON object
+ *
+ * @function
+ * @name getNestedValue
+ * @kind variable
+ * @param {any} feature
+ * @param {any} attributeKey
+ * @returns {string}
+ */
 const getNestedValue = function getNestedValue(feature, attributeKey) {
   let attributeKeyValue = '';
   // Handle if foreignKey is nested json object


johnnyblasta:
> To sum what we discussed during our meeting then, merge with proposed similar function "searchjson" https://github.com/origo-map/origo/pull/2224/files#diff-73a377a3fa3092fb30d691c5c3145af68f00519db030398cef9a51111b1566d5 , as for jsdocs for getContent I'm thinking this pr doesn't necessarily need to include them?

Tested using the "searchjson" function, but found it quiet different and had difficulty implementing using that function.

PR #2238, opened 2025-10-27
Fix/draw export of circle geometries 2052

### LINKED ISSUE #2052: Draw: Export of circle geometries
**Describe the bug**
When exporting a circle geometry the resulting feature has no geometry. 

**To Reproduce**
1. Draw a circle
2. Export to file

**Expected behavior**
A circle round and clear.

**Additional context**
Circles are probably not stnadard so convert store the center and radius as properties or convert the circle to a polygon on export. One downside of the latter is you won't be able to edit the circle since it is made up of multiple short segments.

sweco-se1g7y:
Added a conversion of circles to polygons when exported as geojson file or as part of mapstate, and conversion back to circle when imported in the same way. The polygon version will thus never be used, but will ensure that the geojson file remains valid and can be used by other services if needed.

Resolves #2052

### DIFF
--- a/src/utils/circleFeatureNormalizer.js
+++ b/src/utils/circleFeatureNormalizer.js
@@ -1,12 +1,17 @@
 import { Polygon, Circle } from 'ol/geom';
 import { transform } from 'ol/proj';
 
-// Adds circle info to feature properties for later recreation
+/**
+ * Adds circle info to feature properties for later recreation.
+ *
+ * @param {*} feature The feature to add circle info to.
+ * @param {*} center The center coordinate of the circle.
+ * @param {*} radius The radius of the circle.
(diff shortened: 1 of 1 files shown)


Grammostola:
These are circles in QGIS / other applications presumably since this pr interprets the geojson polygon and (re)creates a https://openlayers.org/en/latest/apidoc/module-ol_geom_Circle-Circle.html which won't look jagged in Origo

Grammostola:
I've had a look and it looks like the brought up points have been seen to. It plays well with drag and drop, reimporting in the draw tool, opening in a gis desktop application. Id seems preserved throughout and coordinates are of the map's system (mapstate) and 4326 for geojson, as is its wont. Can't find anything to complain about really so I suggest we merge it.

Issue #2239, opened 2025-10-28
Let the user set custom scales when printing


sweco-seguro:
**Is your feature request related to a problem? Please describe.**
If a user wants to print in a scale that is not pre-configured there is currently no way to do it, except to change the configuration.

**Describe the solution you'd like**
Add a "valfri skala"-alternative to the dropdown, which shows an input field where a custom scale can be input when it is selected.
If a pre-configured scale is chosen the input field is hidden again.

**Describe alternatives you've considered**
Make the scale-selector a combined input/dropdown field, where you can either input a custom value or choose from the pre-configured list.

**Additional context**
Some mockups.

<img width="389" height="1012" alt="Image" src="https://github.com/user-attachments/assets/bd70c810-1df2-4e35-b243-dcb9302aa215" />

<img width="388" height="1010" alt="Image" src="https://github.com/user-attachments/assets/5771f17c-f836-4bfc-9a4e-db4fd199d386" />

Grammostola:
This was discussed during our dev meeting and we propose a prop for the print control to activate this functionality because sometimes typically a background map might be cached at specific zoom steps and asking for intermediate resolutions might not work so well. This potential pitfall already exists since #1938 and when enabling map interactions in the print preview and mousewheelzooming. I'm not aware of any negative feedback so far but we still think it makes sense to implement the functionality as toggleable and not on by default.

PR #2254, opened 2025-11-18
Feature: Let the user set custom scales when printing

sweco-seguro:
Added the feature described in https://github.com/origo-map/origo/issues/2239

If the print control option "userDefinedScale" is true a user can now input any scale they wish, with the constraint that it's within the min-max resolution of the map (similar to https://github.com/origo-map/origo/pull/1938)

### DIFF
--- a/src/loc/sv_SE.json
+++ b/src/loc/sv_SE.json
@@ -70,7 +70,7 @@
             "portraitAriaLabel": "Stående",
             "landscape": "Liggande",
             "landscapeAriaLabel": "Liggande",
-            "userDefinedScale": "Mata in egen skala"
+            "userDefinedScale": "Ange din egen skala:"
         },
         "draw": {
             "title": "Rita",


johnnyblasta (commented):
Works, but crashes opening print when `"userDefinedScale": true` option isn't set in config:

```
set-scale-control.js:96 Uncaught TypeError: Cannot read properties of undefined (reading 'render')
    at Object.render (set-scale-control.js:96:1)
    at printTemplate (print-settings.template.js:33:1)
    at Object.render (print-settings.js:203:1)
    at Object.render (collapse.js:135:1)
    at Object.render (print-settings.js:268:1)
    at Object.render (print-component.js:728:1)
    at Object.click (index.js:162:1)
    at eval (eventer.js:20:1)
    at Array.forEach (<anonymous>)
    at Object.dispatch (eventer.js:20:1)
```

PR #2255, opened 2025-11-20
fix: check for secure on default visible layer

### LINKED ISSUE #2251: Secured layer but default visible
**Describe the bug**
If a layer is set is visible in config, but the layers is secured and therefore not accessible to the user the icon in legend is the checkbox and not the padlock.

**To Reproduce**
1. Define in config a layer that you don't have access to as default visible.
2. Check the legend and the icon is the checkbox.
3. The leayer can't be turned on or off.

**Expected behavior**
The layer should have been padlocked in the legend

johnnyblasta:
Fixes #2251


steff-o (approved):
LGTM

PR #2257, opened 2025-11-20
fix: null check on draw control

### LINKED ISSUE #2256: Crash when no draw control is configured
**Describe the bug**
Origo crashes on start up if no draw control is configured

**To Reproduce**
1. Start demo app
2. See error in console. No map is visible

**Additional context**
Bug was introduced in #2215

steff-o:
Resolves #2256 by null checking the draw control search result when initializing _StyleWindow_


johnnyblasta (approved):
LGTM

Issue #1977, opened 2024-03-12
Support for OGC API Features


johnnyblasta:
**Is your feature request related to a problem? Please describe.**
In the future, more people will choose to publish in the standard OGC API Features instead of WFS, e.g. there are already detailed plans from the Lantmäteriet in that format and we should be able to support it.

**Describe the solution you'd like**
A new layer type for OGC API Features/STAC (WFS 3).

**Describe alternatives you've considered**
Extend WFS to handle that so-called version 3, but not sure that there is an official ver 3 for WFS.

PR #2270, opened 2025-12-15
feature: real-time layer

### LINKED ISSUE #2262: Real time layer
__Description__
I propose introducing a new layer type that handles real time data that updates frequently

__Proposed Solution__
- A new layer and source type  is implemented 
- The source subscribes to an event source
- The event source is implemented outside the scope of origo, thus making the implementation of the layer source an api specification.
- Events are sent using Server Side Events (SSE) whenever something changes
- The event can also be the entire new current state
- On connection a complete initial state may be sent

The rationale behind the design decisions is:
- SSE is extremely lightweight in the client
- SSE can utilize same authentication mechanisms  as ordinary GET requests (cookies, auth headers etc)
- Putting the communication with the data source server side hides api keys
- The data source can be just about anything. In this way the client only has to support one simple event while the server side can implement just about any complex scenario (MQTT, Database, REST Api, Rabbit MQ etc)

__Alternatives Considered__
Create an integration on the server that populates a database table periodically and just trigger a refresh on the source periodically as proposed in #1631. But that could potentially be resource consuming.

__Additional context__
This solves the same problem as #1631, but in a less brutal way. Also somewhat related to #1433 which was resolved by implementing a javascript timer (external to origo) refreshing the source.

steff-o:
Resolves #2262.

Creating as draft as we are not ready with the backend, which may seriously change the requirements on the layer type. But it is working for you all to enjoy during the Christmas holiday.

In order to test it you will need a backend sending SSE events. The event should be a named event `update` for updates and inserts, they are treated equally like PUT. The payload should be exactly one feature encoded as GeoJson. For deletions a `delete` event should be send with an object with one property called `id` containing the id (as previously sent as id in GeoJson) to delete. To connect to the server the server should accept a query-argument named `layer`.

There is no fetching of an initial state, so the layer will be empty until events arrive. But the server may send a series of events on connect. Nothing special about them, but it will emulate an initial state. I have also considered adding support for fetching an initial state through a WFS-endpoint, but since the rest of the code does not use WFS and the source may not be backed up by a database  I thought it is easier if the real-time server provides it as ordinary events. For large datasets it could possibly become slow, but by sending the events individually instead of one giant combined initial event it will not reach buffer limits.

I have implemented a reference implementation of a server to demonstrate the functionality without having to create a backend: https://github.com/ornskoldsvikskommun/origo-realtime-reference-api It is pretty simple and uses a PostGis database with triggers as source.

The implementation uses Server Side Events (SSE) as specified by the `EventSource` specification. It has some serious limitations when using HTTP/1.1 as it will for each layer hog one of the six available connections for each remote host. Normally this is not a problem as most server sides can be written to support HTTP/2, either natively or by putting a HTTP/2 capable reverse proxy (HAProxy, Nginx, IIS or whatever) in front of it. The only situation where it can't be avoided is when using _IIS_ with _Windows Integrated Authentication_. It will work, but connection will be downgraded to HTTP/1.1. Some other limitations is that it is not possible to send custom headers and the connection can only be instantiated with a GET-request.

# Features
- Automatic update of changed features, attributes and geometries
- Disconnect from events when not visible
- Automatic reconnect when visible
- Toaster when connection is lost
- Toaster when connection is restored
- Feature info, but popup does not move when object moves as featureInfo is a can of worms.
- Named sources, just like WFS layers.

# Configuration
``` jsonc
"source": {
    "realtime": {
      "url": "http://localhost:3003/subscribe", // A query param layer=id will be added. Other query params are possible in the base
      "projection": "EPSG:3006" // Defaults to 4326, so not needed if following spec. Can also be set on layer.
  },

"layers": [
{
      "name": "realtime punktlager", // Internal Origo name
      "id": "punktlager", // Sent as layer parameter to source. If not present name is used
      "title": "RealTime Punkt",
      "group": "root",
      "source": "realtime", // A named source 
      "type": "REALTIME", // Must be REALTIME
      "attributes": [
        {
          "title": "Fritext",
          "name": "fritext",
          "type": "text"
        }
      ],
      "projection": "EPSG:3006", // Defaults to 4326, so not needed if following spec. If present overrides source
      "geometryName": "the_geom", // name of geometry in OL for created features
      "visible": true
    }
]
```

# Future Improvements
Here are some ideas. Some may be added to this PR, some will have to wait for someone to need it.
- Support for websockets if it turns out that SSE isn't as good as I thought. It could easily be implemented as an alternative transport in the same source class
- Configurable CORS. Right now it is always on
- Update popup on events
- Support for multiple popups would be cool
- Better support for initial state
- Configurable if layer should disconnect when hidden
- Filters


steff-o:
@02JanDal, interesting input. Before I wrote this implementation I did a quick survey to see if there were any standards to use before making my own format as standards would be the way to go. Being a GIS application, the first place I looked was OGC, but I thought OGC API Features pub/sub only supported MQTT and I wanted to keep the implementation dependency free and very lightweight footprint and then chose SSE with custom events as in general there seemed to be no clear standard that stood out.

After I read up on OGC API again, most of the OGC API features seems to be about searchability and metadata, which in the end will give you a link to the broker where an actual stream can be found. My implementation cuts to the chase and goes straight to the SSE stream as the author of the Origo configuration would most likely know the correct connection parameters.

My goal was from the beginning to implement a lightweight client which communicates with a custom server side that is easy to implement for different source data instead of making layer types for every possible type of real time source. The aim was to only make the server side a proxy to external real time data sources. As there are too many different kinds of real time sources, I thought that hiding implementation server side would make it easy to provide for any kind of source.

It is a good point that a real time layer could be a OCG Feature layer with real time support, but in theory any kind of layer could potentially return a link to a stream where future updates to the initially received data will be published. In fact one of my first ideas was to extend the WFS layer with real time capabilities, but with static configuration instead of receiving a URL to the stream. WFS would been used to get the initial state, followed by a SSE stream with updates.

When testing my implementation I have also identified the possible need of a "landing url" before connecting to the actual stream. In my case it is mainly because SSE does not support enough mechanisms to provide for authentication and error handling. Taking inspiration from Trafikverkets real time api: https://data.trafikverket.se/documentation/datacache/the-request such an implementation would make use of a REST POST endpoint where the initial filters and authentication is sent. The initial request would then create a session and return a link to the stream for that session using the session id as parameter. When you think about it, that is pretty much like an OGC Features api request that returns a link to a stream.

So, your comment has really questioned my descicions and I'm totally uncertain where to go next. Implementing a complete OGC Features Api layer seems like a bigger task than I expected, but in the end probably useful.  We actaully have an issue on creating a OCG Features layer (#1977). Adding real time capabilities to a layer is actually a pretty small task. My implementation is less than 100 relevant lines of code. Implementing the events in the OGC proposed Geojson payload format would be no big task either. It's pretty much the same format as I use although my format is not as formal. Both are basically a GeoJson wrapped in a control object with some event information . It would add a bit to the overhead though.

Maybe I could have a look at implementing a very minimal OGC Features Api layer with an SSE stream using the geojson or CloudEvents payload.

02JanDal:
Given that it's the only existing server implementation for this I'm aware of but also usually is one of the first to implement new OGC API standards, I'd probably lean towards whatever would work with [pygeoapi](https://docs.pygeoapi.io/en/latest/pubsub.html#). Annoyingly the documentation doesn't provide a lot of hints if one doesn't want to use MQTT though. It should be relatively trivial to create a simple broker that accepts SSE clients and events as sent by pygeoapi's HTTP-broker option, at least for testing (might also be an option to integrate that into origo-server).

Your option 3. definitely has the upside of following the existing (draft) specifications the closest. On the other hand, SSE would be the better technical choice here as it's a unidirectional data flow.

I think a decent way to tell the client that a hub speaks SSE would be to simple use the link `type` attribute set to `text/event-stream`:

```json
"links": [
  ...
  {
    "rel": "hub",
    "href": "https://...",
    "type": "text/event-stream"
  }
]
```

Alternatively, I don't think simply deciding that Origo only speaks SSE (for now) and documenting that would be terrible either.

steff-o:
Now in "ready for review". No actual changes since feb 27, except some additional reconnect logic and clean up . See above post for how it works. Documentation will be written soon.

Issue #2276, opened 2026-01-30
Add a measure subtool to measure existing objects


02JanDal:
**Is your feature request related to a problem? Please describe.**
Users sometimes want to measure an existing object, such as a line or polygon. Re-drawing by hand using the current measurement is both annoying and error prone.

**Describe the solution you'd like**
Allow the user to enable a measure subtool (similar to line and area measurements) and then select an existing object.

**Describe alternatives you've considered**
Drawing by hand.

**Additional context**

PR #2277, opened 2026-01-30
feature: measure tool for measuring existing objects

02JanDal:
https://github.com/user-attachments/assets/d7b80b9e-3921-4d23-a7a2-80c71f6c2c78

Closes: #2276 

Not super happy about the icon, but couldn't find anything better either.


Grammostola:
I tested this with a Geoserver 2.28.1 instance.

- If using WMS layers from postgresql views geoserver assigns random feature ids upon getfeatureinfo request which means multiple measure actions can be taken on a single polygon or line, this results in an increasingly visible style change. This may be an issue with not having configured Geoserver for stable keys for such layers however, but it might be worth documenting for the tool.

- If using WMS layers with `infoFormat` set to something other than `application/json` (default) then the getfeatureinfo that this tool sends in order to get a stable ID and geometry back will probably not result in that, so the tool should probably always issue getfeatureinfo requests with  infoFormat `application/json`.

- I note that only geometry types `line` and `polygon` are accepted. Would it be a smoother user experience if `multiLine` and `multiPolygon` were accepted as well?

steff-o:
Tested with GeoServer and it works like expected, but:
- I agree with @Grammostola that MultiLine and MultiPolygon should be supported. In origo it is not common to create multiparts, but it is possible (with just one ring) and external sources could very well be multiparts. It requires some additional fixing of the old code, but not much. No sure if measure should just explode multiparts and label each sub part or label just the aggregate. Both is better that just ignoring them. Labling both would probably look messy.
- I don't think the stable key issue is an issue. It requires the user to click the same feature multiple times and can easily be cleared. The same problem could also arise in WFS layers if using BBOX without stable keys, which would load the same feature for each new extent. It is just bad practice to not have stable keys.
- The icon should be changed. The current looks like a color picker. I suggest using just a plain old arrow from material:
image

- The featureInfo tool has an option to change the mouse pointer if there is something to click where the pointer is. That would be cool if that could be reused, but not necessary if it means duplication the code.

02JanDal:
Added support for multilines- and polygons, force using application/geo+json (should work with most servers, unless they only support GML or similar), fixed a small bug and changed the icon to left_click_24dp_1F1F1F_FILL0_wght400_GRAD0_opsz24

Stable IDs: I don't think there's a lot I can reasonably do here sadly, other than storing the entire feature including attributes and geometry and compare it

Different pointer on hover actually already works using the functionality from featureinfo.

steff-o:
Tested everything OK except the geojson part as a don't understand the problem. @Grammostola can you have a look.

Also, there are some silly linting errors. Seems like linting is not performed on check in until someone has approved it.

PR #2300, opened 2026-03-30
fix: update origo member mask

Grammostola:
Seeks to fix part of #2299 (the mask)

It isn't perfect for I haven't gotten ahold of the administrative borders that osm utilizes. Most, but not all of the mask so far seems to follow those borders exactly. 

(When and if this is merged I'll see whether the demo-map workflow works as intended)


Grammostola:
I am preparing a revision for osm borders for I got a hold of them eventually so please stand by for an update (before testing)

Grammostola:
The demo_map_workflow.yml keeps the state it had when this PR was created so currently the demo map is a mirror of this pr .. this should not be the case for new PRs.

johnnyblasta (approved):
LGTM
Unesecarry many decimals used, but that was also the case before on some areas.

PR #2306, opened 2026-04-10
feature: Add debouncer to search field

### LINKED ISSUE #2151: Add setting for a minimum interval between keystrokes before search
There always seems to be a search triggered for every keystroke even though they're done i fast succession. Adding a short timer that gets reset on each keystrok would lower the load on the backen.

Make the request with a slight delay, a few ms, before sending the request. This delay could be a setting. A setting of 0 disables the feature. This could be done with a combination of setTimeout and clearTimeout. So each keypress clear any previous timeouts and sets a new one. Given that the specified time has passed and no keystroke has been made the request fires. 

Some other solutions that waits a few milliseconds to wait for another keystroke might work.

**Additional context**
Always querying for every keystroke is not very nice to the backend.

dafvid:
Fixes #2151

Wraps the call to fetch() in a setTimeout with a delay set to value of option searchDelay. Default is 400 ms, 0 practically disables the delay.

### DIFF
--- a/src/controls/search.js
+++ b/src/controls/search.js
@@ -38,7 +38,8 @@ const Search = function Search(options = {}) {
     maxZoomLevel,
     limit,
     hintText,
-    minLength
+    minLength,
+    searchDelay = 400
   } = options;
 
   const {
@@ -69,6 +70,7 @@ const Search = function Search(options = {}) {
   let containerElement;
   let wrapperElement;
   let infowindow;
+  let debounceTimer;
 
   function clear() {
     featureInfo.clear();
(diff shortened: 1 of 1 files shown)


src/controls/search.js
```diff
@@ -49,7 +49,8 @@ const Search = function Search(options = {}) {
     searchlistOptions = {},
     queryType,
     suppressDialog,
-    localization
+    localization,
+    searchDelay = 400
```
steff-o, line 53: Lint complains that searchDelay should be const.

steff-o (changes requested):
Almost approved. Got one linting error. GitHub does not catch it, but running `npm run lint` locally does.

Other than that, approving although the debouncing will fire off a search when backspacing beyond the _minLength_ with the search string value of the last _minLength_ string. E.g when typing "abcde" and wait for the request to fire. Then rapidly backspace five times and wait. This will fire off a search with the string "abcd" if minLength is 4. No biggie, as the result is thrown away but ideally the debouncer should realize that the backspacing resulted in a no-op.

Also there are some unnecessary requests fired when selecting the text but that seems to be an old issue.

Also remember to update documentation with the searchDelay option and link that PR to this PR.

src/controls/search.js
dafvid: It is const. Have you pulled the latest version? [194699a](https://github.com/origo-map/origo/pull/2306/commits/194699a26b639a3fddbaa82337f290ce084b6afc) fixes const

steff-o (approved):
LGTM!

Sorry that I failed the review earlier. I tested the wrong branch.

PR #2315, opened 2026-04-29
Add layer type COG

### LINKED ISSUE #2305: Add support for COG (Cloud Optimized GeoTIFF)
Origo is missing support for COG (Cloud Optimized GeoTIFF) type layers. OpenLayers have support for COG-files. This enables being able to view extremely large images (hundreds of GB) files directly from a webserver without the need for WMS-services. It's similar to ECW but FOSS.

To add a new COG layer type.

There are no alternatives really.

**Additional context**
https://cogeo.org/
https://openlayers.org/en/latest/examples/cog.html

dafvid:
Fixes #2305 

This adds the layertype COG. This is written in part with the help of Claude AI. Please look at how the layer options is set up and if it's integrated correctly with the other layer type.

This requires WebGL to function since it renders arrays of pixels from bytebuffers. WebGL is supported by all major browsers. Also on mobile.

Openlayers does not support COG rendering of skewed and/or rotated images (affine transformed). So any imagery has to unrotated and unskewed to render correctly.

### DIFF
--- a/src/layer/cog.js
+++ b/src/layer/cog.js
@@ -1,0 +1,64 @@
+import GeoTIFFSource from 'ol/source/GeoTIFF';
+import WebGLTile from 'ol/layer/WebGLTile';
+
+function createSource(sourceInfo, sourceOptions, geoTIFFOptions = {}) {
+  return new GeoTIFFSource(Object.assign({}, geoTIFFOptions, {
+    sources: [sourceInfo],
+    sourceOptions
+  }));
+}
+
+const cog = function cog(layerOptions, viewer) {
+  const cogDefault = {
+    layerType: 'tile',
+    crossOrigin: 'anonymous'
+  };
+
+  const cogOptions = Object.assign({}, cogDefault, layerOptions);
+  const mapSource = viewer.getMapSource()[layerOptions.source] || {};
+  const sourceInfo = {};
+  const sourceOptions = {};
+
+  sourceOptions.headers = cogOptions.headers;
+
+  if (cogOptions.source && viewer.getMapSource()[cogOptions.source]) {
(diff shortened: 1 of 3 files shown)


Grammostola:
Nice work, seems to work as intended, incl printing and OL opts for the layer. If the coordinate system of the map differs from the cog tiff and the proj4 definition hasn't been provided with the config then there's a crash. That could be part of the documentation, and and anyway reprojecting might be resource intensive for the client (it is with a 230GB orthophoto at smaller scales). This orthophoto also becomes a bit slow at smaller scales, that could be due to configuration including calling overviews (we typically create overviews at the esri:mosaic dataset level)

I wonder whether it would make sense to put the 'ol/layer/WebGLTile' in a separate file called webgltile.js, like how it is currently arranged where ol/layer/* are in their own files for reuse. WebGLTiles may be interesting to other layer types than cog. 

This pr needs a corresponding issue and PR for the https://github.com/origo-map/origo-documentation

dafvid:
There's not much to put in a separate file really. It's just an import at the top and then a constructor at the bottom. Not sure what the content of webgltile.js would be?

Grammostola:
```js
import WebGLTile from 'ol/layer/WebGLTile';

export default function webgltile(options, source) {
  const opt = options;
  opt.source = source;
  return new WebGLTile(opt);
}
```
This would align it with the other files in this directory which import OL layer classes (and are not employed in Origo as layer types directly)

In addition unless I'm misunderstanding the ol/layer/WebGLTile can use an [ol​/source​/ImageTile](https://openlayers.org/en/latest/apidoc/module-ol_source_ImageTile-ImageTileSource.html in addition to the GeoTIFF source.

PR #2326, opened 2026-05-22
Display entire dms string

### LINKED ISSUE #2318: WGS84 DMS fails to display full coordinates
**Describe the bug**
When activating the location marker for coordinate systems and selecting Wgs84 DMS, half of the coordinates are not displayed. This issue resolves itself if you zoom in or out using the mouse wheel and then cycle back to Wgs84 DMS. However, when you refresh the page and activate the location marker again, it reverts to showing only half of the coordinates for Wgs84 DMS.
 
All other coordinate systems work fine, only WGS84 DMS is affected.
 
**To Reproduce:**
1.Click crosshairs to activate location marker
2. Click the coordinate system to cycle to Wgs84 DMS
3. Observe that only half of the coordinates are displayed
 
**Expected behavior**
All coordinates should be clearly displayed when WGS84 DMS is selected as the coordinate system and the location marker is active.
 
**Screenshots**
When first selecting Wgs84 DMS.
Image

After zooming in and cycling through the coordinate systems back to WGS84 DMS
Image

 
**Desktop** (please complete the following information):
- OS: 26100.8246
- Browser (Chrome, Edge)
- Version: Windows 11 Enterprise
 
**Additional context**
I have tried switching the position of the scale to see if that would make a difference, but it didn’t.
Image

 I have tested different versions of Origo, and none of them worked (latest version tested: 16 February 2026, local version).

steff-o:
Fixes #2318

Fixed by always ensuring that the input field always accommodates the maximum number of possible characters in a dms string. Previous implementation used the center coordinate when toggling. If the coordinate was on an even second the string became too short to display fractions of a second.


Grammostola:
The fix looks good, though there appear to be a few too many files in the pr.

PR #2328, opened 2026-05-25
feature: add multiple attachments

### LINKED ISSUE #2314: Attachments - allow multiple files upload
When adding attachments the file input dialog only allows selecting one file. This makes adding many attachments tedious.

The file input dialog should support multiple files. It should be the new default and only behavior.

steff-o:
Fixes #2314

Enables selecting multiple files according to the browser's files selector.

No new configuration, it is always enabled.


johnnyblasta (approved):
LGTM

PR #2331, opened 2026-05-29
Feature/lazy loading

### LINKED ISSUE #2321: Lazy loading of legend images with GetLegendGraphics
On map startup ALL legend images are loaded even though NONE are visible. This might be fine on smaller maps (< 100 layers) but becomes a DDOS attack on the mapservers above 500 layers. To request one image from every layer of every project on the whole server at once puts a lot of strain on the backend. The startup of the map is slowed significantly since the browser can only request 10 images at a time, the rest of the 700 requests are pending. Unless all the legends are cached in advance the startup of the map takes around 20 s.

To lazy load images just before they are needed instead of during startup. A startup of the map with 700 layers leaves a lot of them in the state of pending since it can only request 10 at a time. 

No real alternative

**Additional context**
We at Södertälje have a working solution that have no effect on user experience beside much faster startup of the map. But the opening of legend groups look and feels just the same even though the images are loaded on hover before the click. We have a PR ready as soon as this issue is accepted.

TobiasSodertalje:
Fixes #2321

Legend images are lazy-loaded by initially rendering them with data-legend-src instead of src. When a user interacts with the legend (hover/click/focus), we convert data-legend-src to src, which triggers the actual image request. This avoids loading all legend images up front and improves initial render performance.


Grammostola:
Appears to work rather well, for mobile users at least with the simulated touch input mode. Legends are gotten when groups are expanded then, else as soon as the cursor hovers over them.

I encountered issues with WMS layers (which is what I've tested so far):
1. If assigned to `group: root` the legend is never fetched
2. When using `"visibleLayersControl": true` with the legend control if a layer with a functioning legend is activated and then the visibleLayersControl is, the legend for the layer disappears while the visibleLayersControl is active

steff-o:
To fix the missing root-group symbols the `addLayer()` function in overlays.js must be changed.
Either by forcing the lazy load src exchange when overlay is created:

```javascript
const addLayer = function addLayer(layer, {
    position
  } = {}) {
    const styleName = layer.get('styleName') || null;
    const layerStyle = styleName ? viewer.getStyle(styleName) : undefined;
    const overlay = Overlay({
      layer, style: layerStyle, position, viewer, localization
    });
    const groupName = layer.get('group');
    if (rootGroupNames.includes(groupName)) {
      rootGroup.addOverlay(overlay);
      // Make sure remote symbols are loaded as root group is
      // visible from start
      activateLazyLegendImages(document.getElementById(overlay.getId()));
    } else if (!(nonGroupNames.includes(groupName))) {
      const groupCmp = groupCmps.find((cmp) => cmp.name === groupName);
      if (groupCmp) {
        groupCmp.addOverlay(overlay);
        document.getElementById(groupCmp.getId()).classList.remove('hidden');

        if (groupCmp.parent) {
          updateLegend(groupCmp);
        }
      } else {
        console.warn(`${localize('addLayerWarningGroup')} ${groupName} ${localize('addLayerWarningDNE')}`);
      }
    }
  };
```

or by passing `lazy = false` to `Overlay()`, which requires some refactoring to determine if it should be lazy or not before calling and passing along the lazy-flag all the way down to the function that actually cares about it.

PR #2335, opened 2026-06-05
fix: replace webpack with vite

### LINKED ISSUE #2332: Implement Vite as build system

sweco-sedalh:
Replaces webpack with vite. Most of the development experience should be the same (except a lot quicker), as both command to start development server and development server port is the same.

Most of the vite configuration is required to replicate the existing structure of the `build` folder.

A summary of the necessary changes:
* Replace webpack configuration files by vite.config.js, working in library mode during build
* Replace JavaScript file references in HTML files with imports
* Move most static files into the `public` folder (those contents are automatically copied by vite during build), index.json is kept at the repository root as it's where people expect it to be
* Rewrite some other files (in particular HTML files) during build
* Add proper metadata to package.json making it easier to consume Origo as a npm package, either pre-built or as source

Not done as part of this PR, could be added or done as a follow up:
* Show eslint issues in web browser (similar to the previous webpack lint configuration)
* Optimize code-splitting/chunking including lazy loading


Grammostola:
Nice! The `build` and `run` speeds are _much_ faster. 

There seems to be a problem with the `run` part of vite.config.js in that the app doesn't load properly in a browser:
`Loading module from “http://localhost:9966/index.html?html-proxy&index=0.js” was blocked because of a disallowed MIME type (“text/html”)` which seems to be fixable by making the viteStaticCopy plugin conditional and only for `build`

There's a similar error when trying to use the dist/ contents after `build` although the build/ contents appear to work

I also note that during `run` .scss and .js file changes are monitored and accounted for whereas changes to index.json is not which differs from the current webpack conf.

steff-o:
Does this mean that applications that want to an upgrade Origo version built with vite has to change its code, i.e. use "module" and "import".

Not sure if that counts as not being backwards compatible or not, but it means that a new build is not a drop-in replacement. If it enables some new possibilities maybe that's the way to go to pave way for new upcoming cool features like chunking.

PR #2356, opened 2026-08-10
fix: ensure that interactions are disabled if the editor control is set not to be active at start

### LINKED ISSUE #2357: The edit control's modify interaction is always active at start
**Describe the bug**
The edit control activates a modify interaction at start even if the option isActive is set to false.

**To Reproduce**
1. Set the editor control option isActive to false.
2. Start the application.
3. Click on a feature in the editor-layer.
4. The modify interaction is active for the current layer.

**Expected behavior**
The modify interaction should not be active if the isActive property is false.

erikedmark:
When the editor-control starts it activates the modify interaction even if the option isActive is set to false. The interaction should only be set active when the control itself is active.

### DIFF
--- a/src/controls/editor/edithandler.js
+++ b/src/controls/editor/edithandler.js
@@ -1002,7 +1002,7 @@ function setEditLayer(layerName) {
 function setGeometryProps(layer) {
   const layerName = layer.get('name');
   editLayers[layerName].set('geometryType', layer.getSource().getFeatures()[0].getGeometry().getType());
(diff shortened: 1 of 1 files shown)


steff-o:
Please note that there is already a PR #2282 that removes the `isActive()` function, which will create a conflict for these two PR:s depending on which gets merged first. The logic in `isActive()` is badly written and should be fixed if it shall remain in the code.

Could you check if #2282 solves your problem as the logic for activating interactions has been rewritten.

PR #2358, opened 2026-08-10
feat: add support for base64-formatted images in the default info window

### LINKED ISSUE #2359: Add support for base64-formatted images in the default popup window
**Problem**
When the default popup is used that renders attribute based content, images of base64-format are not displayed.
The result is that a url with the value "undefined" is used as the src-value of the image tag and furthermore displayed as a broken link in the browser.

**Expected behaviour/solution**
If the image attribute value is of base64-format the the image shuld be used directrly as the src-property for the img-tag.

erikedmark:
Adds support for base64-formatted images in the default attribute based popup-window.

If the attribute value is of type "image or img" and the value looks like a base64-formatted string strarting with the word "data", it will be considered to honor that value as the image src-value, discarding the use of the createUrl-function.

### DIFF
--- a/src/getattributes.js
+++ b/src/getattributes.js
@@ -127,9 +127,15 @@ const getContent = {
     } else {
       const featGet = attribute.img ? feature.get(attribute.img) : feature.get(attribute.name);
       if (featGet) {
-        const url = createUrl(attribute.urlPrefix, attribute.urlSuffix, replacer.replace(feature.get(attribute.img), attributes, null, map));
-        const attribution = attribute.attribution ? `<div class="o-image-attribution">${attribute.attribution}</div>` : '';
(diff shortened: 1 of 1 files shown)


steff-o:
I haven't tested yet, just did a little static code review and noted:

- The code checks for a string that starts with `'data'` when to use the base64 shortcut and ignore urlPrefix/suffix. It is not unlikely that relative url:s, e.g. folder names starts with 'data'. If the check is against `data:` it would be more robust as colon is unlikely to appear unescaped.
- The base64 "shortcut" adds a title not present in the old code. Why?
- The base64 "shortcut" does not call `replacer.replace()`. It is highly unlikely that the base64 string contains any replaceable strings, but the entire base64 string can be in another variable (also very unlikely, but still) referenced by the template string.
- I think the code would look cleaner if the base64 shortcut is moved to createUrl.

PR #2361, opened 2026-08-18
fix: make the left tools reachable

### LINKED ISSUE #2021: The toolbar needs to know the size of the browser

johnnyblasta:
Fixes #2021 by making the tools scrollable. Tested on both desktop and mobile.

### DIFF
--- a/scss/_viewer.scss
+++ b/scss/_viewer.scss
@@ -8,6 +8,31 @@
         flex-flow: column;
     }
 
+    // Make the tools reachable when they don't fit the height of the map, e.g. in
+    // a short browser window or on a phone in landscape mode.
+    #o-tools-left {
+        overflow-x: hidden;
+        overflow-y: auto;
+        overscroll-behavior: contain;
+        // Hide the scrollbar to keep the map clean. Scrolling is done with the
+        // mouse wheel, by touch or by tabbing through the buttons.
+        scrollbar-width: none;
+        // A scroll container clips its content, so the container is stretched to
+        // leave room for the button tooltips. It does not catch any pointer
+        // events, only its buttons do, see .transparent in ui/helpers.
+        width: 100%;
+
+        &::-webkit-scrollbar {
+            display: none;
+        }
+
+        // Keep the toolbars at their own size in the stretched container.
+        >* {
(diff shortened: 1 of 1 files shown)


steff-o:
Tested and seems to do exactly what is advertised, but I have only tested in Chrome. It would be great if someone could test in Safari as webkit is involved.

Also I would like a visual hint that it is possible to scroll when scrollbar is explicitly hidden.

Asked some AI and it suggested adding a fade to last elements to trigger the brain to try to scroll :

tool-scroll



The code for this is pretty simple:
```scss
$fade-height: 80px;
-webkit-mask-image: linear-gradient(to bottom, black calc(100% - #{$fade-height}), transparent 100%);
mask-image: linear-gradient(to bottom, black calc(100% - #{$fade-height}), transparent 100%);
```

Other suggestions:
- Make sure container always clips the last visual menu button to trigger the brain to scroll
- Injecting an arrow below last menu item. Requires some additional checks as when the last item is visual it is not needed
- Fade in a transparent scrollbar when hovering. Would not work on touch screens. In that case the scrollbar must be visible all the time.

johnnyblasta:
There is a visual hint with clipping the last button, see image
image
Although it can fail if the clip is done exactly in the gap between two buttons,

johnnyblasta:
Changed to fade out instead of cutting at the bottom of toolbar.
image

steff-o (approved):
LGTM!

